PR #66, opened 2022-08-17
feat: add flagd provider implementation

### LINKED ISSUE #68: flagD Java Provider
https://github.com/open-feature/java-sdk-contrib/tree/main/providers/flagd

toddbaert:
- adding grpc generation for flagd communication (git submodule pull of schemas repo in maven exec plugin)
- provider implementation
- tests

### DIFF
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
@@ -45,9 +45,9 @@ public class FlagdProvider implements FeatureProvider {
      * @param host     flagd host, defaults to localhost
      * @param port     flagd port, defaults to 8013
      */
-    public FlagdProvider(String protocol, String host, int port) {
+    public FlagdProvider(Protocol protocol, String host, int port) {
         
-        this("https".equalsIgnoreCase(protocol) 
+        this(Protocol.HTTPS == protocol 
             ? ServiceGrpc.newBlockingStub(ManagedChannelBuilder.forAddress(host, port)
                 .useTransportSecurity()
                 .build()) :
@@ -60,7 +60,7 @@ public class FlagdProvider implements FeatureProvider {
      * Create a new FlagdProvider instance.
      */
     public FlagdProvider() {
-        this("http", "localhost", 8013);
+        this(Protocol.HTTP, "localhost", 8013);
     }
 
     /**
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
@@ -29,6 +29,7 @@ import dev.openfeature.javasdk.Value;
 import io.grpc.ManagedChannelBuilder;
 import lombok.extern.slf4j.Slf4j;
 
+
 /**
  * OpenFeature provider for flagd.
  */
@@ -37,7 +38,9 @@ public class FlagdProvider implements FeatureProvider {
 
     private ServiceBlockingStub serviceStub;
     static final String PROVIDER_NAME = "flagD Provider";
-
+    static final int DEFAULT_PORT = 8013;
+    static final String DEFAULT_HOST = "localhost"; 
+    
     /**
      * Create a new FlagdProvider instance.
      *
@@ -60,7 +63,7 @@ public class FlagdProvider implements FeatureProvider {
      * Create a new FlagdProvider instance.
      */
     public FlagdProvider() {
-        this(Protocol.HTTP, "localhost", 8013);
+        this(Protocol.HTTP, DEFAULT_HOST, DEFAULT_PORT);
     }
 
     /**
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
@@ -1,29 +1,294 @@
 package dev.openfeature.contrib.providers.flagd;
 
-import dev.openfeature.javasdk.Client;
-import dev.openfeature.javasdk.NoOpProvider;
-import dev.openfeature.javasdk.OpenFeatureAPI;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
 
-/** 
- * A placeholder.
+import org.apache.commons.lang3.EnumUtils;
+
+import dev.openfeature.flagd.grpc.Schema.ResolveBooleanRequest;
+import dev.openfeature.flagd.grpc.Schema.ResolveBooleanResponse;
+import dev.openfeature.flagd.grpc.Schema.ResolveFloatRequest;
+import dev.openfeature.flagd.grpc.Schema.ResolveFloatResponse;
+import dev.openfeature.flagd.grpc.Schema.ResolveIntRequest;
+import dev.openfeature.flagd.grpc.Schema.ResolveIntResponse;
+import dev.openfeature.flagd.grpc.Schema.ResolveObjectRequest;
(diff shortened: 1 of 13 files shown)


providers/flagd/src/main/proto/schema.proto
```diff
@@ -0,0 +1,97 @@
+syntax = "proto3";
```
toddbaert, line 1: We will pull this in as a submodule.

providers/flagd/src/main/proto/schema.proto
```diff
@@ -0,0 +1,97 @@
...
+
+package schema.v1;
+
+option go_package = "schema/service/v1";
+option java_package = "dev.openfeature.flagd.grpc";
```
toddbaert, line 6: Will add this line to the schema repo

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
```diff
@@ -1,29 +1,185 @@
...
+        stringAttributes.keySet().stream().forEach((String key) -> values.put(key,
+            Value.newBuilder().setStringValue(stringAttributes
+                .get(key))
+                .build()));
+        intAttributes.keySet().stream().forEach((String key) -> values.put(key, Value
+            .newBuilder().setNumberValue(intAttributes.get(key))
+            .build()));
+        return Struct.newBuilder()
+            .putAllFields(values)
+            .build();
```
toddbaert, line 182: I don't like any of this.

eclipse-java-google-style.xml.bak
```diff
@@ -0,0 +1,337 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
justinabrahms: .bak file intentional?

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
```diff
@@ -1,29 +1,296 @@
...
+    /**
+     * Create a new FlagdProvider instance.
+     *
+     * @param protocol transport protocol, "http" or "https"
+     * @param host     flagd host, defaults to localhost
+     * @param port     flagd port, defaults to 8013
+     */
+    public FlagdProvider(String protocol, String host, int port) {
+        
+        this("https".equalsIgnoreCase(protocol)
```
justinabrahms: ⛏️ Seems like putting an awful lot on just a few lines.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
```diff
@@ -1,29 +1,296 @@
...
+     *
+     * @param serviceStub service stub instance to use
+     */
+    public FlagdProvider(ServiceBlockingStub serviceStub) {
+        this.serviceStub = serviceStub;
+    }
+
+    @Override
+    public Metadata getMetadata() {
+        return new Metadata() {
```
justinabrahms, line 80: Should we just make this a default implementation in the java sdk? Take a name param, return it as name.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
```diff
@@ -1,29 +1,296 @@
...
+    private Reason mapReason(String flagDreason) {
+        if (!EnumUtils.isValidEnum(Reason.class, flagDreason)) {
+            // until we have "STATIC" in the spec and SDK, we map STATIC to DEFAULT
+            if ("STATIC".equals(flagDreason)) {
+                return Reason.DEFAULT;
+            } else {
+                return Reason.UNKNOWN;
+            }
+        } else {
+            return Reason.valueOf(flagDreason);
```
justinabrahms, line 172: Does this return null if invalid reason?

justinabrahms (commented):
Tbh, this didn't look fun to write. :-/

eclipse-java-google-style.xml.bak
toddbaert: I was removing this as you were reviewing, apparently. It's gone now!

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
toddbaert: Kinda have to. `Constructor call must be the first statement in a constructorJava(1207959691)`; this messy block allows calling other constructors while also having the logic I want.

Simply adding an `if` makes the compiler mad.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
toddbaert: I think providing as many defaults as possible makes sense, but I'm not 100% sure what you're proposing.

Where would a name param be taken?

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
toddbaert: I should add a test for this now that you mention it, but this checks that `flagDreason` (a string) can be mapped to a `Reason` enum value, and if it can't it returns `Reason.UNKNOWN`.

There's a special case for now with `"STATIC"`, which we want to eventually add to the spec, and which we're mapping to `Reason.DEFAULT` for now.

toddbaert:
@justinabrahms 

> Tbh, this didn't look fun to write. :-/

It wasn't at first, but the SDK changes around the JSON values made it SO MUCH BETTER, I said as much to @beeme1mr earlier. Dealing with the generated protobuf code is a bit tedious, but overall it saves you from writing a ridiculous amount of boilderplate. If you look in `target/generated-sources/` you can see how much java is generated from the protobuf spec.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
justinabrahms:
```
return new Metadata(PROVIDER_NAME);
```

Instead of having to declare an inline class.

justinabrahms (approved):

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
toddbaert: Ya, I'm in favor of that!

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
toddbaert: Added a test for this.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
```diff
@@ -1,29 +1,294 @@
...
+    static final String PROVIDER_NAME = "flagD Provider";
+
+    /**
+     * Create a new FlagdProvider instance.
+     *
+     * @param protocol transport protocol, "http" or "https"
+     * @param host     flagd host, defaults to localhost
+     * @param port     flagd port, defaults to 8013
+     */
+    public FlagdProvider(String protocol, String host, int port) {
```
beeme1mr: Can protocol be an enum and can port be a short?

toddbaert: > Can protocol be an enum and can port be a short?

`short` goes up to 32767, while ports go up to 65535, so I think that's a bad idea. I can make the protocol an enum.

toddbaert: Done!

beeme1mr (approved):

PR #106, opened 2022-09-22
feat: Go feature flag provider

thomaspoignant:
First version of the `provider` for GO Feature Flag.


thomaspoignant:
The stage `Configure GPG Key` in the CI is failing.
I guess this is because this PR is open from a fork, maybe we should have an issue for that.

toddbaert:
> The stage `Configure GPG Key` in the CI is failing. I guess this is because this PR is open from a fork, maybe we should have an issue for that.

I was just about to make a comment about this. I will fix this in another PR and you can rebase on it. It will take only a moment.

toddbaert:
https://github.com/open-feature/java-sdk-contrib/pull/107 after this is merged you can rebase if you want. Sorry about that!

providers/go-feature-flag/src/main/java/org/gofeatureflag/provider/GoFeatureFlagProvider.java
```diff
@@ -0,0 +1,246 @@
...
+     */
+    private void initializeProvider(GoFeatureFlagProviderOptions options) {
+        // Register JavaTimeModule to be able to deserialized java.time.Instant Object
+        requestMapper.disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS);
+        requestMapper.enable(SerializationFeature.INDENT_OUTPUT);
+        requestMapper.registerModule(new JavaTimeModule());
+
+        // init httpClient to call the GO Feature Flag API
+        int timeout = options.getTimeout() == 0 ? 10000 : options.getTimeout();
+        this.httpClient = new OkHttpClient.Builder()
```
toddbaert: Does this client use a keepAlive or connection pool?

If you're often hitting the same domain (I guess you are?) it can improve performance.

providers/go-feature-flag/src/main/java/org/gofeatureflag/provider/bean/GoFeatureFlagUser.java
```diff
@@ -0,0 +1,36 @@
...
+     * fromEvaluationContext is transforming the evaluationContext into a GoFeatureFlagUser
+     *
+     * @param ctx - EvaluationContext from open-feature
+     * @return GoFeatureFlagUser format for GO Feature Flag
+     */
+    public static GoFeatureFlagUser fromEvaluationContext(EvaluationContext ctx) {
+        String key = ctx.getTargetingKey();
+        Value anonymousValue = ctx.getValue("anonymous");
+        boolean anonymous = anonymousValue != null && anonymousValue.isBoolean() ? anonymousValue.asBoolean() : false;
+        Map<String, Object> custom = ctx.asObjectMap();
```
toddbaert: I see how this was useful to you!

providers/go-feature-flag/src/main/java/org/gofeatureflag/provider/bean/GoFeatureFlagUser.java
```diff
@@ -0,0 +1,36 @@
...
+
+    /**
+     * fromEvaluationContext is transforming the evaluationContext into a GoFeatureFlagUser
+     *
+     * @param ctx - EvaluationContext from open-feature
+     * @return GoFeatureFlagUser format for GO Feature Flag
+     */
+    public static GoFeatureFlagUser fromEvaluationContext(EvaluationContext ctx) {
+        String key = ctx.getTargetingKey();
+        Value anonymousValue = ctx.getValue("anonymous");
```
toddbaert: "anonymous" is used many times, you might make it a constant.

providers/go-feature-flag/src/main/java/org/gofeatureflag/provider/GoFeatureFlagProvider.java
```diff
@@ -0,0 +1,246 @@
...
+            return new Value((Structure) object);
+        } else if (object instanceof List) {
+            // need to translate each elem in list to a value
+            return new Value(((List<Object>) object).stream().map(this::objectToValue).collect(Collectors.toList()));
+        } else if (object instanceof Instant) {
+            return new Value((Instant) object);
+        } else if (object instanceof Map) {
+            return new Value(mapToStructure((Map<String, Object>) object));
+        } else {
+            throw new ClassCastException("Could not cast Object to Value");
```
toddbaert: Do you want to handle `null` values (especially in inner keys)? Since `null` is a valid JSON  and that is your content type, I think it should be allowed?

toddbaert (changes requested):
I left a few minor comments for you to think about, but this all looks pretty good to me!

You will need to add this to the main pom.xml's `modules` element though, it's not currently actually building:

```
[INFO] Reactor Summary:
[INFO] 
[INFO] java-sdk-contrib 0.0.2 ............................. SUCCESS [  4.806 s]
[INFO] open-telemetry-hook 0.2.8 .......................... SUCCESS [ 14.724 s]
[INFO] flagd 0.3.2 ........................................ SUCCESS [ 25.281 s]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
```

https://github.com/open-feature/java-sdk-contrib/blob/bc8e3ce5f4011bba43d8fe32d5471d96c8bbe6c4/pom.xml#L29-L32

providers/go-feature-flag/src/main/java/org/gofeatureflag/provider/GoFeatureFlagProvider.java
thomaspoignant: Yes it always hitting the same domain, I will have proper configuration there you are right.

providers/go-feature-flag/src/main/java/org/gofeatureflag/provider/GoFeatureFlagProvider.java
thomaspoignant: GO Feature Flag is not sending `null` value, but I have added the test case to remove the key if we receive a null.

toddbaert (approved):
Very nice!

toddbaert:
@thomaspoignant dependency is available now! Everything looks good to me except a few lint issues.

thomaspoignant:
@toddbaert I've finally finished making `lombok` work with `findbugs`.
The actual configuration does not really like it 😉 

Can you have a last check? It is ready to merge on my side.

toddbaert (approved):

beeme1mr (approved):

PR #116, opened 2022-10-06
feat!: update to sdk 0.3.0

toddbaert:
Absorbs most recent SDK changes. Pretty easy changes for the most part.

### DIFF
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
@@ -5,8 +5,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
 
-import org.apache.commons.lang3.EnumUtils;
-
 import dev.openfeature.flagd.grpc.Schema.ResolveBooleanRequest;
 import dev.openfeature.flagd.grpc.Schema.ResolveBooleanResponse;
 import dev.openfeature.flagd.grpc.Schema.ResolveFloatRequest;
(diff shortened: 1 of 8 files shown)


toddbaert:
@thomaspoignant please take a look at the goff changes. Similar to Node, they are mostly about reason/error code. If this doesn't make sense, let me know, or push changes!

Note: this is to absorb the UNRELEASED Java changes.

providers/go-feature-flag/src/main/java/dev/openfeature/contrib/providers/gofeatureflag/GoFeatureFlagProvider.java
```diff
@@ -215,8 +218,7 @@ private <T> ProviderEvaluation<T> resolveEvaluationGoFeatureFlagProxy(
                 }
 
                 return ProviderEvaluation.<T>builder()
-                        .errorCode(goffResp.getErrorCode())
```
toddbaert, line 218: I removed the error code here. I think that if we got this far, we don't need the error? Otherwise we should probably throw.

I think this is the biggest GOFF change I made.

@thomaspoignant

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
```diff
@@ -155,19 +152,10 @@ public ProviderEvaluation<Value> getObjectEvaluation(String key, Value defaultVa
...
 
-    // Map FlagD reasons to Java SDK reasons.
-    private Reason mapReason(String flagdReason) {
-        if (!EnumUtils.isValidEnum(Reason.class, flagdReason)) {
-            return Reason.UNKNOWN;
-        } else {
-            return Reason.valueOf(flagdReason);
-        }
-    }
-
```
toddbaert, line 170: reasons are strings now.

beeme1mr (approved):

thomaspoignant (approved):
Sorry for the late review.
This is actually perfect, thanks for the change 👌

PR #127, opened 2022-11-15
chore: use JDK 8 in ci

toddbaert:
Even with compiler source and target set to 1.8, it's possible that APIs only available in later runtimes are referenced (see [here](https://maven.apache.org/plugins/maven-compiler-plugin/examples/set-compiler-source-and-target.html) ):

> Merely setting the target option does not guarantee that your code actually runs on a JRE with the specified version. The pitfall is unintended usage of APIs that only exist in later JREs which would make your code fail at runtime with a linkage error. 

I became especially concerned with this with regards to the generated gRPC sources.

I've configured the CI to use JDK 8 for build to prevent this, though 18 and other less-archaic versions will work locally without problem to avoid annoyance.

I also moved to the same checkstyle as the JS-SDK.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -154,7 +154,7 @@
                     <dependency>
                         <groupId>com.puppycrawl.tools</groupId>
                         <artifactId>checkstyle</artifactId>
-                        <version>10.3.2</version>
+                        <version>8.45.1</version>
                     </dependency>
                 </dependencies>
                 <executions>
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -154,7 +154,7 @@
                     <dependency>
                         <groupId>com.puppycrawl.tools</groupId>
                         <artifactId>checkstyle</artifactId>
-                        <version>10.3.2</version>
+                        <version>8.45.1</version>
```
toddbaert, line 157: Had to downgrade these to versions compatible with 1.8.

justinabrahms (approved):

PR #167, opened 2022-12-15
feat: flagsmith provider

ajhelsby:
## This PR
This PR adds a flagsmith provider to [java-sdk-contrib](https://github.com/open-feature/java-sdk-contrib) with the following attributes:
- New Flagsmith Provider
- Documentation on how to use the Flagsmith provider
- Implementation of boolean, string, number and object based flag evaluation
- Packaging of the new Flagsmith module to mvn repository

### Related Issues
N/A

### Notes
N/A

### Follow-up Tasks
N/A

### How to test
Unit tests have been included for the new Flagsmith module


providers/flagsmith/pom.xml
```diff
@@ -0,0 +1,52 @@
...
+            <url>https://flagsmith.com</url>
+        </developer>
+    </developers>
+
+    <dependencies>
+        <!-- Inherits dev.openfeature.javasdk and the test dependencies from the parent pom -->
+        <dependency>
+            <groupId>com.flagsmith</groupId>
+            <artifactId>flagsmith-java-client</artifactId>
+            <version>5.0.0</version>
```
matthewelwell: This should be [5.1.1](https://mvnrepository.com/artifact/com.flagsmith/flagsmith-java-client/5.1.1)

providers/flagsmith/src/main/java/dev.openfeature.contrib.providers.flagsmith/FlagsmithClientConfigurer.java
```diff
@@ -0,0 +1,157 @@
...
+ */
+public class FlagsmithClientConfigurer {
+
+    /**
+     * Check the options that have been provided to see if there are any issues.
+     * Exceptions will be thrown if there are issues found with the options.
+     *
+     * @param options the options used to create the provider
+     */
+    public static void validateOptions(FlagsmithProviderOptions options) {
```
matthewelwell: Couldn't this be done on the FlagsmithProviderOptions class itself? Or perhaps this should just be a private method called by the initializeProvider method?

providers/flagsmith/src/main/java/dev.openfeature.contrib.providers.flagsmith/FlagsmithProvider.java
```diff
@@ -0,0 +1,232 @@
...
+    ) throws OpenFeatureError {
+        T flagValue = null;
+        ErrorCode errorCode = null;
+        Reason reason = null;
+        String variationType = "";
+        try {
+
+            Flags flags = Objects.isNull(ctx.getTargetingKey()) || ctx.getTargetingKey().isEmpty()
+                ? flagsmith.getEnvironmentFlags()
+                : flagsmith.getIdentityFlags(ctx.getTargetingKey());
```
matthewelwell: Is it worth adding a todo here to add traits once the attributes are added to the context?

providers/flagsmith/src/main/java/dev.openfeature.contrib.providers.flagsmith/FlagsmithProvider.java
```diff
@@ -0,0 +1,232 @@
...
+            // Check if the flag is enabled, return default value if not
```
matthewelwell, line 133: I'm not sure why we'd return a default value if the flag is not enabled? 'off' is a valid state, no? I would suggest that we should only return a default value if the API is down or the feature doesn't exist.

providers/flagsmith/src/main/java/dev.openfeature.contrib.providers.flagsmith/FlagsmithProvider.java
```diff
@@ -0,0 +1,232 @@
...
+                                         .reason(Reason.DISABLED.name())
+                                         .build();
+            }
+
+            Object value = flags.getFeatureValue(key);
+            // Convert the value received from Flagsmith.
+            flagValue = convertValue(value, expectedType);
+
+            if (flagValue.getClass() != expectedType) {
+                throw new TypeMismatchError("Flag value " + key + " had unexpected type "
```
matthewelwell: Does this get caught somewhere else? Or is it ok to throw a subclass of OpenFeatureError?

providers/flagsmith/src/main/java/dev.openfeature.contrib.providers.flagsmith/FlagsmithProvider.java
```diff
@@ -0,0 +1,232 @@
...
+                    + flagValue.getClass() + ", expected " + expectedType + ".");
+            }
+
+        } catch (FlagsmithApiError flagsmithApiError) {
+            flagValue = defaultValue;
+            reason = Reason.ERROR;
+            errorCode = ErrorCode.PARSE_ERROR;
```
matthewelwell: Not sure why this is a PARSE_ERROR?

providers/flagsmith/src/main/java/dev.openfeature.contrib.providers.flagsmith/FlagsmithProvider.java
```diff
@@ -0,0 +1,232 @@
...
+            // need to translate each elem in list to a value
+            return new Value(((List<Object>) object).stream()
+                                                    .map(this::objectToValue)
+                                                    .collect(Collectors.toList()));
+        } else if (object instanceof Instant) {
+            return new Value((Instant) object);
+        } else if (object instanceof Map) {
+            return new Value(mapToStructure((Map<String, Object>) object));
+        } else {
+            throw new TypeMismatchError("Flag value " + object + " had unexpected type "
```
matthewelwell, line 239: seems like we're throwing this in the method above as well, do we need it in both places?

providers/flagsmith/src/test/java/dev.openfeature.contrib.providers.flagsmith/FlagsmithProviderTest.java
```diff
@@ -0,0 +1,105 @@
...
+        result = method.invoke(flagsmithProvider, key, null, evaluationContext);
+
+        // Then
+        ProviderEvaluation<Object> evaluation = (ProviderEvaluation<Object>) result;
+        assertEquals(flagsmithResult, evaluation.getValue());
+        assertNull(evaluation.getErrorCode());
+    }
+
+    @Test
+    void shouldNotResolveFlagIfFlagIsInactiveInFlagsmithInsteadUsingDefaultValue() {
```
matthewelwell: As above, I'm not sure why this is the case?

providers/flagsmith/src/test/java/dev.openfeature.contrib.providers.flagsmith/FlagsmithProviderTest.java
```diff
@@ -0,0 +1,105 @@
...
+
+    public static MockWebServer mockFlagsmithClient;
+
+    final QueueDispatcher dispatcher = new QueueDispatcher() {
+        @SneakyThrows
+        @Override
+        public MockResponse dispatch(RecordedRequest request) {
+            if (request.getPath().startsWith("/flags/")) {
+                return new MockResponse()
+                    .setBody(readMockResponse("environment_flags_response.json"))
```
matthewelwell: This seems wrong to me - the response json you have below looks like a response to the `environment-document` endpoint, yet you're binding it to the `flags` endpoint?

ajhelsby: Yes agreed, it's been changed to the flags endpoint result

providers/flagsmith/pom.xml
ajhelsby: Updated to the latest

providers/flagsmith/src/main/java/dev.openfeature.contrib.providers.flagsmith/FlagsmithClientConfigurer.java
ajhelsby: It's been put into a configurer class to stop the model class becoming too bloated. It's already being called by the constructor of the provider.

providers/flagsmith/src/main/java/dev.openfeature.contrib.providers.flagsmith/FlagsmithProvider.java
ajhelsby: This is the behaviour defined by the spec. We return the default and set the reason to disabled.

providers/flagsmith/src/main/java/dev.openfeature.contrib.providers.flagsmith/FlagsmithProvider.java
ajhelsby: This gets caught directly in the java-sdk

providers/flagsmith/src/main/java/dev.openfeature.contrib.providers.flagsmith/FlagsmithProvider.java
ajhelsby: Agreed it's been changed to a GeneralError

providers/flagsmith/src/test/java/dev.openfeature.contrib.providers.flagsmith/FlagsmithProviderTest.java
ajhelsby: This is the behaviour according to the spec

providers/flagsmith/src/main/java/dev.openfeature.contrib.providers.flagsmith/FlagsmithProvider.java
ajhelsby: This error is thrown when the object type is unknown wheras the other is thrown when the type does not match the one expected.

providers/flagsmith/src/main/java/dev.openfeature.contrib.providers.flagsmith/FlagsmithProvider.java
ajhelsby: Added a todo

providers/flagsmith/README.md
```diff
@@ -0,0 +1,54 @@
...
+## Usage
+
+The `FlagsmithProvider` communicates with Flagsmith using the Flagsmith java sdk. Information on the 
+sdk can be found in the Flagsmith documentaiton here https://docs.flagsmith.com/clients/server-side. 
+The following code snippet shows how to initialize the `FlagsmithProvider`:
+```java
+FlagsmithProviderOptions options = FlagsmithProviderOptions.builder()
+                                                           .apiKey("API_KEY")
+                                                           .build();
+FlagsmithProvider provider = new FlagdProvider();
```
matthewelwell:
```suggestion
FlagsmithProvider provider = new FlagsmithProvider();
```

providers/flagsmith/pom.xml
matthewelwell: Doesn't look like this has been done?

providers/flagsmith/README.md
```diff
@@ -0,0 +1,54 @@
...
+| Option name | Type    | Default   | Description
+| ----------- | ------- | --------- | ---------
+| apiKey      | String  |  | Your API Token. Note that this is either the `Environment API` key or the `Server Side SDK Token`
+| headers      | HashMap<String, String>  |  | Add custom headers which will be sent with each network request to the Flagsmith API.
+| envFlagsCacheKey      | String  |  | Enable in-memory caching for the Flagsmith API.
+| expireCacheAfterWriteTimeUnit      | TimeUnit  |  | 
+| expireCacheAfterWrite      | int  |  | 
+| expireCacheAfterAccessTimeUnit      | TimeUnit  |  | 
+| expireCacheAfterAccess      | int  |  | 
+| maxCacheSize      | int  |  |
```
Kavindu-Dodan: It is good to have descriptions for these options as well.

providers/flagsmith/src/main/java/dev.openfeature.contrib.providers.flagsmith/FlagsmithClientConfigurer.java
```diff
@@ -0,0 +1,167 @@
...
+        }
+
+        if (options.getExpireCacheAfterAccess() != 0
+            && options.getExpireCacheAfterAccessTimeUnit() != null) {
+            flagsmithCacheConfig.expireAfterAccess(
+                options.getExpireCacheAfterAccess(),
+                options.getExpireCacheAfterAccessTimeUnit());
+        }
+
+        if (options.getMaxCacheSize() != 0) {
```
Kavindu-Dodan: The default is zero and comes from `FlagsmithProviderOptions` (primitive default). Should we allow a zero cache ? IMO this is misleading and a non-zero default makes no sense

providers/flagsmith/src/main/java/dev.openfeature.contrib.providers.flagsmith/FlagsmithProvider.java
```diff
@@ -0,0 +1,284 @@
...
+    }
+
+    /**
+     * The method objectToValue is wrapping an object into a Value.
+     *
+     * @param object the object you want to wrap
+     * @return the wrapped object
+     */
+    @SneakyThrows
+    private Value objectToValue(Object object) {
```
Kavindu-Dodan, line 211: @toddbaert  do you see such generic conversion can be exposed from `dev.openfeature.sdk.Value` class itself? 🤔

providers/flagsmith/README.md
ajhelsby: Added descriptions to the cache elements

providers/flagsmith/src/main/java/dev.openfeature.contrib.providers.flagsmith/FlagsmithClientConfigurer.java
```diff
@@ -0,0 +1,167 @@
...
+ */
+public class FlagsmithClientConfigurer {
+
+    /**
+     * Check the options that have been provided to see if there are any issues.
+     * Exceptions will be thrown if there are issues found with the options.
+     *
+     * @param options the options used to create the provider
+     */
+    public static void validateOptions(FlagsmithProviderOptions options) {
```
p-maks: do we need this validation here, why do we need them as separate method? should flagmisth sdk validate them when initialising client?
if you still want to use it make it private and call it from initializeProvider()?

providers/flagsmith/src/main/java/dev.openfeature.contrib.providers.flagsmith/FlagsmithClientConfigurer.java
```diff
@@ -0,0 +1,167 @@
...
+        if (options == null) {
+            throw new InvalidOptions("No options provided");
+        }
+
+        if (options.getApiKey() == null || options.getApiKey().isEmpty()) {
+            throw new InvalidOptions("Flagsmith API key has not been set.");
```
p-maks: Exception suffix?

providers/flagsmith/src/main/java/dev.openfeature.contrib.providers.flagsmith/FlagsmithProvider.java
```diff
@@ -0,0 +1,308 @@
...
+     * @param expectedType the type we expect for this value
+     * @param <T>          the type we want to convert to
+     * @return A converted object
+     */
+    private <T> T convertValue(Object value, Class<?> expectedType) {
+        boolean isPrimitive = expectedType == Boolean.class
+            || expectedType == String.class
+            || expectedType == Integer.class
+            || expectedType == Double.class;
+        T flagValue;
```
p-maks: use short version `T flagValue = isPrimitive ? (T) value : (T) objectToValue(value)`

providers/flagsmith/src/main/java/dev/openfeature/contrib/providers/flagsmith/exceptions/InvalidCacheOptions.java
```diff
@@ -0,0 +1,18 @@
+package dev.openfeature.contrib.providers.flagsmith.exceptions;
+
+import dev.openfeature.sdk.ErrorCode;
+import lombok.Getter;
+
+/**
+ * InvalidOptions is the super Exception used when we have a configuration exception.
+ */
+@Getter
+public class InvalidCacheOptions extends FlagsmithProviderException {
```
p-maks: add `Exception` suffix

providers/flagsmith/src/main/java/dev/openfeature/contrib/providers/flagsmith/exceptions/InvalidOptions.java
```diff
@@ -0,0 +1,17 @@
+package dev.openfeature.contrib.providers.flagsmith.exceptions;
+
+import dev.openfeature.sdk.ErrorCode;
+import lombok.Getter;
+
+/**
+ * InvalidOptions is the super Exception used when we have a configuration exception.
+ */
+@Getter
+public class InvalidOptions extends FlagsmithProviderException {
```
p-maks: add `Exception` suffix

p-maks (commented):
some minor suggestions

providers/flagsmith/src/main/java/dev.openfeature.contrib.providers.flagsmith/FlagsmithClientConfigurer.java
ajhelsby: Done

providers/flagsmith/src/main/java/dev/openfeature/contrib/providers/flagsmith/exceptions/InvalidCacheOptions.java
ajhelsby: Done

providers/flagsmith/src/main/java/dev/openfeature/contrib/providers/flagsmith/exceptions/InvalidOptions.java
ajhelsby: Done

providers/flagsmith/src/main/java/dev.openfeature.contrib.providers.flagsmith/FlagsmithProvider.java
ajhelsby: Done

providers/flagsmith/src/main/java/dev.openfeature.contrib.providers.flagsmith/FlagsmithClientConfigurer.java
ajhelsby: Agreed non required defaults have been set to -1

providers/flagsmith/src/main/java/dev.openfeature.contrib.providers.flagsmith/FlagsmithClientConfigurer.java
```diff
@@ -0,0 +1,169 @@
...
+ */
+public class FlagsmithClientConfigurer {
+
+    /**
+     * Check the options that have been provided to see if there are any issues.
+     * Exceptions will be thrown if there are issues found with the options.
+     *
+     * @param options the options used to create the provider
+     */
+    private static void validateOptions(FlagsmithProviderOptions options) {
```
matthewelwell: Being picky, I prefer to see public methods above all private methods.

providers/flagsmith/src/main/java/dev/openfeature/contrib/providers/flagsmith/exceptions/FlagsmithJsonException.java
```diff
@@ -0,0 +1,19 @@
+package dev.openfeature.contrib.providers.flagsmith.exceptions;
+
+import dev.openfeature.sdk.ErrorCode;
+import dev.openfeature.sdk.exceptions.GeneralError;
+import lombok.Getter;
+
+/**
+ * A Flagsmith provider exception is the main exception for the provider.
```
matthewelwell: Think this is a copypasta error?

providers/flagsmith/src/main/java/dev/openfeature/contrib/providers/flagsmith/exceptions/InvalidCacheOptionsExceptions.java
```diff
@@ -0,0 +1,18 @@
+package dev.openfeature.contrib.providers.flagsmith.exceptions;
+
+import dev.openfeature.sdk.ErrorCode;
+import lombok.Getter;
+
+/**
+ * InvalidOptions is the super Exception used when we have a configuration exception.
```
matthewelwell: Perhaps needs to be updated to reflect the correct class name?

providers/flagsmith/src/main/java/dev/openfeature/contrib/providers/flagsmith/exceptions/InvalidOptionsExceptions.java
```diff
@@ -0,0 +1,17 @@
+package dev.openfeature.contrib.providers.flagsmith.exceptions;
+
+import dev.openfeature.sdk.ErrorCode;
+import lombok.Getter;
+
+/**
+ * InvalidOptions is the super Exception used when we have a configuration exception.
```
matthewelwell:
```suggestion
 * InvalidOptionsException is the super Exception used when we have a configuration exception.
```

providers/flagsmith/src/main/java/dev/openfeature/contrib/providers/flagsmith/exceptions/InvalidOptionsExceptions.java
```diff
@@ -0,0 +1,17 @@
...
+ */
+@Getter
+public class InvalidOptionsExceptions extends FlagsmithProviderException {
```
matthewelwell: Is there a reason this is plural?

providers/flagsmith/src/test/java/dev.openfeature.contrib.providers.flagsmith/FlagsmithProviderTest.java
```diff
@@ -0,0 +1,340 @@
...
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+public class FlagsmithProviderTest {
+
+    public static MockWebServer mockFlagsmithClient;
+    public static FlagsmithProvider flagsmithProvider;
+
+    final QueueDispatcher dispatcher = new QueueDispatcher() {
+        @SneakyThrows
```
matthewelwell, line 48: I don't know what this annotation does but I like it! 😄

providers/flagsmith/src/test/java/dev.openfeature.contrib.providers.flagsmith/FlagsmithProviderTest.java
```diff
@@ -0,0 +1,340 @@
...
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+public class FlagsmithProviderTest {
+
+    public static MockWebServer mockFlagsmithClient;
```
matthewelwell: Should this be called `mockWebServer` rather than `mockFlagsmithClient` - `mockFlagsmithClient` to me indicates that we're mocking the `FlagsmithClient` class rather than any HTTP interactions (which I believe is what we're actually doing).

providers/flagsmith/src/test/java/dev.openfeature.contrib.providers.flagsmith/FlagsmithProviderTest.java
```diff
@@ -0,0 +1,340 @@
...
+                .of(FlagsmithProviderOptions.builder().apiKey("API_KEY").maxCacheSize(1).build()),
+            Arguments
+                .of(FlagsmithProviderOptions.builder().apiKey("API_KEY").expireCacheAfterWrite(1)
+                                            .build()),
+            Arguments.of(FlagsmithProviderOptions.builder().apiKey("API_KEY").recordCacheStats(true)
+                                                 .build())
+        );
+    }
+
+    @BeforeEach
```
matthewelwell, line 126: Seems like this BeforeEach isn't _actually_ used in all the tests - can we perhaps use test groups or something for this?

providers/flagsmith/src/main/java/dev/openfeature/contrib/providers/flagsmith/exceptions/InvalidOptionsExceptions.java
ajhelsby: No mispelling

providers/flagsmith/src/main/java/dev.openfeature.contrib.providers.flagsmith/FlagsmithProvider.java
```diff
@@ -0,0 +1,302 @@
...
+    public ProviderEvaluation<Boolean> getBooleanEvaluation(
+        String key, Boolean defaultValue, EvaluationContext evaluationContext) {
+        // When isUsingBooleanConfigValue the feature_state_value will be used as the flag value
+        if (this.options.isUsingBooleanConfigValue()) {
+            return resolveFlagsmithEvaluation(key, defaultValue, evaluationContext, Boolean.class);
+        }
+        // Otherwise the enabled field will be used as the boolean flag value
+        try {
+            Flags flags = getFlags(evaluationContext);
+            //Todo once sdk has been updated to 5.1.2 these checks won't be relevant
```
matthewelwell: 5.1.2 is released now

providers/flagsmith/src/main/java/dev.openfeature.contrib.providers.flagsmith/FlagsmithProvider.java
toddbaert: We do have some functionality in the SDK to help with _some_ stuff like this (see [here](https://github.com/open-feature/java-sdk/blob/main/src/main/java/dev/openfeature/sdk/MutableStructure.java#L92-L111)) but this is really going the other direction (JSON impl -> OpenFeature Value vs OpenFeature Value -> JSON impl).

The challenge that you can see here is this sort of code often has to convert a _specific_ JSON implementation (in this case, Jackson) which we _cannot_ directly reference in the SDK. I'm fine with including java/jakarta refs, but I don't think we'd want to include any implementations like jackson in the SDK. If you think there's a way to ease this for developers without adding those deps, that would be a very powerful addition to the Java SDK.

providers/flagsmith/src/main/java/dev.openfeature.contrib.providers.flagsmith/FlagsmithClientConfigurer.java
```diff
@@ -0,0 +1,169 @@
...
+     * @param options the options used to create the provider
+     */
+    static FlagsmithClient initializeProvider(FlagsmithProviderOptions options) {
+
+        validateOptions(options);
+
+        FlagsmithClient.Builder flagsmithBuilder = FlagsmithClient
+            .newBuilder();
+        // Set main configuration settings
+        if (options.getApiKey() != null) {
```
toddbaert: This check seems a little redundent with the `validateOptions` above

providers/flagsmith/src/main/java/dev.openfeature.contrib.providers.flagsmith/FlagsmithProvider.java
```diff
@@ -0,0 +1,263 @@
...
+
+    @Override
+    public Metadata getMetadata() {
+        return () -> NAME;
+    }
+
+    @Override
+    public List<Hook> getProviderHooks() {
+        return FeatureProvider.super.getProviderHooks();
+    }
```
toddbaert: Is there a reason to override the default impl just to call it?

providers/flagsmith/README.md
```diff
@@ -0,0 +1,55 @@
...
+| writeTimeout      | int  | 5000 | The network timeout in milliseconds when writing.
+| readTimeout      | int  | 5000 | The network timeout in milliseconds when reading.
+| sslSocketFactory      | SSLSocketFactory  |  | Override the sslSocketFactory.
+| trustManager      | X509TrustManager  |  | X509TrustManager used when overriding the sslSocketFactory.
+| httpInterceptor      | Interceptor  |  | Add a custom HTTP interceptor in the form of an okhttp3.Interceptor object.
+| retries      | int  | 3 | Add a custom com.flagsmith.config.Retry object to configure the backoff / retry configuration.
+| localEvaluation      | boolean  | false | Controls which mode to run in; local or remote evaluation.
+| environmentRefreshIntervalSeconds      | int  | 60 | Set environment refresh rate with polling manager.
+| enableAnalytics      | boolean  | false | Controls whether Flag Analytics data is sent to the Flagsmith API
+| usingBooleanConfigValue      | boolean  | false | Determines whether to resolve a feature value as a boolean or use the isFeatureEnabled as the flag itself. These values will be false and true respectively.
```
toddbaert, line 60: It seems like you did end up implementing this sort of switch! I think it's a good idea.

providers/flagsmith/src/main/java/dev.openfeature.contrib.providers.flagsmith/FlagsmithProvider.java
```diff
@@ -0,0 +1,263 @@
...
+            if (!isFlagEnabled) {
+                return buildEvaluation(defaultValue, null, Reason.DISABLED, null);
+            }
+
+            Object value = flags.getFeatureValue(key);
+            // Convert the value received from Flagsmith.
+            flagValue = convertValue(value, expectedType);
+
+        } catch (FlagsmithClientError flagsmithApiError) {
+            return buildEvaluation(defaultValue, ErrorCode.GENERAL, Reason.ERROR, null);
```
toddbaert: You might be able to do some basic mapping between HTTP codes (if they are available in `FlagsmithClientError`) and well-defined error states such as `ErrorCode.FLAG_NOT_FOUND`, but I leave that up to you!

providers/flagsmith/src/main/java/dev.openfeature.contrib.providers.flagsmith/FlagsmithProvider.java
```diff
@@ -0,0 +1,263 @@
...
+        }
```
toddbaert: You may want to consider simply throwing here instead of returning an evaluation. The SDK will handle setting the error and returning the default according to which `OpenFeatureError` impl you throw.

providers/flagsmith/src/main/java/dev.openfeature.contrib.providers.flagsmith/FlagsmithProvider.java
```diff
@@ -0,0 +1,263 @@
...
+    public ProviderEvaluation<Value> getObjectEvaluation(
+        String key, Value defaultValue, EvaluationContext evaluationContext) {
+        return resolveFlagsmithEvaluation(key, defaultValue, evaluationContext, Value.class);
+    }
+
+    /**
+     * Get all the flags for a given environment or identity. The Method will use
+     * getEnvironmentFlags from the Flagsmith sdk if no targeting key is provided
+     * in the EvaluationContext. If a targeting key is provided then the
+     * getIdentityFlags method will be used.
```
toddbaert, line 96: I think this is fine, but you should mention this in your doc IMHO.

You also _may_ want to warn if a context attr is set without setting a targeting key, since it would have no effect in this case if I understand correctly.

providers/flagsmith/src/test/java/dev.openfeature.contrib.providers.flagsmith/FlagsmithProviderTest.java
```diff
@@ -0,0 +1,409 @@
+package dev.openfeature.contrib.providers.flagsmith;
```
toddbaert, line 1: Tests seem good to me! Testing the provider directly is definitely good for testing it's configuration and setup.

In regards to testing the overall reliability of the provider, I would recommend you convert some of these to "integration style" tests wherein you actually register the provider with openfeature and evaluate from there using a client instance. This way you're actually testing the provider's adherence to the contract a little more comprehensively.

I hope that makes sense.

providers/flagsmith/src/main/java/dev.openfeature.contrib.providers.flagsmith/FlagsmithProvider.java
```diff
@@ -0,0 +1,263 @@
...
+     *
+     * @param ctx an EvaluationContext object with flag evaluation options
+     * @return a Flagsmith Flags object with all the respective flags
+     * @throws FlagsmithClientError Thrown when there are issue retrieving the flags
+     *                              from Flagsmith
+     */
+    private Flags getFlags(EvaluationContext ctx) throws FlagsmithClientError {
+        return Objects.isNull(ctx.getTargetingKey()) || ctx.getTargetingKey().isEmpty()
+            ? flagsmith.getEnvironmentFlags()
+            // Todo add traits when attributes are added to context
```
toddbaert: I think I would recommend we implement traits before we release this, or context functionality won't work as expected. Let me know if I've misunderstood.

toddbaert (changes requested):
This all looks really good! I think the only thing I'd consider a blocker is [this](https://github.com/open-feature/java-sdk-contrib/pull/167/files#r1054726112), which feels to me like a minimum release requirement. Please let me know if you disagree or I'm misunderstanding.

providers/flagsmith/src/main/java/dev.openfeature.contrib.providers.flagsmith/FlagsmithClientConfigurer.java
ajhelsby: Moved below the public methods

providers/flagsmith/src/main/java/dev/openfeature/contrib/providers/flagsmith/exceptions/FlagsmithJsonException.java
ajhelsby: Yes indeed, fixed now

providers/flagsmith/src/main/java/dev/openfeature/contrib/providers/flagsmith/exceptions/InvalidOptionsExceptions.java
ajhelsby: Fixed

providers/flagsmith/src/test/java/dev.openfeature.contrib.providers.flagsmith/FlagsmithProviderTest.java
ajhelsby: I want to keep the word Flagsmith in it, but agreed client clashes with the name in the sdk, so lets go with mockFlagsmithServer

providers/flagsmith/src/main/java/dev.openfeature.contrib.providers.flagsmith/FlagsmithProvider.java
ajhelsby: Updated

providers/flagsmith/src/main/java/dev.openfeature.contrib.providers.flagsmith/FlagsmithProvider.java
ajhelsby: Unfortunately there are not at the moment, we are looking into adding some more bespoke error codes on the Flagsmith SDK side that would allow us to do this

providers/flagsmith/src/main/java/dev/openfeature/contrib/providers/flagsmith/exceptions/InvalidCacheOptionsExceptions.java
ajhelsby: Done

providers/flagsmith/src/main/java/dev.openfeature.contrib.providers.flagsmith/FlagsmithProvider.java
ajhelsby: Switched it to throwing the exceptions now instead

providers/flagsmith/README.md
ajhelsby: 👍🏻

providers/flagsmith/src/main/java/dev.openfeature.contrib.providers.flagsmith/FlagsmithProvider.java
ajhelsby: Not really, just noticed it had been done in the go-feature-flag provider, I've removed it ow, since it's not needed

providers/flagsmith/src/main/java/dev.openfeature.contrib.providers.flagsmith/FlagsmithClientConfigurer.java
ajhelsby: Agreed, it's been removed

providers/flagsmith/src/main/java/dev.openfeature.contrib.providers.flagsmith/FlagsmithProvider.java
ajhelsby: We've managed to implement this now

providers/flagsmith/src/main/java/dev.openfeature.contrib.providers.flagsmith/FlagsmithProvider.java
ajhelsby: I've added a description on how to do this in the readme

release-please-config.json
```diff
@@ -12,6 +12,16 @@
...
       ]
     },
+    "providers/flagsmith": {
+      "package-name": "dev.openfeature.contrib.providers.flagsmith",
+      "release-type": "simple",
+      "bump-minor-pre-major": true,
+      "bump-patch-for-minor-pre-major": true,
+      "versioning": "default",
+      "extra-files": [
+        "pom.xml"
```
beeme1mr: You'll also need to include `README.md` to the `extra-files` list. This will configure Release Please to automatically bump the version in the readme as part of a release PR.

toddbaert:
I will re-review this tomorrow!

toddbaert (approved):
LGTM, @ajhelsby. I've added a couple very small changes (added readme to release-plz files and added you as a CODEOWNER for this dir) and rebased on `main` to update. I hope that's OK!

I think this is ready to merge now - I will do so by EOD unless I hear otherwise.

release-please-config.json
toddbaert: I did this.

providers/flagsmith/src/test/java/dev.openfeature.contrib.providers.flagsmith/FlagsmithProviderTest.java
toddbaert: This isn't a blocker.

PR #168, opened 2022-12-16
feat: flagd caching

### LINKED ISSUE #139: [flagd-provider] add caching
Implement caching based on the strategy defined [here](https://github.com/open-feature/flagd/blob/main/docs/caching.md).

Go implementation, for reference: https://github.com/open-feature/go-sdk-contrib/pull/48

skyerus:
## This PR

Implements caching of flagd.

### Related Issues

Fixes #139

### DIFF
--- a/providers/flagd/pom.xml
+++ b/providers/flagd/pom.xml
@@ -92,7 +92,7 @@
               <goal>exec</goal>
             </goals>
 						<configuration>
-							<!-- run: git submodule update \-\-init \-\-recursive \-\-remote -->
+							<!-- run: git submodule update \-\-init \-\-recursive -->
 							<executable>git</executable>
 							<arguments>
 								<argument>submodule</argument>
--- a/providers/flagd/pom.xml
+++ b/providers/flagd/pom.xml
@@ -99,7 +99,6 @@
 								<argument>update</argument>
 								<argument>--init</argument>
 								<argument>--recursive</argument>
-								<argument>--remote</argument>
 							</arguments>
 						</configuration>
           </execution>
--- a/providers/flagd/pom.xml
+++ b/providers/flagd/pom.xml
@@ -61,6 +61,12 @@
 			<version>6.0.53</version>
 			<scope>provided</scope>
 		</dependency>
+
+		<dependency>
+			<groupId>org.apache.commons</groupId>
+			<artifactId>commons-collections4</artifactId>
+			<version>4.4</version>
+		</dependency>
 	</dependencies>
 
 	<build>
@@ -86,13 +92,14 @@
               <goal>exec</goal>
             </goals>
 						<configuration>
-							<!-- run: git submodule update \-\-init \-\-recursive -->
+							<!-- run: git submodule update \-\-init \-\-recursive \-\-remote -->
 							<executable>git</executable>
 							<arguments>
 								<argument>submodule</argument>
 								<argument>update</argument>
 								<argument>--init</argument>
 								<argument>--recursive</argument>
+								<argument>--remote</argument>
 							</arguments>
 						</configuration>
           </execution>
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/EventStreamCallback.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/EventStreamCallback.java
@@ -1,0 +1,10 @@
+package dev.openfeature.contrib.providers.flagd;
+
+/**
+ * Defines behaviour required of event stream callbacks.
+ */
+public interface EventStreamCallback {
+    void setEventStreamAlive(Boolean alive);
+    
+    void restartEventStream() throws Exception;
+}
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/EventStreamCallback.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/EventStreamCallback.java
@@ -3,8 +3,8 @@ package dev.openfeature.contrib.providers.flagd;
 /**
  * Defines behaviour required of event stream callbacks.
  */
-public interface EventStreamCallback {
+interface EventStreamCallback {
     void setEventStreamAlive(Boolean alive);
-    
+
     void restartEventStream() throws Exception;
 }
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/EventStreamObserver.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/EventStreamObserver.java
@@ -2,6 +2,7 @@ package dev.openfeature.contrib.providers.flagd;
 
 import java.util.Map;
 import io.grpc.stub.StreamObserver;
+import lombok.extern.slf4j.Slf4j;
 import dev.openfeature.flagd.grpc.Schema.EventStreamResponse;
 import dev.openfeature.sdk.ProviderEvaluation;
 import com.google.protobuf.Value;
@@ -9,6 +10,7 @@ import com.google.protobuf.Value;
 /**
  * EventStreamObserver handles events emitted by flagd.
  */
+@Slf4j
 public class EventStreamObserver implements StreamObserver<EventStreamResponse> {
     private EventStreamCallback callback;
     private Boolean cacheEnabled;
@@ -18,9 +20,8 @@ public class EventStreamObserver implements StreamObserver<EventStreamResponse> 
     private Map<String, ProviderEvaluation<Integer>> integerCache;
     private Map<String, ProviderEvaluation<dev.openfeature.sdk.Value>> objectCache;
 
-
-    static final String CONFIGURATION_CHANGE = "configuration_change";
-    static final String PROVIDER_READY = "provider_ready";
+    private static final String configurationChange = "configuration_change";
+    private static final String providerReady = "provider_ready";
 
     EventStreamObserver(
         Boolean cacheEnabled, Map<String, ProviderEvaluation<Boolean>> booleanCache,
@@ -40,27 +41,27 @@ public class EventStreamObserver implements StreamObserver<EventStreamResponse> 
     @Override
     public void onNext(EventStreamResponse value) {
         switch (value.getType()) {
-            case CONFIGURATION_CHANGE:
+            case configurationChange:
                 this.handleConfigurationChangeEvent(value);
                 break;
-            case PROVIDER_READY:
+            case providerReady:
                 this.handleProviderReadyEvent();
                 break;
             default:
-                // log
+                log.warn("unhandled event type {}", value.getType());
                 return;
         }
     }
 
     @Override
     public void onError(Throwable t) {
-        t.printStackTrace();
+        log.error("event stream", t);
         this.purgeCache();
         this.callback.setEventStreamAlive(false);
         try {
             this.callback.restartEventStream();
         } catch (Exception e) {
-            e.printStackTrace();
+            log.error("restart event stream", e);
         }
     }
 
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/EventStreamObserver.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/EventStreamObserver.java
@@ -1,0 +1,101 @@
+package dev.openfeature.contrib.providers.flagd;
+
+import java.util.Map;
+import io.grpc.stub.StreamObserver;
+import dev.openfeature.flagd.grpc.Schema.EventStreamResponse;
+import dev.openfeature.sdk.ProviderEvaluation;
+import com.google.protobuf.Value;
+
+/**
+ * EventStreamObserver handles events emitted by flagd.
+ */
+public class EventStreamObserver implements StreamObserver<EventStreamResponse> {
+    private EventStreamCallback callback;
+    private Boolean cacheEnabled;
+    private Map<String, ProviderEvaluation<Boolean>> booleanCache;
+    private Map<String, ProviderEvaluation<String>> stringCache;
+    private Map<String, ProviderEvaluation<Double>> doubleCache;
+    private Map<String, ProviderEvaluation<Integer>> integerCache;
+    private Map<String, ProviderEvaluation<dev.openfeature.sdk.Value>> objectCache;
+
+
+    static final String CONFIGURATION_CHANGE = "configuration_change";
+    static final String PROVIDER_READY = "provider_ready";
+
+    EventStreamObserver(
+        Boolean cacheEnabled, Map<String, ProviderEvaluation<Boolean>> booleanCache,
+        Map<String, ProviderEvaluation<String>> stringCache, Map<String, ProviderEvaluation<Double>> doubleCache,
+        Map<String, ProviderEvaluation<Integer>> integerCache,
+        Map<String, ProviderEvaluation<dev.openfeature.sdk.Value>> objectCache, EventStreamCallback callback
+    ) {
+        this.cacheEnabled = cacheEnabled;
+        this.booleanCache = booleanCache;
+        this.stringCache = stringCache;
+        this.doubleCache = doubleCache;
+        this.integerCache = integerCache;
+        this.objectCache = objectCache;
+        this.callback = callback;
+    }
+
+    @Override
+    public void onNext(EventStreamResponse value) {
+        switch (value.getType()) {
+            case CONFIGURATION_CHANGE:
+                this.handleConfigurationChangeEvent(value);
+                break;
+            case PROVIDER_READY:
+                this.handleProviderReadyEvent();
+                break;
+            default:
+                // log
+                return;
+        }
+    }
+
+    @Override
+    public void onError(Throwable t) {
+        t.printStackTrace();
+        this.purgeCache();
+        this.callback.setEventStreamAlive(false);
+        try {
+            this.callback.restartEventStream();
+        } catch (Exception e) {
+            e.printStackTrace();
+        }
+    }
+
+    @Override
+    public void onCompleted() {
+        this.purgeCache();
+        this.callback.setEventStreamAlive(false);
+    }
+
+    private void handleConfigurationChangeEvent(EventStreamResponse value) {
+        if (!this.cacheEnabled) {
+            return;
+        }
+
+        Map<String, Value> data = value.getData().getFieldsMap();
+        Value flagKeyValue = data.get("flagKey");
+        String flagKey = flagKeyValue.getStringValue();
+
+        this.booleanCache.remove(flagKey);
+        this.stringCache.remove(flagKey);
+        this.doubleCache.remove(flagKey);
+        this.integerCache.remove(flagKey);
+        this.objectCache.remove(flagKey);
+    }
+
+    private void handleProviderReadyEvent() {
+        this.purgeCache();
+        this.callback.setEventStreamAlive(true);
+    }
+    
+    private void purgeCache() {
+        this.booleanCache.clear();
+        this.stringCache.clear();
+        this.doubleCache.clear();
+        this.integerCache.clear();
+        this.objectCache.clear();
+    }
+}
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/EventStreamObserver.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/EventStreamObserver.java
@@ -4,7 +4,6 @@ import java.util.Map;
 import io.grpc.stub.StreamObserver;
 import lombok.extern.slf4j.Slf4j;
 import dev.openfeature.flagd.grpc.Schema.EventStreamResponse;
-import dev.openfeature.sdk.ProviderEvaluation;
 import com.google.protobuf.Value;
 
 /**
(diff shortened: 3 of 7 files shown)


providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/FlagdProviderTest.java
```diff
@@ -79,11 +80,14 @@ public static void init() {
...
         final String path = "/some/path";
 
-        ServiceBlockingStub mockStub = mock(ServiceBlockingStub.class);
+        ServiceBlockingStub mockBlockingStub = mock(ServiceBlockingStub.class);
+        ServiceStub mockStub = mock(ServiceStub.class);
         NettyChannelBuilder mockChannelBuilder = getMockChannelBuilderSocket();
 
         try (MockedStatic<ServiceGrpc> mockStaticService = mockStatic(ServiceGrpc.class)) {
-            mockStaticService.when(() -> ServiceGrpc.newBlockingStub(any(ManagedChannel.class)))
+            mockStaticService.when(() -> ServiceGrpc.newBlockingStub(any())) // why does any(ManagedChannel.class) fail?
```
skyerus: I'm not sure why `any(ManagedChannel.class)` doesn't work here, any ideas @toddbaert @Kavindu-Dodan ?

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/EventStreamObserver.java
```diff
@@ -0,0 +1,101 @@
...
+    private Boolean cacheEnabled;
+    private Map<String, ProviderEvaluation<Boolean>> booleanCache;
+    private Map<String, ProviderEvaluation<String>> stringCache;
+    private Map<String, ProviderEvaluation<Double>> doubleCache;
+    private Map<String, ProviderEvaluation<Integer>> integerCache;
+    private Map<String, ProviderEvaluation<dev.openfeature.sdk.Value>> objectCache;
+
+
+    static final String CONFIGURATION_CHANGE = "configuration_change";
+    static final String PROVIDER_READY = "provider_ready";
```
Kavindu-Dodan: Both could be private if the intention is to use as constants in this class

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/EventStreamObserver.java
```diff
@@ -0,0 +1,101 @@
...
+import java.util.Map;
+import io.grpc.stub.StreamObserver;
+import dev.openfeature.flagd.grpc.Schema.EventStreamResponse;
+import dev.openfeature.sdk.ProviderEvaluation;
+import com.google.protobuf.Value;
+
+/**
+ * EventStreamObserver handles events emitted by flagd.
+ */
+public class EventStreamObserver implements StreamObserver<EventStreamResponse> {
```
Kavindu-Dodan, line 13: Let's annotate with lombok `@Slf4j` (refer - [1] ) .  This will expose a logger which can be accessed with `log.<level>`




[1] https://github.com/open-feature/java-sdk-contrib/blob/main/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java#L41

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/EventStreamObserver.java
```diff
@@ -0,0 +1,101 @@
...
+    public void onNext(EventStreamResponse value) {
+        switch (value.getType()) {
+            case CONFIGURATION_CHANGE:
+                this.handleConfigurationChangeEvent(value);
+                break;
+            case PROVIDER_READY:
+                this.handleProviderReadyEvent();
+                break;
+            default:
+                // log
```
Kavindu-Dodan: Let's use logger as highlighted above :)

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/EventStreamObserver.java
```diff
@@ -0,0 +1,101 @@
...
+
+    @Override
+    public void onError(Throwable t) {
+        t.printStackTrace();
+        this.purgeCache();
+        this.callback.setEventStreamAlive(false);
+        try {
+            this.callback.restartEventStream();
+        } catch (Exception e) {
+            e.printStackTrace();
```
Kavindu-Dodan: Let's log instead of `e.printStackTrace()`. Please check other such usages as well 

ex:- ` log.warn("Unknown event type: " +value.getType(), e)`

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
```diff
@@ -76,18 +139,91 @@ public FlagdProvider(String socketPath) {
...
+        this.eventStreamAlive = false;
+        if (cache != null) {
+            initCache(cache, maxCacheSize);
+        }
+        this.maxEventStreamRetries = maxEventStreamRetries;
+        this.handleEvents();
+    }
+
+    private void initCache(String cache, int maxSize) {
+        switch (cache) {
```
Kavindu-Dodan: You can use fallthrough behavior and avoid the recursive call :) 

Something like below should do, 
```
switch (cache) {
    case DISABLED:
        return;
    case LRU_CACHE:
    default:
        this.booleanCache = Collections.synchronizedMap(
                new LRUMap<String, ProviderEvaluation<Boolean>>(maxSize));
        this.stringCache = Collections.synchronizedMap(new LRUMap<String, ProviderEvaluation<String>>(maxSize));
        this.doubleCache = Collections.synchronizedMap(new LRUMap<String, ProviderEvaluation<Double>>(maxSize));
        this.integerCache = Collections.synchronizedMap(
                new LRUMap<String, ProviderEvaluation<Integer>>(maxSize));
        this.objectCache = Collections.synchronizedMap(new LRUMap<String, ProviderEvaluation<Value>>(maxSize));
        return;
}
  ```

File: providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/FlagdProviderTest.java (hunk 1 above)
Comment (Kavindu-Dodan) [reply providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/FlagdProviderTest.java]: Do you get any error ? 

May be you need to mock from higher level as `newBlockingStub()` expects `io.grpc.Channel`

So `any(Channel.class)` might work.

File: providers/flagd/pom.xml [hunk 7]
```diff
@@ -86,13 +92,14 @@
               <goal>exec</goal>
             </goals>
 						<configuration>
-							<!-- run: git submodule update \-\-init \-\-recursive -->
+							<!-- run: git submodule update \-\-init \-\-recursive \-\-remote -->
```
toddbaert:
```suggestion
							<!-- run: git submodule update \-\-init \-\-recursive -->
```

providers/flagd/pom.xml
```diff
@@ -86,13 +92,14 @@
...
 							<executable>git</executable>
 							<arguments>
 								<argument>submodule</argument>
 								<argument>update</argument>
 								<argument>--init</argument>
 								<argument>--recursive</argument>
+								<argument>--remote</argument>
```
toddbaert:
```suggestion
```

Adding this param make the build non-deterministic. We should only pull the remote intentionally, when we want to update the submodule.

providers/flagd/pom.xml
toddbaert: https://github.com/open-feature/java-sdk-contrib/pull/168/files#r1054820482

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/EventStreamCallback.java
```diff
@@ -0,0 +1,10 @@
+package dev.openfeature.contrib.providers.flagd;
+
+/**
+ * Defines behaviour required of event stream callbacks.
+ */
+public interface EventStreamCallback {
```
toddbaert: Can this be package private?

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/EventStreamObserver.java
```diff
@@ -0,0 +1,101 @@
...
+ * EventStreamObserver handles events emitted by flagd.
+ */
+public class EventStreamObserver implements StreamObserver<EventStreamResponse> {
+    private EventStreamCallback callback;
+    private Boolean cacheEnabled;
+    private Map<String, ProviderEvaluation<Boolean>> booleanCache;
+    private Map<String, ProviderEvaluation<String>> stringCache;
+    private Map<String, ProviderEvaluation<Double>> doubleCache;
+    private Map<String, ProviderEvaluation<Integer>> integerCache;
+    private Map<String, ProviderEvaluation<dev.openfeature.sdk.Value>> objectCache;
```
toddbaert: Is there a reason to maintain 5 caches besides the need to support multiple types?

I think we could have one cache of type `Object` and Handle everything that way, casting as needed since we know the type can't change while in the cache.

providers/flagd/pom.xml
```diff
@@ -61,6 +61,12 @@
 			<version>6.0.53</version>
 			<scope>provided</scope>
 		</dependency>
+
+		<dependency>
+			<groupId>org.apache.commons</groupId>
+			<artifactId>commons-collections4</artifactId>
+			<version>4.4</version>
+		</dependency>
```
toddbaert, line 69: `org.apache.commons` is a safe group to include on the basis of it's popularity, glad you found an LRU impl there.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/EventStreamObserver.java
```diff
@@ -0,0 +1,101 @@
...
+        Map<String, Value> data = value.getData().getFieldsMap();
+        Value flagKeyValue = data.get("flagKey");
+        String flagKey = flagKeyValue.getStringValue();
+
+        this.booleanCache.remove(flagKey);
+        this.stringCache.remove(flagKey);
+        this.doubleCache.remove(flagKey);
+        this.integerCache.remove(flagKey);
+        this.objectCache.remove(flagKey);
+    }
```
toddbaert: This is impacted by https://github.com/open-feature/flagd/issues/238

toddbaert:
@skyerus I see a `!` in the title - is there actually a breaking change here, and if so, could you note it?

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/EventStreamObserver.java
skyerus: Thanks, I've left a comment on the issue

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/EventStreamObserver.java
skyerus: I considered doing the same thing while working on the Go implementation (storing as `interface{}`) but decided against it to avoid the overhead of type casting in the code and at runtime. Would be good to get some other opinions on this as there are definitely pros/cons to each approach
@james-milligan @Kavindu-Dodan @beeme1mr

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/FlagdProviderTest.java
skyerus: Thanks @Kavindu-Dodan, that fixed it!
The following line fails when using `Channel.class` though despite also expecting `io.grpc.Channel`:
```
mockStaticService.when(() -> ServiceGrpc.newStub(any(Channel.class)))
                        .thenReturn(mockStub);
```
```
java.lang.NullPointerException: Cannot invoke "dev.openfeature.flagd.grpc.ServiceGrpc$ServiceStub.eventStream(dev.openfeature.flagd.grpc.Schema$EventStreamRequest, io.grpc.stub.StreamObserver)" because "this.serviceStub" is null
	at dev.openfeature.contrib.providers.flagd.FlagdProviderTest.path_arg_should_build_domain_socket_with_correct_path(FlagdProviderTest.java:101)
```
I'll just use `any()` for that call unless something jumps out

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/EventStreamCallback.java
skyerus: Done

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
skyerus: Nice suggestion, done

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/EventStreamObserver.java
skyerus: Done, thanks

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/EventStreamObserver.java
skyerus: Done

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/EventStreamObserver.java
skyerus: Ah thanks, wasn't sure on the best logging practice here :) Done

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/EventStreamObserver.java
skyerus: Done

skyerus:
> @skyerus I see a `!` in the title - is there actually a breaking change here, and if so, could you note it?

The constructors have changed, to anyone using the non-exhaustive constructors (e.g. `new FlagdProvider();`) this won't be a breaking change. To anyone using the exhaustive constructors they will be forced to make changes to include the new parameters. This could be avoided by sustaining the existing constructors but is this worth the maintenance?

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/EventStreamObserver.java
Kavindu-Dodan: I went through the PR again and have a two suggestions, 

1 - Extract caching implementation to its own class

It is better to abstract the cache implementation to its own class (ex:- FlagDCache). This way, we get set of clear contracts and make further changes easy 

For example, cache consumer (ex:- `FlagdProvider`, `EventStreamObserver`) doesn't need to deal with Maps. And will rely on the instance which exposes methods like `FlagDCacheInstance.purge()` or `FlagDCacheInstance.remove(<key>)`

2 - Multiple maps vs Single map

I think it's better to use a single map. For the type, I would suggest using ` dev.openfeature.sdk.Value` as the only Value type and utilize type-specific methods (ex:- `Value.asBoolean()`) it exposes. This way we have a good contract between our cache and the `FeatureProvider` implementation.

Kavindu-Dodan: And regarding type casting, multiple maps do avoid the requirement of type casting. But, that brings other concerns like the cache size enforcement and maintenance of them (ex:- purging, instantiation... )

If we opt to use `Value` from sdk, we can abstract out the casting issue to the Value class (it does use downcasting from Object to type). And later on, we could improve `Value` class internals to avoid casting if we really need to.

skyerus: Thanks for the suggestions @Kavindu-Dodan , I've implemented as such :)

Kavindu-Dodan (approved):
lgtm

Kavindu-Dodan:
> > @skyerus I see a `!` in the title - is there actually a breaking change here, and if so, could you note it?
> 
> The constructors have changed, to anyone using the non-exhaustive constructors (e.g. `new FlagdProvider();`) this won't be a breaking change. To anyone using the exhaustive constructors they will be forced to make changes to include the new parameters. This could be avoided by sustaining the existing constructors but is this worth the maintenance?

From the changeset, I do not see that we have a breaking change 🤔 

We have preserved all public constructors, `public FlagdProvider(String socketPath) , public FlagdProvider(String host, int port, boolean tls, String certPath) & public FlagdProvider() `. But only have provided overloading options to them. The only breaking change is on the package private constructor, which no one should be using. 

@skyerus  could you please double check on this from your end ? 

otherwise, PR is good to merge :)

skyerus:
@Kavindu-Dodan you are right, thanks! I've amended the title to not be a breaking change :)

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/EventStreamObserver.java
```diff
@@ -0,0 +1,74 @@
...
+    public void onNext(EventStreamResponse value) {
+        switch (value.getType()) {
+            case configurationChange:
+                this.handleConfigurationChangeEvent(value);
+                break;
+            case providerReady:
+                this.handleProviderReadyEvent();
+                break;
+            default:
+                log.warn("unhandled event type {}", value.getType());
```
toddbaert:
```suggestion
                log.debug("unhandled event type {}", value.getType());
```

We actually get a lot of other events ("keep-alive") for instance. I would recommend we make this a `debug` to avoid noise.

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/FlagdProviderTest.java
```diff
@@ -381,4 +420,91 @@ private NettyChannelBuilder getMockChannelBuilderSocket() {
         when(mockChannelBuilder.build()).thenReturn(null);
         return mockChannelBuilder;
     }
+
+    private void do_resolvers_cache_responses(String reason, Boolean eventStreamAlive, Boolean shouldCache) {
```
toddbaert, line 542: Good idea extracting this, it makes the tests pretty clear.

toddbaert (approved):
LGTM. Approved, but please consider changing [this log level](https://github.com/open-feature/java-sdk-contrib/pull/168#discussion_r1060804089).

skyerus:
Converting to draft to implement eventing changes: https://github.com/open-feature/flagd/pull/241

toddbaert (changes requested):
I notice now that we don't have a test for the cache invalidation. I think we should add that. Please let me know if I'm wrong.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/EventStreamObserver.java
```diff
@@ -0,0 +1,76 @@
...
+        this.callback.setEventStreamAlive(false);
+    }
+
+    private void handleConfigurationChangeEvent(EventStreamResponse value) {
+        if (!this.cache.getEnabled()) {
+            return;
+        }
+
+        Map<String, Value> data = value.getData().getFieldsMap();
+        Value flagsValue = data.get("flags");
```
Kavindu-Dodan: @skyerus This get operation has the potential to create a NPE. Similar to https://github.com/open-feature/go-sdk-contrib/issues/84 for GO, we will need special handling here. 

Plus, let's extract `flags` to a static constant (similar to configurationChange or providerReady)

skyerus: Nice spot, thanks!
Implemented a null check, please check that I've checked sufficiently against a NPE.

skyerus:
> I notice now that we don't have a test for the cache invalidation. I think we should add that. Please let me know if I'm wrong.

Great shout! I've added a test for cache invalidation

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/FlagdProviderTest.java
```diff
@@ -362,17 +407,129 @@ void reason_mapped_correctly_if_unknown() {
...
+        OpenFeatureAPI.getInstance().setProvider(new FlagdProvider(serviceBlockingStubMock, serviceStubMock, "lru", 100, 5));
 
         FlagEvaluationDetails<Boolean> booleanDetails = api.getClient()
                 .getBooleanDetails(FLAG_KEY, false, new MutableContext());
         assertEquals(Reason.UNKNOWN.toString(), booleanDetails.getReason()); // reason should be converted to UNKNOWN
     }
 
+
+    @Test
+    void invalidate_cache() {
```
skyerus, line 424: Here @toddbaert

Kavindu-Dodan (approved):

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/EventStreamObserver.java
Kavindu-Dodan: @skyerus yes, the new logic looks good and sufficiently covers the usage case.

toddbaert (approved):

PR #169, opened 2022-12-18
feat: add spec compliant otel hook

### LINKED ISSUE #144: Add OTel Hook
Now that [OTel 2529](https://github.com/open-telemetry/opentelemetry-specification/pull/2529) is merged, we should implement the Open Telemetry hook (currently exists as a stub).

Basically, this would entail creating a hook which:

- in the `after` stage: gets the current span and adds an appropriately named event (`feature_flag`), with the attributes (`feature_flag.key`, `feature_flag.provider_name`, and `feature_flag.variant`)
- in the `error` stage, records the exception

In pseudocode, this looks something like:

```
class OTelHook implements Hook {
  after(context, details) {
    var currentTrace = trace.getActiveSpan();
    currentTrace.addEvent("feature_flag", { "feature_flag.key": details.key, /*  other props */ } );
  }

  error(err) {
    var currentTrace = trace.getActiveSpan();
    currentTrace.recordException(err);
  }
}
```
See: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/feature-flags.md

thiyagu06:
## This PR

added otel spec compliant hook.

### Related Issues

Fixes #144

### How to test

add the hook and see the result in Jaeger tracing UI.

`openFeatureAPI.addHooks(new OpenTelemetryHook());`

image

### DIFF
--- a/hooks/README.md
+++ b/hooks/README.md
@@ -7,12 +7,15 @@ evaluation to a span as a span event. This can be used to determine the impact a
 enabling enhanced observability use cases, such as A/B testing or progressive feature releases.
 
 ## Installation
-
+<!-- x-release-please-start-version -->
+```xml
+    <dependency>
+        <groupId>dev.openfeature.contrib.hooks</groupId>
+        <artifactId>otel</artifactId>
+        <version>0.4.0</version>
+    </dependency>
 ```
-    <groupId>dev.openfeature.contrib.hooks</groupId>
-	<artifactId>otel</artifactId>
-	<version>0.4.0</version> 
-```
+<!-- x-release-please-end-version -->
 
 ## Usage
 
--- a/hooks/README.md
+++ b/hooks/README.md
@@ -1,3 +1,25 @@
 # OpenFeature Java Hooks
 
-Hooks are a mechanism whereby application developers can add arbitrary behavior to flag evaluation. They operate similarly to middleware in many web frameworks. Please see the [spec](https://github.com/open-feature/spec/blob/main/specification/sections/04-hooks.md) for more details.
+The OpenTelemetry hook for OpenFeature provides
+a [spec compliant] (https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/feature-flags.md)
+way to automatically add a feature flag
+evaluation to a span as a span event. This can be used to determine the impact a feature has on a request,
+enabling enhanced observability use cases, such as A/B testing or progressive feature releases.
+
+## Installation
+
+```
+    <groupId>dev.openfeature.contrib.hooks</groupId>
+	<artifactId>otel</artifactId>
+	<version>0.4.0</version> 
+```
+
+## Usage
+
+OpenFeature provider various ways to register hooks. The location that a hook is registered affects when the hook is
+run. It's recommended to register the `OpenTelemetryHook` globally in most situations, but it's possible to only enable
+the hook on specific clients. You should **never** register the `OpenTelemetryHook` globally and on a client.
+
+```
+     OpenFeatureAPI.getInstance().addHooks(new OpenTelemetryHook());
+```
--- a/hooks/open-telemetry/src/main/java/dev/openfeature/contrib/hooks/otel/OpenTelemetryHook.java
+++ b/hooks/open-telemetry/src/main/java/dev/openfeature/contrib/hooks/otel/OpenTelemetryHook.java
@@ -1,13 +1,11 @@
 package dev.openfeature.contrib.hooks.otel;
 
-import com.fasterxml.jackson.databind.ObjectMapper;
 import dev.openfeature.sdk.FlagEvaluationDetails;
 import dev.openfeature.sdk.Hook;
 import dev.openfeature.sdk.HookContext;
 import io.opentelemetry.api.common.AttributeKey;
 import io.opentelemetry.api.common.Attributes;
 import io.opentelemetry.api.trace.Span;
-import lombok.SneakyThrows;
 
 import java.util.Map;
 
@@ -25,8 +23,6 @@ public class OpenTelemetryHook implements Hook {
 
     private static final String EVENT_NAME = "feature_flag";
 
-    private final ObjectMapper objectMapper = new ObjectMapper();
-
     /**
      * Create a new OpenTelemetryHook instance.
      */
@@ -40,11 +36,10 @@ public class OpenTelemetryHook implements Hook {
      * @param details Information about how the flag was resolved, including any resolved values.
      * @param hints  An immutable mapping of data for users to communicate to the hooks.
      */
-    @SneakyThrows
     @Override
     public void after(HookContext ctx, FlagEvaluationDetails details, Map hints) {
         Span currentSpan = Span.current();
-        String value = details.getValue() != null ? getValue(details.getValue()) : "";
+        String value = details.getValue() != null ? details.getValue().toString() : "";
         if (currentSpan != null) {
             Attributes attributes = Attributes.of(
                     flagKeyAttributeKey, ctx.getFlagKey(),
@@ -70,10 +65,5 @@ public class OpenTelemetryHook implements Hook {
                     providerNameAttributeKey, ctx.getProviderMetadata().getName());
             currentSpan.recordException(error, attributes);
         }
-    }
-
-    @SneakyThrows
-    private String getValue(Object value) {
-        return objectMapper.writeValueAsString(value);
     }
 }
--- a/hooks/open-telemetry/src/main/java/dev/openfeature/contrib/hooks/otel/OpenTelemetryHook.java
+++ b/hooks/open-telemetry/src/main/java/dev/openfeature/contrib/hooks/otel/OpenTelemetryHook.java
@@ -1,29 +1,79 @@
 package dev.openfeature.contrib.hooks.otel;
 
-import dev.openfeature.sdk.Client;
-import dev.openfeature.sdk.NoOpProvider;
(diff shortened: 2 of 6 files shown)


hooks/README.md
```diff
@@ -1,3 +1,25 @@
...
+way to automatically add a feature flag
+evaluation to a span as a span event. This can be used to determine the impact a feature has on a request,
+enabling enhanced observability use cases, such as A/B testing or progressive feature releases.
+
+## Installation
+
+```
+    <groupId>dev.openfeature.contrib.hooks</groupId>
+	<artifactId>otel</artifactId>
+	<version>0.4.0</version>
```
beeme1mr: This version should be controlled with release please. @toddbaert should be able to provide some examples.

toddbaert: Yep! Just add annotations as seen [here](https://github.com/open-feature/java-sdk-contrib/blob/8fe9ebd3f0491098e22ac195d7d9e2347fec1cc6/providers/flagd/README.md?plain=1#L8-L16) - it will make sure the version is updated in your example as we release.

@thiyagu06

beeme1mr:
Thanks @thiyagu06! We'll review this tomorrow but it looks great at first glance.

I noticed in the screenshot you attached that the variant is quoted. Is that expected?

thiyagu06:
> Thanks @thiyagu06! We'll review this tomorrow but it looks great at first glance.
> 
> I noticed in the screenshot you attached that the variant is quoted. Is that expected?

yeah. Thanks for pointing that. the variant is double quoted because, I convert the value into json format. https://github.com/open-feature/java-sdk-contrib/blob/4a544efcb2deaa57b998a02a50114ff57ebadc20/hooks/open-telemetry/src/main/java/dev/openfeature/contrib/hooks/otel/OpenTelemetryHook.java#L77 The reason behind that is I'm not sure what kind of values use will be adding for the `Object ` type. Is that okay if we could use toString method instead of using jackson serialise it.

hooks/README.md
beeme1mr: The Release Please config will also need to be updated to include the readme.

https://github.com/open-feature/java-sdk-contrib/blob/main/release-please-config.json#L31

toddbaert:
> > Thanks @thiyagu06! We'll review this tomorrow but it looks great at first glance.
> > I noticed in the screenshot you attached that the variant is quoted. Is that expected?
> 
> yeah. Thanks for pointing that. the variant is double quoted because, I convert the value into json format.
> 
> https://github.com/open-feature/java-sdk-contrib/blob/4a544efcb2deaa57b998a02a50114ff57ebadc20/hooks/open-telemetry/src/main/java/dev/openfeature/contrib/hooks/otel/OpenTelemetryHook.java#L77
> 
> The reason behind that is I'm not sure what kind of values use will be adding for the `Object ` type. Is that okay if we could use toString method instead of using jackson serialise it.

Yes, in fact, I think we should try to avoid dependencies such as jackson here. I think `.toString()` is fine.

Complex object values are represented by our `dev.openfeature.sdk.Structure` interfaces, which house attribute maps and can have their own `toString` implementation (though for the current implementations, the default works fine). I don't think we need to standardize on JSON here.

If in the future we decide to standardize the representation of these objects, we could implement one in the overridden `toString`.

hooks/open-telemetry/src/main/java/dev/openfeature/contrib/hooks/otel/OpenTelemetryHook.java
```diff
@@ -1,29 +1,69 @@
...
+     * @param hints  An immutable mapping of data for users to communicate to the hooks.
      */
-    public static boolean test() {
-        OpenFeatureAPI.getInstance().setProvider(new NoOpProvider());
-        Client client = OpenFeatureAPI.getInstance().getClient();
-        return client.getBooleanValue("test2", true);
+    @Override
+    public void after(HookContext ctx, FlagEvaluationDetails details, Map hints) {
+        Span currentSpan = Span.current();
+        String value = details.getValue() != null ? details.getValue().toString() : "";
```
toddbaert: As per the [semantic convention](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/feature-flags.md), we need to first use the `FlagEvaluationDetails.getVariant` to populate the `"feature_flag.variant"` and only fall back to the value if the variant is not set.

I'm also not sure about an empty string if the value is `null`. Is that better than `null`?

thiyagu06:
> > > Thanks @thiyagu06! We'll review this tomorrow but it looks great at first glance.
> > > I noticed in the screenshot you attached that the variant is quoted. Is that expected?
> > 
> > 
> > yeah. Thanks for pointing that. the variant is double quoted because, I convert the value into json format.
> > https://github.com/open-feature/java-sdk-contrib/blob/4a544efcb2deaa57b998a02a50114ff57ebadc20/hooks/open-telemetry/src/main/java/dev/openfeature/contrib/hooks/otel/OpenTelemetryHook.java#L77
> > 
> > The reason behind that is I'm not sure what kind of values use will be adding for the `Object ` type. Is that okay if we could use toString method instead of using jackson serialise it.
> 
> Yes, in fact, I think we should try to avoid dependencies such as jackson here. I think `.toString()` is fine.
> 
> Complex object values are represented by our `dev.openfeature.sdk.Structure` interfaces, which house attribute maps and can have their own `toString` implementation (though for the current implementations, the default works fine). I don't think we need to standardize on JSON here.
> 
> If in the future we decide to standardize the representation of these objects, we could implement one in the overridden `toString`.

Yeah. I understand that. Fixed it

hooks/open-telemetry/src/main/java/dev/openfeature/contrib/hooks/otel/OpenTelemetryHook.java
```diff
@@ -1,29 +1,69 @@
...
-    /** 
+public class OpenTelemetryHook implements Hook {
+
+    private final AttributeKey<String> flagKeyAttributeKey = AttributeKey.stringKey("feature_flag.flag_key");
+
+    private final AttributeKey<String> providerNameAttributeKey = AttributeKey.stringKey("feature_flag.provider_name");
+
+    private final AttributeKey<String> variantAttributeKey = AttributeKey.stringKey("feature_flag.variant");
+
+    private static final String EVENT_NAME = "feature_flag";
```
toddbaert: You can probably build the attribute names by concatenating them with `EVENT_NAME` instead of repeating `"feature_flag"`.

hooks/open-telemetry/src/test/java/dev/openfeature/contrib/hooks/otel/OpenTelemetryHookTest.java
```diff
@@ -1,15 +1,116 @@
 package dev.openfeature.contrib.hooks.otel;
 
-import static org.assertj.core.api.Assertions.assertThat;
-
+import dev.openfeature.sdk.FlagEvaluationDetails;
```
toddbaert, line 3: tests look good! thank you.

toddbaert (changes requested):
My only real concern is: https://github.com/open-feature/java-sdk-contrib/pull/169/files#r1055525423

Once this is addressed it looks like it can be merged. Thanks a lot, especially for even giving this a test in Jaeger.

hooks/open-telemetry/src/main/java/dev/openfeature/contrib/hooks/otel/OpenTelemetryHook.java
thiyagu06: Yes.. You're right. I interpret it in the wrong way. now, fixed it. 

> I'm also not sure about an empty string if the value is null. Is that better than null?

I just gave a second thought about this. If we convert the null value to empty string, it might convey different meaning to the user as well. so decided to keep is as null.

hooks/open-telemetry/src/main/java/dev/openfeature/contrib/hooks/otel/OpenTelemetryHook.java
```diff
@@ -1,29 +1,69 @@
...
      */
-    public static boolean test() {
-        OpenFeatureAPI.getInstance().setProvider(new NoOpProvider());
-        Client client = OpenFeatureAPI.getInstance().getClient();
-        return client.getBooleanValue("test2", true);
+    @Override
+    public void after(HookContext ctx, FlagEvaluationDetails details, Map hints) {
+        Span currentSpan = Span.current();
+        if (currentSpan != null) {
+            String variant = details.getVariant() != null ? details.getVariant() : String.valueOf(details.getValue());
```
toddbaert, line 43: I think `String.valueOf(details.getValue()` will throw if `details.getValue()` is `null` (which is should not be, but just in case we might want to handle that).

thiyagu06: no, it's not. It just the return the string `null` https://docs.oracle.com/javase/7/docs/api/java/lang/String.html#valueOf(java.lang.Object)

toddbaert: You're right!

hooks/open-telemetry/src/main/java/dev/openfeature/contrib/hooks/otel/OpenTelemetryHook.java
toddbaert: Makes sense.

toddbaert (approved):
Looks good to me, thank you!

PR #201, opened 2023-01-17
fix: allow flagd-provider cache to be disabled

skyerus:
## This PR

Fixes issue when using flagd-provider with cache disabled.


toddbaert:
I think it would be a good idea to add one test case for this. That's my general rule for discovering any sort of bug - it means a test was probably missing!

skyerus:
It was discovered while working on the [integration tests](https://github.com/open-feature/java-sdk/pull/249), it's covered there now but I think you're right that it should be covered by a unit test too

Kavindu-Dodan (approved):

toddbaert (approved):

PR #204, opened 2023-01-18
feat: exposed eventStreamAliveSync to allow application authors to block until event stream is alive

skyerus:
## This PR

Exposes a sync mechanism to allow users of flagd provider to block until the event stream is alive.

### DIFF
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
@@ -638,6 +638,8 @@ public class FlagdProvider implements FeatureProvider, EventStreamCallback {
 
     /**
      * Call .wait() on this to block until the event stream is alive.
+     * Can be used in instances where the provider being connected to the event stream is a prerequisite
+     * to execution (e.g. testing). Not necessary for standard usage.
      *
      * @return eventStreamAliveSync
      */
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
@@ -87,6 +87,7 @@ public class FlagdProvider implements FeatureProvider, EventStreamCallback {
     private int maxEventStreamRetries = DEFAULT_MAX_EVENT_STREAM_RETRIES;
 
     private ReadWriteLock lock = new ReentrantReadWriteLock();
+    private Object eventStreamAliveSync;
 
     /**
      * Create a new FlagdProvider instance.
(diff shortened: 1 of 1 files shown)


providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
```diff
@@ -630,4 +635,13 @@ public void restartEventStream() throws Exception {
...
     }
+
+    /**
+     * Call .wait() on this to block until the event stream is alive.
+     *
+     * @return eventStreamAliveSync
+     */
+    public Object getEventStreamAliveSync() {
+        return this.eventStreamAliveSync;
+    }
```
toddbaert: If this isn't waited, and evaluations start happening, is it possible we missed events about flag changes? Should we not cache until the stream is definitely connected?

skyerus: There is already a check that the event stream is alive before retrieving from cache: https://github.com/open-feature/java-sdk-contrib/blob/main/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java#L192

toddbaert: OK so correct me if I'm wrong: 

The race was just really in tests - because the tests could be flaky if evaluations had started but the connection hadn't yet been made - it wouldn't be an issue in production, because the end result would be the cache just takes a few moments to start up... but that breaks the test assumptions since the 2nd eval might not be cached.

Is that right?

skyerus: That's exactly right. A similar problem occurred in the Go implementation, I went back and exposed a similar block until ready

toddbaert: Great, makes sense to me. Thanks!

toddbaert (approved):

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
```diff
@@ -630,4 +635,13 @@ public void restartEventStream() throws Exception {
 
         this.handleEvents();
     }
+
+    /**
+     * Call .wait() on this to block until the event stream is alive.
```
toddbaert, line 640: Can you add a tiny bit of explanation here on why this might be needed? Maybe mention it's probably not need in prod. I just dont want people to think it's necessary when it's not.

skyerus: Done

Kavindu-Dodan (approved):
I am interested to see the usage of this in tests :)

skyerus:
> I am interested to see the usage of this in tests :)

You can see it being used in tests here: https://github.com/open-feature/java-sdk/pull/249 :)

Issue #211, opened 2023-02-02
Flagsmith classes not public


glaucioscheibel:
Flagsmith provider class and option class was created using package restricted access, so examples doesn't work.

toddbaert:
@glaucioscheibel Thanks for the report!

I've opened a [fix PR](https://github.com/open-feature/java-sdk-contrib/pull/213)

toddbaert:
Fixed with https://github.com/open-feature/java-sdk-contrib/pull/213

toddbaert:
This fix will be released when https://github.com/open-feature/java-sdk-contrib/pull/209 is merged (plus an hour or so wait for Maven to index it).

PR #213, opened 2023-02-06
fix: make flagsmith classes public

toddbaert:
Making access modifiers for supposedly public classes public.


toddbaert:
Closes: https://github.com/open-feature/java-sdk-contrib/issues/211

toddbaert:
@dabeeeenster @ajhelsby I also noticed that the FlagsmithClient was a a static member. That seems like it could be a potential problem to me if users instantiate multiple flagsmith providers (probably an edge case). Our bug check pointed it out when I made the class public.

I've made the internal flagsmith member and instance member now. Is that OK?

glaucioscheibel (approved):
LGTM

beeme1mr (approved):

matthewelwell (approved):

PR #234, opened 2023-02-27
chore: make resolver generic

toddbaert:
This PR contains no functional changes. It just refactors the flagd provider to use a generic `resolve` method which removes a lot of duplication and cruft and reduces code by about ~100 lines.

All of this is already tested.

### DIFF
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdCache.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdCache.java
@@ -1,11 +1,11 @@
 package dev.openfeature.contrib.providers.flagd;
 
-import dev.openfeature.sdk.ProviderEvaluation;
-import dev.openfeature.sdk.Value;
-
-import java.util.Map;
-import org.apache.commons.collections4.map.LRUMap;
 import java.util.Collections;
+import java.util.Map;
+
+import org.apache.commons.collections4.map.LRUMap;
+
+import dev.openfeature.sdk.ProviderEvaluation;
 
 /**
  * Exposes caching mechanism for flag evaluations.
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdCache.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdCache.java
@@ -27,7 +27,8 @@ public class FlagdCache {
                 return;
             case LRU_CACHE:
             default:
-                this.store = Collections.synchronizedMap(new LRUMap<String, ProviderEvaluation<? extends Object>>(maxCacheSize));
+                this.store = Collections
+                        .synchronizedMap(new LRUMap<String, ProviderEvaluation<? extends Object>>(maxCacheSize));
         }
 
         this.enabled = true;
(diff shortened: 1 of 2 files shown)


providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
```diff
@@ -210,197 +229,58 @@ public String getName() {
...
-                .value(new Value(r.getValue()))
-                .variant(r.getVariant())
-                .reason(r.getReason())
-                .build();
-            this.cache.put(key, value);
-        }
-        
-        return result;
+        return this.resolve(key, ctx, request,
+                this.serviceBlockingStub.withDeadlineAfter(this.deadline, TimeUnit.MILLISECONDS)::resolveBoolean, null);
```
toddbaert, line 235: This is the TLDR of this change... now we call `this.resolve` with a `request` and a resolver function ref (`::resolveBoolean`) instead of duplicating this stuff. We also take an optional lambda describing how to transform the gRPC value field.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdCache.java
```diff
@@ -11,7 +11,7 @@
  * Exposes caching mechanism for flag evaluations.
  */
 public class FlagdCache {
-    private Map<String,ProviderEvaluation<Value>> store;
+    private Map<String,ProviderEvaluation<? extends Object>> store;
```
thisthat, line 14: I would expect that all elements stored can be casted to `Value`
```suggestion
    private Map<String,ProviderEvaluation<? super Value>> store;
```

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
skyerus: Nice! Making a note to do something similar in the Go provider

skyerus (approved):

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdCache.java
toddbaert: `Value` is actually a wrapper type that represents a JSON value (including JSON structures). We could convert all our resolved values to this, but it's not necessary and would require creating `Value` instances, not just a cast. I've chosen `? extends Object` because it's the most "honest" type for this - gRPC can return Bools, Floats, Longs, Strings, Structs here, so Object makes sense to me - but maybe I'm missing something.

Previously we were doing the sort of conversion I mention, but it's not strictly necessary.

thisthat: Looking into the code, the type held by `ProviderEvaluation` is never checked/used. Your suggestion is valid, sorry for the noise 😇

thisthat (approved):

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdCache.java
toddbaert: Glad to have a check from you!

beeme1mr (approved):

Kavindu-Dodan (approved):

PR #241, opened 2023-03-15
feat: A new inline evaluating provider based on jsonlogic.com

justinabrahms:
## This PR
This adds a new provider based on jsonlogic data format.

### Related Issues

Refs https://github.com/open-feature/ofep/pull/57

### Follow-up Tasks

Haven't yet done the "convert into a value" yet, but this is a reasonable start.

### How to test

The tests run. Feel free to provide additional test case suggestions.


beeme1mr:
Hey @justinabrahms, thanks for putting this together. It's great to see new providers like this added to the contribs repo. I'll let @toddbaert and/or @Kavindu-Dodan handle the full review (I'm not a Java person) but the basics look good.

By the way, flagd extends JSON logic to support deterministic splits. If that's ever a requirement here, please consider using the same flag configuration and hash algorithm.

https://github.com/open-feature/flagd/blob/main/docs/configuration/fractional_evaluation.md

providers/inline-evaluating-provider/main/java/dev/openfeature/contrib/providers/inlineeval/RuleFetcher.java
```diff
@@ -0,0 +1,20 @@
...
+     */
+    void initialize(EvaluationContext initialContext);
+
+    /**
+     * Given a key name, return the JSONLogic rules for it
+     * @param key The key to fetch logic for
+     * @return json logic rules or null
+     */
+    String getRuleForKey(String key);
+}
```
toddbaert: Abstracting this into an interface seems like a good idea. As a thought experiment, I imagined adding a simple additional RPC to flagd to allow it's ruleset to be downloaded. In that hypothetical flow, the ruleset would be cached with `initialize` and then `getRuleForKey` could be used for evaluation. This doesn't QUITE work for flagd, since in the flagd schema the result of the rule is actually a `variant` which is the [key representing the value](https://github.com/open-feature/flagd/blob/main/config/samples/example_flags.flagd.json#L47-L70). Some additional mapping would be required to pull the associated value out of flagd's variants.

... but my point is I think the abstraction covers a lot of use-cases.

Do you think it's worth adding a very basic implementation at some point? Perhaps from a file?

providers/inline-evaluating-provider/test/java/dev/openfeature/contrib/providers/inlineeval/ExampleFetcher.java
```diff
@@ -0,0 +1,23 @@
...
+        return """
+                {"if": [
+                  {"or": [
+                    {"in": [{"var": "userId"}, [1,2,3,4]]},
+                    {"in": [{"var": "category"}, ["pies", "cakes"]]},
+                  ]},
+                  true,
+                  false
+                ]}
+                """;
```
toddbaert: I'm surprised `TextBlocks` work in the CI TBH, since we use JDK8 there and they seem to be added with Java 15.

toddbaert:
Could you add a `src/` dir under the `inline-evaluating-provider` dir and put the source and test files there, like in the other libs? I think as it stands this might cause some issues because it's [not a standard maven project dir](https://maven.apache.org/guides/introduction/introduction-to-the-standard-directory-layout.html).

EDIT: Actually, after testing a bit I don't think your tests are running in the CI for this reason, and the package will probably contain no bytecode. I added a test that should have failed and it did not.

toddbaert (changes requested):
Looks good in theory, but I think you'll need to resolve [this](https://github.com/open-feature/java-sdk-contrib/pull/241#issuecomment-1472659288) or no actual java bytecode will be published.

toddbaert:
You may want to add an entry [here](https://github.com/open-feature/java-sdk-contrib/blob/main/.release-please-manifest.json), or this will default to 1.0 on release (up to you though).

providers/inline-evaluating-provider/main/java/dev/openfeature/contrib/providers/inlineeval/RuleFetcher.java
justinabrahms: We could do that, sure. It would read json from a file and do a simple get by key.

providers/inline-evaluating-provider/pom.xml
```diff
@@ -0,0 +1,47 @@
...
+        <artifactId>parent</artifactId>
+        <version>0.1.0</version>
+        <relativePath>../../pom.xml</relativePath>
+    </parent>
+    <!-- The group id MUST start with dev.openfeature, or publishing will fail. OpenFeature has verified ownership of this (reversed) domain. -->
+    <groupId>dev.openfeature.contrib.providers</groupId>
+    <artifactId>inline-evaluating-provider</artifactId>
+    <version>0.0.1</version> <!--x-release-please-version -->
+
+    <name>inline-evaluating-provider</name>
```
thiyagu06: can we rename it as `jsonlogic-eval-provider`?  when we say inline it's more generic and I feel even `env-var`, `in-memory-eval` all  falls into this category?

providers/inline-evaluating-provider/src/main/java/dev/openfeature/contrib/providers/inlineeval/FileBasedFetcher.java
```diff
@@ -0,0 +1,26 @@
...
+    JSONObject rules;
+
+    public FileBasedFetcher(URI filename) throws IOException {
+        String jsonData = String.join("", Files.readAllLines(Paths.get(filename)));
+        rules = new JSONObject(jsonData);
+    }
+
+    @Override
+    public String getRuleForKey(String key) {
+        return rules.getJSONObject(key).toString();
```
thiyagu06: We need a null check here. or is it good idea to change the return type to `Optional<String>` in `RuleFetcher.java`

providers/inline-evaluating-provider/src/main/java/dev/openfeature/contrib/providers/inlineeval/InlineEvaluatingProvider.java
```diff
@@ -0,0 +1,75 @@
...
+                    .reason("Unable to find rules for the given key")
+                    .build();
+        }
+
+        try {
+            return ProviderEvaluation.<T>builder()
+                    .value(resultToType.apply(this.logic.apply(rule, ctx.asObjectMap())))
+                    .build();
+        } catch (JsonLogicException e) {
+            throw new RuntimeException(e);
```
thiyagu06: can this be a https://github.com/open-feature/java-sdk/blob/main/src/main/java/dev/openfeature/sdk/exceptions/ParseError.java?

providers/inline-evaluating-provider/src/test/java/dev/openfeature/contrib/providers/inlineeval/InlineEvaluatingProviderTest.java
```diff
@@ -0,0 +1,71 @@
...
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.util.Collections;
+import java.util.HashMap;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.mockito.Mockito.*;
+
+class InlineEvaluatingProviderTest {
```
thiyagu06: add test case to verify the behaviour when rule is missing for given key?

providers/inline-evaluating-provider/src/main/java/dev/openfeature/contrib/providers/inlineeval/InlineEvaluatingProvider.java
```diff
@@ -0,0 +1,75 @@
...
+import io.github.jamsesso.jsonlogic.JsonLogic;
+import io.github.jamsesso.jsonlogic.JsonLogicException;
+
+import java.util.function.Function;
+
+public class InlineEvaluatingProvider implements FeatureProvider {
+    private final JsonLogic logic;
+    private final RuleFetcher fetcher;
+
+    void initialize(EvaluationContext initialContext) {
```
thiyagu06: Why do we need to this non public method here? can we delegate the `initialize`  responsibility to an application author when they create an instance of RuleFetcher?

justinabrahms: Should be public. This is in an ofep currently, but this provider will need to do some bootstrapping as part of it's lifecycle.

providers/jsonlogic-eval-provider/src/main/java/dev/openfeature/contrib/providers/jsonlogic/JsonlogicProvider.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    public JsonlogicProvider(JsonLogic logic, RuleFetcher fetcher) {
+        this.logic = logic;
+        this.fetcher = fetcher;
+    }
+
+    @Override
+    public Metadata getMetadata() {
+        return null;
```
thiyagu06: return metadata object instead of null.

toddbaert (approved):
LGTM with the exception of the metadata as @thiyagu06 pointed out: https://github.com/open-feature/java-sdk-contrib/pull/241#discussion_r1161057903

providers/jsonlogic-eval-provider/README.md
```diff
@@ -0,0 +1,41 @@
+# JSONLogic Evaluation Provider
+
+This provider does inline evaluation (e.g. no hot-path remote calls) based on JSONLogic. This should allow you to
```
Kavindu-Dodan:
```suggestion
This provider does inline evaluation (e.g. no hot-path remote calls) based on [JSONLogic](https://jsonlogic.com/). This should allow you to 
```

providers/jsonlogic-eval-provider/src/main/java/dev/openfeature/contrib/providers/jsonlogic/FileBasedFetcher.java
```diff
@@ -0,0 +1,49 @@
...
+ * A {@link RuleFetcher} which reads in the rules from a file. It assumes that the keys are the flag keys and the
+ * values are the json logic rules.
+ */
+@SuppressFBWarnings(
+        value = "PATH_TRAVERSAL_IN",
+        justification = "This is expected to read files based on user input"
+)
+public class FileBasedFetcher implements RuleFetcher {
+    private final Logger log;
+    JSONObject rules;
```
Kavindu-Dodan: can be `private final`

providers/jsonlogic-eval-provider/src/main/java/dev/openfeature/contrib/providers/jsonlogic/JsonlogicProvider.java
```diff
@@ -0,0 +1,86 @@
...
+        return evalRuleForKey(key, defaultValue, ctx, (o) -> (T) o);
+    }
+
+    private <T> ProviderEvaluation<T> evalRuleForKey(
+            String key, T defaultValue, EvaluationContext ctx, Function<Object, T> resultToType) {
+        String rule = fetcher.getRuleForKey(key);
+        if (rule == null) {
+            return ProviderEvaluation.<T>builder()
+                    .value(defaultValue)
+                    .reason("Unable to find rules for the given key")
```
Kavindu-Dodan: Minor - reason is best mapped to a known value referring to SDK (ex:- ERROR) [1] and then the message could go into `errorMessage(String)`

[1] - https://github.com/open-feature/java-sdk/blob/main/src/main/java/dev/openfeature/sdk/Reason.java#L7

providers/jsonlogic-eval-provider/src/main/java/dev/openfeature/contrib/providers/jsonlogic/FileBasedFetcher.java
```diff
@@ -0,0 +1,49 @@
...
+
+    /**
+     * Create a file based fetcher give a file URI.
+     * @param filename URI to a given file.
+     * @throws IOException when we can't load the file correctly
+     */
+    public FileBasedFetcher(URI filename) throws IOException {
+        this.log = Logger.getLogger(String.valueOf(FileBasedFetcher.class));
```
Kavindu-Dodan: The logger can be static, initialized for the type (`FileBasedFetcher`)

Kavindu-Dodan (approved):
Approving with some comments :)

PR #244, opened 2023-03-18
feat: env-var provider (#243)

### LINKED ISSUE #243: Add a simple environment variables provider implementation
Moved from: https://github.com/open-feature/java-sdk/issues/339

JS SDK already has a simple env var provider implementation [here](https://github.com/open-feature/js-sdk-contrib/tree/main/libs/providers/env-var). It may be used as an example.

madhead:
## This PR

This PR adds a basic implementation of the Environment Variables feature provider, discussed in https://github.com/open-feature/java-sdk/issues/339.

### Related Issues

- https://github.com/open-feature/java-sdk/issues/339
- https://github.com/open-feature/java-sdk-contrib/issues/243

Fixes #243

### Notes

- It's [hard to mock](https://stackoverflow.com/q/64182348/750510) the `java.lang.System`, so I've abstracted it away with an `OS` class. It should be invisible to the end-users, though.
- This implementation doesn't support structured values. Should I add it?

### Follow-up Tasks

- Maybe extend the `EnvVarProvider` to support `Value`s?
- Documentation update?

### How to test

A unit test is included in the PR.


beeme1mr (approved):
Looks like a good start. Thanks for putting this together.

toddbaert:
I'll take a look at this tomorrow!

providers/env-var/src/main/java/dev/openfeature/contrib/providers/envvar/EnvVarProvider.java
```diff
@@ -0,0 +1,81 @@
...
+    public ProviderEvaluation<Value> getObjectEvaluation(String key, Value defaultValue, EvaluationContext ctx) {
+        throw new ValueNotConvertableError("EnvVarProvider supports only primitives");
+    }
+
+    private <T> ProviderEvaluation<T> evaluateEnvironmentVariable(
+            String key,
+            T defaultValue,
+            Function<String, T> parse
+    ) {
+        final String value = os.getenv(key);
```
toddbaert, line 60: One thing we've done in other ENV VAR providers is to map `SCREAMING_SNAKE_CASE` env vars to `hyphen-case-keys`. This _may_ be desirable, since generally people like to use `SCREAMIN_SNAKE` for env vars, but that's generally not true for feature flag keys.

You could also create some transformer interface to allow users to control this translation.... and it doesn't have to be done in this PR.

providers/env-var/src/main/java/dev/openfeature/contrib/providers/envvar/EnvVarProvider.java
```diff
@@ -0,0 +1,81 @@
...
+
+        if (value == null) {
+            return ProviderEvaluation.<T>builder()
+                    .value(defaultValue)
+                    .reason(Reason.DEFAULT.toString())
+                    .build();
+        }
```
toddbaert: I think it's a bit more idiomatic to throw a FlagNotFound here (though probably a tiny bit less performant). The SDK will handle defaulting.

providers/env-var/src/test/java/dev/openfeature/contrib/providers/envvar/EnvVarProviderTest.java
```diff
@@ -0,0 +1,193 @@
...
+    @Test
+    void shouldThrowOnGetObjectEvaluation() {
+        assertThrows(
+                ValueNotConvertableError.class,
+                () -> new EnvVarProvider().getObjectEvaluation("any-key", new Value(), new ImmutableContext())
+        );
+    }
+
+    @TestFactory
+    Iterable<DynamicTest> shouldEvaluateValuesCorrectly() {
```
toddbaert, line 28: I've never seen this table-testing style in Java, I like it!

toddbaert (approved):
Approved with some minor suggestions. If I don't hear back in the next few days, I'll merge as is!

Thanks a lot for this addition!

providers/env-var/src/main/java/dev/openfeature/contrib/providers/envvar/EnvVarProvider.java
madhead: Done!

providers/env-var/src/main/java/dev/openfeature/contrib/providers/envvar/EnvVarProvider.java
madhead: Yes, let me make this in a separate PR later.

providers/env-var/src/main/java/dev/openfeature/contrib/providers/envvar/OS.java
```diff
@@ -0,0 +1,11 @@
+package dev.openfeature.contrib.providers.envvar;
+
+/**
+ * This internal class abstracts away Java's {@link System} class for test purposes.
+ * It is not intended to be used directly.
+ */
+class OS {
```
Kavindu-Dodan, line 7: Minor comment (potential improvement)

We could make this an interface (contract) and have a default implementation based on `System.getenv(name)`

That would allow having custom env variable handling (ex:- prefixes, casing and even sources like spring configurations which come from different sources [1] ) and still use the same provider 

Could be a follow up work :) 

[1] - https://docs.spring.io/spring-boot/docs/1.0.1.RELEASE/reference/html/boot-features-external-config.html

Kavindu-Dodan (approved):
Approving regardless of the comment [1] as this is pure personal opinion :) 

[1] - https://github.com/open-feature/java-sdk-contrib/pull/244/files#r1147943367

providers/env-var/README.md
```diff
@@ -0,0 +1,27 @@
...
+    <artifactId>env-var</artifactId>
+    <version>0.0.1</version>
+</dependency>
+```
+
+<!-- x-release-please-end-version -->
+
+## Usage
+
+`EnvVarProvider` doesn't have eny options.
```
thiyagu06:
```suggestion
`EnvVarProvider` doesn't have any options.
```

PR #259, opened 2023-03-27
fix: [flagd]NPE of flagd due to null context value

### LINKED ISSUE #258: [flagd] NPE with null valued contexts
flagd's `convertPrimitive(Value)` results in an NPE as we explicitly set `null` [1] for context null values. 

This must be corrected by using `NullValue` [2], which allows the representation to accept null values. 

[1] - https://github.com/open-feature/java-sdk-contrib/blob/dev.openfeature.contrib.providers.flagd-v0.5.7/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java#L389
[2] - https://protobuf.dev/reference/java/api-docs/com/google/protobuf/NullValue.html

Kavindu-Dodan:
## This PR

Fixes #258 for flagd provider

Underlying protobuf now use `NullValue` to convey valid `null` value of the context.


thisthat (approved):
LGTM

beeme1mr (approved):

PR #284, opened 2023-04-20
test: introduce test for handling zero values responses

### LINKED ISSUE #135: [flagd-provider] add unit test for zero-value resolutions from flagd
Flagd sends empty values in flag [response payloads](https://github.com/open-feature/schemas/blob/main/protobuf/schema/v1/schema.proto#L54) when the value is equivalent to the zero-value of the associated go type. We should test that this works as expected and still returns an equivalent zero-value (ie: `false`, `0`).

DoD:
- manual test confirming zero values are returned from the flagd provider when they are missing from the gRPC message (boolean set to false or numeric set to 0 in flagd config)
- new tests that assert correct values for zero-value flags

see: https://github.com/open-feature/flagd/issues/218

odubajDT:
## This PR

- adds tests for zero value responses

### Related Issues

Fixes #135

### DIFF
--- a/providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/FlagdProviderTest.java
+++ b/providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/FlagdProviderTest.java
@@ -299,6 +299,77 @@ class FlagdProviderTest {
     }
 
     @Test
+    void zero_value() {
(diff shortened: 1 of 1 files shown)


Kavindu-Dodan (approved):

toddbaert (approved):

toddbaert:
perfect!

PR #286, opened 2023-04-21
feat: flagd manual otel interceptor for grpc

### LINKED ISSUE #283: OTel Span in flagd Provider
The flagd provider should start an OTel span that will be passed to flagd to get cross-service tracing for evaluation.

Kavindu-Dodan:
## This PR

Fixes https://github.com/open-feature/java-sdk-contrib/issues/283 by introducing manual tracing [1] for flagd Java sdk 

This PR introduce a new constructor to provide an OpenTelemetry SDK. Internally, this is used to intercept grpc communication and add simple span for evaluation.

```
 OpenFeatureAPI api = OpenFeatureAPI.getInstance();
api.setProvider(new FlagdProvider(optenTelemetry));
```   

With a proper setup, we get distributed telemetry for flagd ecosystem

image

image

Note - a follow-up refactoring is ready to improve constructor behavior https://github.com/open-feature/java-sdk-contrib/pull/294 

[1] - https://opentelemetry.io/docs/instrumentation/java/manual/

### DIFF
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
@@ -1,21 +1,7 @@
 package dev.openfeature.contrib.providers.flagd;
 
-import java.io.File;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.concurrent.TimeUnit;
-import java.util.concurrent.locks.Lock;
-import java.util.concurrent.locks.ReadWriteLock;
-import java.util.concurrent.locks.ReentrantReadWriteLock;
-import java.util.function.Function;
-import java.util.stream.Collectors;
-
-import javax.net.ssl.SSLException;
-
 import com.google.protobuf.Descriptors.FieldDescriptor;
 import com.google.protobuf.Message;
-
 import com.google.protobuf.NullValue;
 import dev.openfeature.flagd.grpc.Schema.EventStreamRequest;
 import dev.openfeature.flagd.grpc.Schema.EventStreamResponse;
@@ -40,7 +26,24 @@ import io.netty.channel.epoll.EpollDomainSocketChannel;
 import io.netty.channel.epoll.EpollEventLoopGroup;
 import io.netty.channel.unix.DomainSocketAddress;
 import io.netty.handler.ssl.SslContextBuilder;
+import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.api.trace.SpanKind;
+import io.opentelemetry.api.trace.Tracer;
+import io.opentelemetry.context.Scope;
+import io.opentelemetry.sdk.OpenTelemetrySdk;
 import lombok.extern.slf4j.Slf4j;
+
+import javax.net.ssl.SSLException;
+import java.io.File;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+import java.util.function.Function;
+import java.util.stream.Collectors;
 
 /**
  * OpenFeature provider for flagd.
@@ -81,15 +84,18 @@ public class FlagdProvider implements FeatureProvider, EventStreamCallback {
     static final int BASE_EVENT_STREAM_RETRY_BACKOFF_MS = 1000;
 
     private long deadline = DEFAULT_DEADLINE;
+
     private ServiceBlockingStub serviceBlockingStub;
     private ServiceStub serviceStub;
+
+    private Tracer tracer;
 
     private Boolean eventStreamAlive;
     private FlagdCache cache;
 
     private int eventStreamAttempt = 1;
     private int eventStreamRetryBackoff = BASE_EVENT_STREAM_RETRY_BACKOFF_MS;
-    private int maxEventStreamRetries = DEFAULT_MAX_EVENT_STREAM_RETRIES;
+    private int maxEventStreamRetries;
 
     private ReadWriteLock lock = new ReentrantReadWriteLock();
     private Object eventStreamAliveSync;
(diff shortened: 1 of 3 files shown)


providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
```diff
@@ -180,6 +186,17 @@ public FlagdProvider() {
                 fallBackToEnvOrDefault(MAX_EVENT_STREAM_RETRIES_ENV_VAR_NAME, DEFAULT_MAX_EVENT_STREAM_RETRIES));
     }
 
+    public FlagdProvider(OpenTelemetrySdk telemetrySdk) {
```
Kavindu-Dodan: follow up - need a refactoring to provide a option builder for construction

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
```diff
@@ -454,6 +471,12 @@ private static ServiceBlockingStub buildServiceBlockingStub(String host, Integer
         return ServiceGrpc.newBlockingStub(channelBuilder(host, port, tls, certPath, socketPath).build());
     }
 
+    private static ServiceBlockingStub buildServiceBlockingStub(OpenTelemetrySdk sdk) {
+        return ServiceGrpc
+                .newBlockingStub(channelBuilder(null, null, null, null, null)
```
Kavindu-Dodan, line 480: follow up - need a refactoring to provide a option builder for construction

thiyagu06:
Just out of curiosity, can't this instrumentation be implemented using [hooks](https://openfeature.dev/specification/sections/hooks)?

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
thisthat: we could introduce AutoValue here for having also immutability 

https://github.com/google/auto/blob/main/value/userguide/index.md

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
```diff
@@ -532,7 +559,21 @@ private <ValT extends Object, ReqT extends Message, ResT extends Message> Provid
...
+
+        if (tracer != null) {
+            final Span span = tracer.spanBuilder("resolve")
+                    .setSpanKind(SpanKind.CLIENT)
+                    .startSpan();
+            span.setAttribute("feature_flag.key", key);
+            try (Scope scope = span.makeCurrent()) {
+                response = resolverRef.apply((ReqT) req);
+            } finally {
+                span.end();
```
thisthat, line 573: here I would inspect the response to add other attributes, ie ff.variant or exceptions, and set the span status in case of error.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
```diff
@@ -532,7 +559,21 @@ private <ValT extends Object, ReqT extends Message, ResT extends Message> Provid
...
 
         // run the referenced resolver method
-        ResT response = resolverRef.apply((ReqT) req);
+        final ResT response;
+
+        if (tracer != null) {
+            final Span span = tracer.spanBuilder("resolve")
+                    .setSpanKind(SpanKind.CLIENT)
+                    .startSpan();
+            span.setAttribute("feature_flag.key", key);
```
thisthat, line 568:
```suggestion
            span.setAttribute("feature_flag.key", key);
            span.setAttribute("feature_flag.provider_name", "flagd");
```

thisthat:
> Just out of curiosity, can't this instrumentation be implemented using [hooks](https://openfeature.dev/specification/sections/hooks)?

Hey @thiyagu06, not really. With the hooks, you cannot instrument the provider to propagate the W3C context.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
Kavindu-Dodan: I am opting for a builder pattern-based approach [1] I will do this in a dedicated PR to reduce complexity :) 

[1] - https://github.com/open-feature/java-sdk-contrib/pull/294/files#diff-04582b0607c583d36a31ea656e9229beceab5f2abc3139710831765da374ce89R43-R47

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
Kavindu-Dodan: Right now we do not convey any evaluation error through the response. So this is not possible. See the protobuf definition of the contract [1] 

[1] - https://github.com/open-feature/schemas/blob/main/protobuf/schema/v1/schema.proto

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
Kavindu-Dodan: Updated with the latest commit :)

thiyagu06:
> > Just out of curiosity, can't this instrumentation be implemented using [hooks](https://openfeature.dev/specification/sections/hooks)?
> 
> Hey @thiyagu06, not really. With the hooks, you cannot instrument the provider to propagate the W3C context.

Yes. got it. My concern here is, for an application authors who uses an auto instrumentation [of open telemetry](https://opentelemetry.io/docs/instrumentation/java/automatic/), creating an instance of OpenTelemetrySdk and passing it as an constructor argument `api.setProvider(new FlagdProvider(telemetrySdk));` doesn't look good. 

Instead, I would prefer to obtain the opentelemetry instance by calling `GlobalOpenTelemetry.get()` in `providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java` and perform the operation which are needed. any thoughts?

thisthat:
>  My concern here is, for an application authors who uses an auto instrumentation [of open telemetry](https://opentelemetry.io/docs/instrumentation/java/automatic/), creating an instance of OpenTelemetrySdk and passing it as an constructor argument `api.setProvider(new FlagdProvider(telemetrySdk));` doesn't look good.
> 
> Instead, I would prefer to obtain the opentelemetry instance by calling `GlobalOpenTelemetry.get()` in `providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java` and perform the operation which are needed. any thoughts?

Gotcha! Thanks for noticing it. I agree with you. We should not use the SDK part of OTel here and instead only rely on the API since we are a [third party library](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/library-guidelines.md#requirements) 👍

Kavindu-Dodan:
@thisthat & @thiyagu06 

I re-checked the documentation and the recommended way it to pass down an instance of `OpenTelemetry` instead of `OpenTelemetrySDK` [1]. Usage of `GlobalOpenTelemetry` is discouraged. 

> As an aside, if you are writing library instrumentation, it is strongly recommended that you provide your users the ability to inject an instance of OpenTelemetry into your instrumentation code. If this is not possible for some reason, you can fall back to using an instance from the GlobalOpenTelemetry class. Note that you can’t force end-users to configure the global, so this is the most brittle option for library instrumentation.

I will update this and this will allow us to get rid of sdk dependency. 

[1] - https://opentelemetry.io/docs/instrumentation/java/manual/#example

toddbaert (approved):
After reading your docs I think I agree with this solution. Thanks for taking the time.

I think you may want to add some documentation about this to the README, and maybe even some of the links from [this comment](https://github.com/open-feature/java-sdk-contrib/pull/286#issuecomment-1521935162).

Approving but I'd like to hear once again from @thiyagu06 or @thisthat .

thisthat (approved):

Kavindu-Dodan:
> After reading your docs I think I agree with this solution. Thanks for taking the time.
> 
> I think you may want to add some document ion about this to the README, and maybe even some of the links from [this comment](https://github.com/open-feature/java-sdk-contrib/pull/286#issuecomment-1521935162).
> 
> Approving but I'd like to hear once again from @thiyagu06 or @thisthat .

Thank you 🙌 

Regarding the documentation, yes agree with the input. I will include an explanation with the next PR https://github.com/open-feature/java-sdk-contrib/pull/294 where I try to improve the overall usage of constructors :)

toddbaert:
> > After reading your docs I think I agree with this solution. Thanks for taking the time.
> > I think you may want to add some document ion about this to the README, and maybe even some of the links from [this comment](https://github.com/open-feature/java-sdk-contrib/pull/286#issuecomment-1521935162).
> > Approving but I'd like to hear once again from @thiyagu06 or @thisthat .
> 
> Thank you raised_hands
> 
> Regarding the documentation, yes agree with the input. I will include an explanation with the next PR #294 where I try to improve the overall usage of constructors :)

Works for me!

PR #287, opened 2023-04-22
feat: replace OS class with interface

### LINKED ISSUE #257: [env-var-provider] Replace OS class with an interface
See https://github.com/open-feature/java-sdk-contrib/pull/244#discussion_r1147943367

> Minor comment (potential improvement)
> 
> We could make this an interface (contract) and have a default implementation based on System.getenv(name)
> 
> That would allow having custom env variable handling (ex:- prefixes, casing and even sources like spring configurations which come from different sources [1] ) and still use the same provider
> 
> Could be a follow up work :)
>
> [1] - https://docs.spring.io/spring-boot/docs/1.0.1.RELEASE/reference/html/boot-features-external-config.html

lopitz:
## This PR
-introduces a new interface abstracting the access of environment variables

### Related Issues
Fixes #257

### Notes
I left the default constructor of the class `EnvVarProvider` intact, in case it is already used somewhere. If it is ok, I'd rather kill it, but this would introduce a breaking change.

### How to test
The existing unit tests now uses an implementation of the newly created interface (as lambda, as it is a functional interface). No additional testing is required as the interface doesn't contain any logic.

### DIFF
--- a/providers/env-var/src/main/java/dev/openfeature/contrib/providers/envvar/EnvironmentGateway.java
+++ b/providers/env-var/src/main/java/dev/openfeature/contrib/providers/envvar/EnvironmentGateway.java
@@ -1,7 +1,7 @@
 package dev.openfeature.contrib.providers.envvar;
 
 /**
- * This is an abstraction to fetch an environment variables. It can be used to support
+ * This is an abstraction to fetch environment variables. It can be used to support
  * environment-specific access or provide additional functionality, like prefixes,
  * casing and even sources like spring configurations which come from different sources.
  * Also, a test double could implement this interface, making the tests independent of the actual environment.
--- a/providers/env-var/src/main/java/dev/openfeature/contrib/providers/envvar/EnvironmentGateway.java
+++ b/providers/env-var/src/main/java/dev/openfeature/contrib/providers/envvar/EnvironmentGateway.java
@@ -1,0 +1,11 @@
+package dev.openfeature.contrib.providers.envvar;
+
+/**
+ * This is an abstraction to fetch an environment variables. It can be used to support
+ * environment-specific access or provide additional functionality, like prefixes,
+ * casing and even sources like spring configurations which come from different sources.
+ * Also, a test double could implement this interface, making the tests independent of the actual environment.
+ */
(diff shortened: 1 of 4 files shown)


providers/env-var/src/main/java/dev/openfeature/contrib/providers/envvar/EnvironmentGateway.java
```diff
@@ -0,0 +1,11 @@
+package dev.openfeature.contrib.providers.envvar;
+
+/**
+ * This is an abstraction to fetch an environment variables. It can be used to support
```
toddbaert:
```suggestion
 * This is an abstraction to fetch environment variables. It can be used to support
```

toddbaert:
> I left the default constructor of the class EnvVarProvider intact, in case it is already used somewhere. If it is ok, I'd rather kill it, but this would introduce a breaking change.

I think I like having a very simple default implementation, but I don't feel too strongly about it. Would you rather force the consumer to implement the `EnvironmentGateway`?

toddbaert (approved):
@lopitz LGTM. I changed the PR title to `feat` which will increment the version indicating there's a new feature, and denote that in the changelog. I also removed the `[env-var-provider]` bit as release-please will detect the altered path an indicate the change only impacts this component.

providers/env-var/src/main/java/dev/openfeature/contrib/providers/envvar/EnvironmentGateway.java
```diff
@@ -0,0 +1,11 @@
...
+ * environment-specific access or provide additional functionality, like prefixes,
+ * casing and even sources like spring configurations which come from different sources.
+ * Also, a test double could implement this interface, making the tests independent of the actual environment.
+ */
+public interface EnvironmentGateway {
```
Kavindu-Dodan, line 9: We can mark this as a `@FunctionalInterface` and make the contract even more strict :)

Kavindu-Dodan (approved):
Approving with minor remark :)

Kavindu-Dodan:
We have approvals and I am merging this. Thank you @lopitz  for the contribution

PR #288, opened 2023-04-22
feat: provide key transformations

### LINKED ISSUE #256: [env-var-provider] Suppot keys transformation
See https://github.com/open-feature/java-sdk-contrib/pull/244#discussion_r1146778474:

> One thing we've done in other ENV VAR providers is to map SCREAMING_SNAKE_CASE env vars to hyphen-case-keys. This may be desirable, since generally people like to use SCREAMIN_SNAKE for env vars, but that's generally not true for feature flag keys.
>
> You could also create some transformer interface to allow users to control this translation.... and it doesn't have to be done in this PR.

lopitz:
## This PR
adds generic and well known key transformations. This is needed, if environment variable patterns deviate from feature flag key patterns, e.g. SCREMING_SNAKE_CASE for environment keys vs. hypen-case feature flag keys.

### Related Issues
Fixes #256 

### Notes
This builds upon PR #287.

This is  a more-functional approach. I was playing around with the chain of responsibility, but it was less readable (from my point of view).

### How to test
Unit tests and examples are provided.

### DIFF
--- a/providers/env-var/src/main/java/dev/openfeature/contrib/providers/envvar/EnvironmentKeyTransformer.java
+++ b/providers/env-var/src/main/java/dev/openfeature/contrib/providers/envvar/EnvironmentKeyTransformer.java
@@ -13,7 +13,7 @@ import org.apache.commons.text.CaseUtils;
  *
  * <p>This class also supports chaining/combining different transformers incl. self-written ones by providing
  * a transforming function in the constructor. <br>
- * Currently the following transformations are supported out of the box:
+ * Currently, the following transformations are supported out of the box:
  * <ul>
  *     <li>{@link #toLowerCaseTransformer() converting to lower case}</li>
  *     <li>{@link #toUpperCaseTransformer() converting to UPPER CASE}</li>
@@ -29,23 +29,21 @@ import org.apache.commons.text.CaseUtils;
  * <pre>
  * {@code
  * // Definition of the EnvVarProvider:
- * EnvironmentGateway transformingGateway = EnvironmentKeyTransformer
- *     .hyphenCaseToScreamingSnake()
- *     .withSource(new OS());
+ * EnvironmentKeyTransformer transformer = EnvironmentKeyTransformer
+ *     .hyphenCaseToScreamingSnake();
  *
- * FeatureProvider provider = new EnvVarProvider(transformingGateway);
+ * FeatureProvider provider = new EnvVarProvider(transformer);
  * }
  * </pre>
  * 2. chained/composed transformations:
  * <pre>
  * {@code
  * // Definition of the EnvVarProvider:
- * EnvironmentGateway transformingGateway = EnvironmentKeyTransformer
+ * EnvironmentKeyTransformer transformer = EnvironmentKeyTransformer
  *     .toLowerCaseTransformer()
- *     .andThen(EnvironmentKeyTransformer.replaceUnderscoreWithDotTransformer())
- *     .withSource(actualGateway);
+ *     .andThen(EnvironmentKeyTransformer.replaceUnderscoreWithDotTransformer());
  *
- * FeatureProvider provider = new EnvVarProvider(transformingGateway);
+ * FeatureProvider provider = new EnvVarProvider(transformer);
  * }
  * </pre>
  * 3. freely defined transformation function:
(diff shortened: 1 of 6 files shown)


providers/env-var/src/main/java/dev/openfeature/contrib/providers/envvar/EnvironmentKeyTransformer.java
```diff
@@ -0,0 +1,113 @@
...
+
+    private final Function<String, String> transformation;
+
+    @Override
+    public String getEnvironmentVariable(String key) {
+        return transformation.apply(key);
+    }
+
+    public EnvironmentKeyTransformer andThen(EnvironmentGateway another) {
+        return new EnvironmentKeyTransformer(transformation.compose(another::getEnvironmentVariable));
```
Kavindu-Dodan: Minor - assuming the intention here is to first apply this transformation and then apply the `another`, we can use `andThen` instead of `compose` [1]. But let me know if the idea is different here :) 

[1] - https://docs.oracle.com/javase/8/docs/api/java/util/function/Function.html#andThen-java.util.function.Function-

providers/env-var/src/main/java/dev/openfeature/contrib/providers/envvar/EnvironmentKeyTransformer.java
```diff
@@ -0,0 +1,113 @@
+package dev.openfeature.contrib.providers.envvar;
```
Kavindu-Dodan, line 1: Implementation is great 🙌

Kavindu-Dodan (approved):
Awesome contribution. Approving with minor remarks.!

providers/env-var/src/main/java/dev/openfeature/contrib/providers/envvar/EnvironmentKeyTransformer.java
lopitz: i think, looking at the context and the names of the methods i chose, `andThen` makes more sense 😄 . thx for pointing it out.

providers/env-var/src/main/java/dev/openfeature/contrib/providers/envvar/EnvironmentKeyTransformer.java
```diff
@@ -0,0 +1,113 @@
...
+
+import java.util.function.*;
+
+import lombok.RequiredArgsConstructor;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.text.CaseUtils;
+
+/**
+ * This class provides a way to transform any given key to another value. This is helpful, if keys in the code have a
```
thiyagu06, line 10: would be nice if we add an info about this capability in readme.

thiyagu06 (approved):
Approving with comments on adding info about this capability in readme.

providers/env-var/src/main/java/dev/openfeature/contrib/providers/envvar/EnvironmentKeyTransformer.java
lopitz: thanks for pointing it out. it actually forced me to change the implementation a little to avoid additional configuration when using key transformations.

The usage is added to the README.md

beeme1mr (approved):

toddbaert (approved):
Looks great! Great documentation too.

providers/env-var/src/main/java/dev/openfeature/contrib/providers/envvar/EnvironmentKeyTransformer.java
```diff
@@ -0,0 +1,109 @@
...
+    public static EnvironmentKeyTransformer replaceDotWithUnderscoreTransformer() {
+        return new EnvironmentKeyTransformer(REPLACE_DOT_WITH_UNDERSCORE);
+    }
+
+    public static EnvironmentKeyTransformer hyphenCaseToScreamingSnake() {
+        return new EnvironmentKeyTransformer(REPLACE_HYPHEN_WITH_UNDERSCORE)
+            .andThen(EnvironmentKeyTransformer.toUpperCaseTransformer());
+    }
+
+    public static EnvironmentKeyTransformer doNothing() {
```
thiyagu06, line 106: NoOp is the convention I see it in open feature SDK implementation for this kind of scenarios. I'm okay with `doNothing` as well. @Kavindu-Dodan @toddbaert any thoughts on this?

Kavindu-Dodan: Good point. Thinking deeper here, this is an identity function similar to `Function.identity` [1] rather than a `NoOp`(performing nothing vs returning input itself)

Maybe we can rename(potentially breaking change)/deprecate & introduce a new function () for a later release :)

[1] - https://docs.oracle.com/javase/8/docs/api/java/util/function/Function.html#identity--

thiyagu06: yes.. I prefer to deprecate the `doNothing` and use the `Function.identity`. I will find sometime during the week and open a PR.

lopitz: I explicitly wanted to provide a `doNothing`/`noOp` transformer to let the consumer of the api not to bother, if they don't need to. Having said that, using `Function.identity` instead of `s -> s` is certainly an improvement. 

So, I would like to see a dedicated `doNothing` or `noOp`/`NO_OP` transformer and the replacement of `s -> s` with `Function.identity`. What are your thoughts?

Kavindu-Dodan: @lopitz I feel this as a good improvement :) feel free to open a PR for this

PR #294, opened 2023-04-24
chore: flagd constructor refactoring

Kavindu-Dodan:
A follow-up of #286
___

## This PR

Is a refactoring for flagd provider constructor. Current constructors are difficult to maintain and with more options, we run into complexity of breaking the construction contract. Plus, mix & match of args becomes difficult (reason for #286 to have a simple constructor to get feature done)

This PR introduces a builder. The builder builds argument options with built-in default derivations (instead of relying on nulls constructor args).

Example,

```java
FlagdProvider flagdProvider =
        new FlagdProvider(FlagdOptions.builder()
                            .host("localhost")
                            .port(8023)
                            .openTelemetry(openTelemetry).build());
```

Along with the changes, I am deprecating `old` constructors. We shall aim to keep them as `@Deprecated` for next release. And then, we shall remove them from the release after. 

Follow-up efforts - unit tests and improvements on testing

### DIFF
--- a/providers/flagd/pom.xml
+++ b/providers/flagd/pom.xml
@@ -10,7 +10,7 @@
     </parent>
     <groupId>dev.openfeature.contrib.providers</groupId>
     <artifactId>flagd</artifactId>
-    <version>0.5.8-snapshot</version> <!--x-release-please-version -->
+    <version>0.5.8</version> <!--x-release-please-version -->
 
     <name>flagd</name>
     <description>FlagD provider for Java</description>
@@ -38,7 +38,7 @@
             <!-- we only support unix sockets on linux, via epoll native lib -->
             <groupId>io.netty</groupId>
             <artifactId>netty-transport-native-epoll</artifactId>
-            <version>4.1.92.Final</version>
+            <version>4.1.91.Final</version>
             <!-- TODO: with 5+ (still alpha), arm is support and we should package multiple versions -->
             <classifier>linux-x86_64</classifier>
         </dependency>
@@ -71,7 +71,7 @@
 
         <dependency>
             <groupId>io.opentelemetry</groupId>
-            <artifactId>opentelemetry-api</artifactId>
+            <artifactId>opentelemetry-sdk</artifactId>
             <version>1.25.0</version>
         </dependency>
     </dependencies>
--- a/providers/flagd/pom.xml
+++ b/providers/flagd/pom.xml
@@ -10,7 +10,7 @@
     </parent>
     <groupId>dev.openfeature.contrib.providers</groupId>
     <artifactId>flagd</artifactId>
-    <version>0.5.8</version> <!--x-release-please-version -->
+    <version>0.5.8-snapshot</version> <!--x-release-please-version -->
 
     <name>flagd</name>
     <description>FlagD provider for Java</description>
@@ -71,7 +71,7 @@
 
         <dependency>
             <groupId>io.opentelemetry</groupId>
-            <artifactId>opentelemetry-sdk</artifactId>
+            <artifactId>opentelemetry-api</artifactId>
             <version>1.25.0</version>
         </dependency>
     </dependencies>
--- a/providers/flagd/pom.xml
+++ b/providers/flagd/pom.xml
@@ -10,7 +10,7 @@
     </parent>
     <groupId>dev.openfeature.contrib.providers</groupId>
     <artifactId>flagd</artifactId>
-    <version>0.5.8-snapshot</version> <!--x-release-please-version -->
+    <version>0.5.8</version> <!--x-release-please-version -->
 
     <name>flagd</name>
     <description>FlagD provider for Java</description>
@@ -38,7 +38,7 @@
             <!-- we only support unix sockets on linux, via epoll native lib -->
             <groupId>io.netty</groupId>
             <artifactId>netty-transport-native-epoll</artifactId>
-            <version>4.1.91.Final</version>
+            <version>4.1.92.Final</version>
             <!-- TODO: with 5+ (still alpha), arm is support and we should package multiple versions -->
             <classifier>linux-x86_64</classifier>
         </dependency>
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdGrpcInterceptor.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdGrpcInterceptor.java
@@ -7,48 +7,55 @@ import io.grpc.ClientInterceptor;
 import io.grpc.ForwardingClientCall;
 import io.grpc.Metadata;
 import io.grpc.MethodDescriptor;
-import io.opentelemetry.api.OpenTelemetry;
 import io.opentelemetry.context.Context;
 import io.opentelemetry.context.propagation.TextMapSetter;
+import io.opentelemetry.sdk.OpenTelemetrySdk;
 
 import javax.annotation.Nullable;
 
 /**
- * FlagdGrpcInterceptor is an interceptor for grpc communication from java-sdk to flagd.
- * <a href="https://github.com/open-telemetry/opentelemetry-java-docs">credits</a>
+ * FlagdGrpcInterceptor is an interceptor for grpc communication from java-sdk to flagd
+ * <p>
+ *  <a href="https://github.com/open-telemetry/opentelemetry-java-docs">credits</a>
  */
 final class FlagdGrpcInterceptor implements ClientInterceptor {
     private static final TextMapSetter<Metadata> SETTER = new Setter();
 
-    private final OpenTelemetry openTelemetry;
+    private final OpenTelemetrySdk openTelemetry;
 
-    FlagdGrpcInterceptor(final OpenTelemetry openTelemetry) {
+    FlagdGrpcInterceptor(final OpenTelemetrySdk openTelemetry) {
         this.openTelemetry = openTelemetry;
     }
 
-    @Override public <ReqT, RespT> ClientCall<ReqT, RespT> interceptCall(MethodDescriptor<ReqT, RespT> methodDescriptor,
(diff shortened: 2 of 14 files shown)


providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdGrpcInterceptor.java
```diff
@@ -0,0 +1,56 @@
...
+import io.grpc.CallOptions;
+import io.grpc.Channel;
+import io.grpc.ClientCall;
+import io.grpc.ClientInterceptor;
+import io.grpc.ForwardingClientCall;
+import io.grpc.Metadata;
+import io.grpc.MethodDescriptor;
+import io.opentelemetry.context.Context;
+import io.opentelemetry.context.propagation.TextMapSetter;
+import io.opentelemetry.sdk.OpenTelemetrySdk;
```
thisthat: OpenTelemetry recommends only relying on the API for 3rd party libraries. Only App developer should use an implementation of the SDK. This way, you won't get into the case where different SDKs are loaded into the runtime, creating a mess in handling the context.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdOptions.java
```diff
@@ -0,0 +1,86 @@
...
+     * */
+    @Builder.Default
+    private int maxEventStreamRetries =
+            fallBackToEnvOrDefault(MAX_EVENT_STREAM_RETRIES_ENV_VAR_NAME, DEFAULT_MAX_EVENT_STREAM_RETRIES);
+
+    /**
+     * Inject OpenTelemetrySdk for the library runtime. Providing sdk will initiate distributed tracing for flagd grpc
+     * connectivity.
+     * */
+    private OpenTelemetrySdk telemetrySdk;
```
thisthat: if we rely on the Global instance of OpenTelemetry, we don't have to store this object: https://github.com/open-telemetry/opentelemetry-java/blob/main/api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java#L69

thisthat:
Thanks for the refactoring @Kavindu-Dodan  👍 I've created https://github.com/open-feature/java-sdk-contrib/issues/295 to track the removal of the deprecated methods

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdOptions.java
thiyagu06: Recommendation as per javadoc https://github.com/open-telemetry/opentelemetry-java/blob/main/api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java#L38, we need to store the reference.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdGrpcInterceptor.java
Kavindu-Dodan: This is fixed with the latest commits and migrated to `OpenTelemetry`

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdOptions.java
Kavindu-Dodan: Yes, recommendation is to pass down a reference of the OpenTelemetry. Using `GlobalOpenTelemetry.get()` is discouraged [1]

[1] - https://opentelemetry.io/docs/instrumentation/java/manual/#example

Kavindu-Dodan:
> Thanks for the refactoring @Kavindu-Dodan 👍 I've created #295 to track the removal of the deprecated methods

This is great :) thank you .!

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdGrpcInterceptor.java
```diff
@@ -0,0 +1,56 @@
...
+
+    @Override public <ReqT, RespT> ClientCall<ReqT, RespT> interceptCall(MethodDescriptor<ReqT, RespT> methodDescriptor,
+                                                                         CallOptions callOptions, Channel channel) {
+
+        final ClientCall<ReqT, RespT> call = channel.newCall(methodDescriptor, callOptions);
+
+        return new ForwardingClientCall.SimpleForwardingClientCall<ReqT, RespT>(call) {
+            @Override
+            public void start(Listener<RespT> responseListener, Metadata headers) {
+                openTelemetry.getPropagators().getTextMapPropagator().inject(Context.current(), headers, SETTER);
```
thisthat, line 37: I never used Lombok, but I think this is not safe since https://github.com/open-feature/java-sdk-contrib/pull/294/files#diff-f368ee4f4edf9090cfaf37caa1e694ef41d8e5753ac86a3dc05f1381f133bc0dR85 doesn't have a default value

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdGrpcInterceptor.java
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * FlagdGrpcInterceptor is an interceptor for grpc communication from java-sdk to flagd.
+ * <a href="https://github.com/open-telemetry/opentelemetry-java-docs">credits</a>
+ */
+final class FlagdGrpcInterceptor implements ClientInterceptor {
+    private static final TextMapSetter<Metadata> SETTER = new Setter();
+
+    private final OpenTelemetry openTelemetry;
+
+    FlagdGrpcInterceptor(final OpenTelemetry openTelemetry) {
```
thisthat: Maybe we can add `@NotNull` so it is easier to understand the contract of the class

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/TracedResolving.java
```diff
@@ -0,0 +1,37 @@
...
+import java.util.function.Function;
+
+/**
+ * {@link TracedResolving} a request to response resolver with tracing for telemetry.
+ */
+public class TracedResolving implements ResolveStrategy {
+
+    private final Tracer tracer;
+
+    TracedResolving(OpenTelemetry telemetry) {
```
thisthat: ditto

thisthat (approved):
LGTM, I left only minor comments

providers/flagd/pom.xml
```diff
@@ -10,7 +10,7 @@
     </parent>
     <groupId>dev.openfeature.contrib.providers</groupId>
     <artifactId>flagd</artifactId>
-    <version>0.5.8-snapshot</version> <!--x-release-please-version -->
```
Kavindu-Dodan, line 13: this was a mistake 😞

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdGrpcInterceptor.java
Kavindu-Dodan: added the annotation :)

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/TracedResolving.java
Kavindu-Dodan: added the annotation :)

Kavindu-Dodan:
@toddbaert  as agreed [1] , I added (and formatted) the readme in this PR  [2]

[1] - https://github.com/open-feature/java-sdk-contrib/pull/286#issuecomment-1522337078
[2] - https://github.com/open-feature/java-sdk-contrib/blob/f5db2818ce4c269d36b3a6ff61a10e6bb962c708/providers/flagd/README.md#opentelemetry-support

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/ResolveStrategy.java
```diff
@@ -0,0 +1,13 @@
+package dev.openfeature.contrib.providers.flagd;
+
+import com.google.protobuf.Message;
+
+import java.util.function.Function;
+
+/**
+ * Request to Response resolving strategy.
+ * */
+public interface ResolveStrategy {
```
toddbaert:
```suggestion
interface ResolveStrategy {
```

These should all be package private, I think.

Kavindu-Dodan: Good one, yes . I will update with next commit :)

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
```diff
@@ -554,54 +586,41 @@ private <ValT extends Object, ReqT extends Message, ResT extends Message> Provid
...
-            span.setAttribute("feature_flag.provider_name", "flagd");
-            try (Scope scope = span.makeCurrent()) {
-                response = resolverRef.apply((ReqT) req);
-            } finally {
-                span.end();
-            }
-        } else {
-            response = resolverRef.apply((ReqT) req);
-        }
+        Message response = strategy.resolve(resolverRef, req, key);
```
toddbaert, line 594: Thanks for explaining this to me!

toddbaert (approved):

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/ResolveStrategy.java
Kavindu-Dodan: corrected with the latest commit :)

toddbaert (approved):

PR #332, opened 2023-05-30
feat!: otel metric hook

### LINKED ISSUE #329: [FEATURE] Implement OTel metrics hook
**Background**

Consider OFEP [1] - https://github.com/open-feature/ofep/pull/62 

Go-SDK hook implementation is ready[2] and this issue focuses on Java-sdk implementation of the hook.

**Outcome**

OTel hooks for metrics as define by OFEP

[1] - https://github.com/open-feature/ofep/pull/62 
[2] - https://github.com/open-feature/go-sdk-contrib/pull/217

Kavindu-Dodan:
## This PR

Fixes #329 by implementing the OTel metrics hook OFEP proposed through https://github.com/open-feature/ofep/pull/62 

⚠️ Breaking change - Renaming Traces and Metrics classes to give better meaning

- Library (`dev.openfeature.contrib.hooks`) version is updated to 2.0.0
- Renamed `OpenTelemetryHook` to `TracesHook`
- Renamed `OpenTelemetryHookOptions` to `TracesHookOptions`

### DIFF
--- a/hooks/open-telemetry/pom.xml
+++ b/hooks/open-telemetry/pom.xml
@@ -28,9 +28,13 @@
 		<!-- we inherent dev.openfeature.javasdk and the test dependencies from the parent pom -->
 		<dependency>
 			<groupId>io.opentelemetry</groupId>
-			<artifactId>opentelemetry-sdk</artifactId>
+			<artifactId>opentelemetry-api</artifactId>
 		</dependency>
-
+		<dependency>
+			<groupId>io.opentelemetry</groupId>
+			<artifactId>opentelemetry-sdk-testing</artifactId>
+			<scope>test</scope>
+		</dependency>
 	</dependencies>
 
 	<dependencyManagement>
--- a/hooks/open-telemetry/pom.xml
+++ b/hooks/open-telemetry/pom.xml
@@ -9,7 +9,7 @@
 	</parent>
 	<groupId>dev.openfeature.contrib.hooks</groupId>
 	<artifactId>otel</artifactId>
-	<version>1.0.3</version> <!--x-release-please-version -->
+	<version>2.0.0</version> <!--x-release-please-version -->
 
 	<name>open-telemetry-hook</name>
 	<description>Open Telemetry Hook</description>
--- a/hooks/open-telemetry/pom.xml
+++ b/hooks/open-telemetry/pom.xml
@@ -1,51 +1,59 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
-	<modelVersion>4.0.0</modelVersion>
-	<parent>
-		<groupId>dev.openfeature.contrib</groupId>
-		<artifactId>parent</artifactId>
-		<version>0.1.0</version>
-		<relativePath>../../pom.xml</relativePath>
-	</parent>
-	<groupId>dev.openfeature.contrib.hooks</groupId>
-	<artifactId>otel</artifactId>
-	<version>2.0.0</version> <!--x-release-please-version -->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>dev.openfeature.contrib</groupId>
+        <artifactId>parent</artifactId>
+        <version>0.1.0</version>
+        <relativePath>../../pom.xml</relativePath>
+    </parent>
+    <groupId>dev.openfeature.contrib.hooks</groupId>
+    <artifactId>otel</artifactId>
+    <version>2.0.0</version> <!--x-release-please-version -->
 
-	<name>open-telemetry-hook</name>
-	<description>Open Telemetry Hook</description>
-	<url>https://openfeature.dev</url>
+    <name>open-telemetry-hook</name>
+    <description>Open Telemetry Hook</description>
+    <url>https://openfeature.dev</url>
 
-	<developers>
-		<developer>
-			<id>toddbaert</id>
-			<name>Todd Baert</name>
-			<organization>OpenFeature</organization>
-			<url>https://openfeature.dev/</url>
-		</developer>
-	</developers>
+    <developers>
+        <developer>
+            <id>toddbaert</id>
+            <name>Todd Baert</name>
+            <organization>OpenFeature</organization>
+            <url>https://openfeature.dev/</url>
+        </developer>
+    </developers>
 
-	<dependencyManagement>
-		<dependencies>
-			<dependency>
-				<groupId>io.opentelemetry</groupId>
-				<artifactId>opentelemetry-bom</artifactId>
-				<version>1.28.0</version>
-				<type>pom</type>
-				<scope>import</scope>
-			</dependency>
-		</dependencies>
-	</dependencyManagement>
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>io.opentelemetry</groupId>
+                <artifactId>opentelemetry-bom</artifactId>
+                <version>1.28.0</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
 
-	<dependencies>
-		<dependency>
-			<groupId>io.opentelemetry</groupId>
-			<artifactId>opentelemetry-api</artifactId>
-		</dependency>
-		<dependency>
-			<groupId>io.opentelemetry</groupId>
-			<artifactId>opentelemetry-sdk-testing</artifactId>
-			<scope>test</scope>
-		</dependency>
-	</dependencies>
+    <dependencies>
+        <dependency>
+            <groupId>dev.openfeature</groupId>
+            <artifactId>sdk</artifactId>
+            <version>[1.4,2.0)</version>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>io.opentelemetry</groupId>
+            <artifactId>opentelemetry-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.opentelemetry</groupId>
+            <artifactId>opentelemetry-sdk-testing</artifactId>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
 
 </project>
--- a/hooks/open-telemetry/pom.xml
+++ b/hooks/open-telemetry/pom.xml
@@ -24,19 +24,6 @@
 		</developer>
 	</developers>
 
-	<dependencies>
-		<!-- we inherent dev.openfeature.javasdk and the test dependencies from the parent pom -->
-		<dependency>
-			<groupId>io.opentelemetry</groupId>
-			<artifactId>opentelemetry-api</artifactId>
-		</dependency>
-		<dependency>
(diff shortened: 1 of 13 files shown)


pom.xml
```diff
@@ -55,7 +55,7 @@
             <groupId>dev.openfeature</groupId>
             <artifactId>sdk</artifactId>
             <!-- 1.0 <= v < 2.0 -->
-            <version>[1.0,2.0)</version>
+            <version>1.4.0-SNAPSHOT</version>
```
Kavindu-Dodan: Development only - must be removed when review ready

hooks/open-telemetry/src/main/java/dev/openfeature/contrib/hooks/otel/OpenTelemetryMetricHook.java
```diff
@@ -0,0 +1,167 @@
...
+import static dev.openfeature.contrib.hooks.otel.OtelCommons.providerNameAttributeKey;
+import static dev.openfeature.contrib.hooks.otel.OtelCommons.variantAttributeKey;
+
+
+/**
+ * OpenTelemetry metric hook records metrics at different {@link Hook} stages.
+ */
+@Slf4j
+@SuppressWarnings("PMD.TooManyStaticImports")
+public class OpenTelemetryMetricHook implements Hook {
```
toddbaert:
```suggestion
public class MetricsHook implements Hook {
```

I think `dev.openfeature.contrib.hooks.otel` conveys well-enough this is an OTel component, so we don't need it in the class name.

Could also be `MetricHook` not sure which pluralization is better.

hooks/open-telemetry/src/main/java/dev/openfeature/contrib/hooks/otel/OpenTelemetryHook.java
```diff
@@ -3,27 +3,24 @@
...
+import static dev.openfeature.contrib.hooks.otel.OtelCommons.EVENT_NAME;
+import static dev.openfeature.contrib.hooks.otel.OtelCommons.flagKeyAttributeKey;
+import static dev.openfeature.contrib.hooks.otel.OtelCommons.providerNameAttributeKey;
+import static dev.openfeature.contrib.hooks.otel.OtelCommons.variantAttributeKey;
+
 /**
  * The OpenTelemetry hook provides a way to automatically add a feature flag evaluation to a span as a span event.
  * Refer to <a href="https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/feature-flags.md">OpenTelemetry</a>
  */
 public class OpenTelemetryHook implements Hook {
```
toddbaert:
```suggestion
public class TraceHook implements Hook {
```
TracingHook could also be OK, but I think we should make this a breaking change and rename this old hook to be more precise.

hooks/open-telemetry/src/main/java/dev/openfeature/contrib/hooks/otel/OpenTelemetryMetricHook.java
```diff
@@ -0,0 +1,167 @@
...
+                        .build();
+
+        evaluationRequestCounter = meter.counterBuilder(EVALUATION_REQUESTS_TOTAL)
+                .setDescription("feature flag evaluation request counter").build();
+
+        evaluationSuccessCounter = meter.counterBuilder(FLAG_EVALUATION_SUCCESS_TOTAL)
+                .setDescription("feature flag evaluation success counter").build();
+
+        evaluationErrorCounter = meter.counterBuilder(FLAG_EVALUATION_ERROR_TOTAL)
+                .setDescription("feature flag evaluation error counter").build();
```
toddbaert: Are there metric standard conventions in OTel that allow us to standardize these descriptions?

I noticed these don't use capitals/punctualtion (both here and in Go). Personally, I like that, but I'm wondering if there's a recommended style?

Kavindu-Dodan: This is optional and does not have any specific format [1]

> The description is an optional free-form text provided by the author of the instrument. The API MUST treat it as an opaque string.

So I don't think we need any formatting here

[1] - https://opentelemetry.io/docs/specs/otel/metrics/api/#instrument-description

toddbaert: Cool. I'm fine with keeping them as is, since they are sentence fragments. If anyone has any strong preference for capitalization/punctuation, I will let them make that argument.

Kavindu-Dodan: Thank you. I am open for any suggestions here :)

hooks/open-telemetry/src/main/java/dev/openfeature/contrib/hooks/otel/OpenTelemetryMetricHook.java
Kavindu-Dodan: Agree with this proposal 😞  but I think we already have the OpenTelemetry prefix in existing classes [1]. Changing only this is odd.

[1] - https://github.com/open-feature/java-sdk-contrib/tree/main/hooks/open-telemetry/src/main/java/dev/openfeature/contrib/hooks/otel

hooks/open-telemetry/src/main/java/dev/openfeature/contrib/hooks/otel/OpenTelemetryHook.java
Kavindu-Dodan: See the comment here [1]

[1] - https://github.com/open-feature/java-sdk-contrib/pull/332#discussion_r1254989278

hooks/open-telemetry/src/main/java/dev/openfeature/contrib/hooks/otel/OpenTelemetryMetricHook.java
toddbaert: > Agree with this proposal disappointed but I think we already have the OpenTelemetry prefix in existing classes [1]. Changing only this is odd.
> 
> [1] - https://github.com/open-feature/java-sdk-contrib/tree/main/hooks/open-telemetry/src/main/java/dev/openfeature/contrib/hooks/otel

Ya I think we should release a 2.0 and rename them all. That would be my vote.

hooks/open-telemetry/src/main/java/dev/openfeature/contrib/hooks/otel/OpenTelemetryMetricHook.java
```diff
@@ -0,0 +1,167 @@
...
+
+
+/**
+ * OpenTelemetry metric hook records metrics at different {@link Hook} stages.
+ */
+@Slf4j
+@SuppressWarnings("PMD.TooManyStaticImports")
+public class OpenTelemetryMetricHook implements Hook {
+
+    private static final String METER_NAME = "go.openfeature.dev";
```
toddbaert: I'm a bit confused about this name. Why "go"?

Kavindu-Dodan: arghh, this comes from go sdk. I copied the cosntants and this came with that 😞 . will fix

hooks/open-telemetry/src/main/java/dev/openfeature/contrib/hooks/otel/OpenTelemetryMetricHook.java
Kavindu-Dodan: Do you suggest doing that along with this change 🤔  Or should we first release this and then follow up on the v2.0

toddbaert: I don't think it needs to be done with this change, but I think it should be done before this change is released. I think having an `OpenTelemetryHook` and a `OpenTelemetryMetricHook` is inherently confusing - it's not clear what the first one does.

I can create an issue for after this is merged, if you'd prefer not to do it in this PR.

Kavindu-Dodan: Agree. I would like a follow-up as this PR focus on a feature. And the follow-up is a breaking change with a refactoring

Kavindu-Dodan: I will do this change with this PR and bump the version to 2.0.0

toddbaert: Fine with me!

Kavindu-Dodan: Done - https://github.com/open-feature/java-sdk-contrib/blob/4c6f93da9e27cdc87cf666c996a0fdc23ce5cab9/hooks/open-telemetry/src/main/java/dev/openfeature/contrib/hooks/otel/MetricsHook.java

hooks/open-telemetry/src/main/java/dev/openfeature/contrib/hooks/otel/OpenTelemetryHook.java
Kavindu-Dodan: Done - https://github.com/open-feature/java-sdk-contrib/blob/4c6f93da9e27cdc87cf666c996a0fdc23ce5cab9/hooks/open-telemetry/src/main/java/dev/openfeature/contrib/hooks/otel/TracesHook.java

Kavindu-Dodan:
@toddbaert like you proposed, I have made this a breaking change. Plus refactored by renaming OTel hook classes

hooks/open-telemetry/pom.xml
```diff
@@ -28,9 +28,13 @@
 		<!-- we inherent dev.openfeature.javasdk and the test dependencies from the parent pom -->
 		<dependency>
 			<groupId>io.opentelemetry</groupId>
-			<artifactId>opentelemetry-sdk</artifactId>
+			<artifactId>opentelemetry-api</artifactId>
```
toddbaert: Do we need to specify a minimum version now? Particularly because of metrics?

toddbaert (approved):
LGTM - I think [this](https://github.com/open-feature/java-sdk-contrib/pull/332/files#r1264081220) is my only concern.

pom.xml
```diff
@@ -55,7 +55,7 @@
             <groupId>dev.openfeature</groupId>
             <artifactId>sdk</artifactId>
             <!-- 1.0 <= v < 2.0 -->
-            <version>[1.0,2.0)</version>
+            <version>[1.4.0,2.0)</version>
```
toddbaert: This will make `1.4.0` a requirement for everything. I think we want to add this section to just the otel hook pom, which will override this one.

toddbaert (changes requested):
https://github.com/open-feature/java-sdk-contrib/pull/332/files#r1264136869

hooks/open-telemetry/pom.xml
Kavindu-Dodan: This is managed by the dependency management entry [1]. I think the comment above this is misleading and dependency management placement is not ideal (event I was confused how this is resolved)


[1] - https://github.com/open-feature/java-sdk-contrib/blob/main/hooks/open-telemetry/pom.xml#L36-L46

Kavindu-Dodan: restructured the pom to reduce confusion

pom.xml
Kavindu-Dodan: hmm, I will check if this is possible

Kavindu-Dodan: Yup, it's possible - https://github.com/open-feature/java-sdk-contrib/pull/332/files#diff-04c08d30909eade5e2f0b110b60da00732703796061a62a38bdb7ae66db549caR40-R46

Updated hook pom to have this depdency

toddbaert (approved):

PR #361, opened 2023-07-20
feat!: add events support

### LINKED ISSUE #314: [flagd-provider] add events support
Once the java-sdk has `events` and `init/shutdown` support, we need to update the flagd-provider to emit events and do proper init/shutdown. We already have a stream implemented for cache-busing, so this should be pretty simple.

Definition of done:
- flagd does does startup logic (like initializing its stream) in `initialize` function
- flagd shuts down stream in `close/shutdown` function
- flagd emits appropriate events on flag changes, and errors
- associated testing

Blocked by:
- java-sdk implementation: https://github.com/open-feature/java-sdk/issues/437

Resources:
- [flag provider guide](https://github.com/open-feature/flagd/blob/main/docs/other_resources/creating_providers.md#provider-lifecycle-initialization-and-shutdown)

thisthat:
Signed-off-by: Giovanni Liva <giovanni.liva@dynatrace.com>

## This PR

- introduce support for Events

### Related Issues

Fixes #314 

### How to test

Run `mvn verify`

### DIFF
--- a/providers/flagd/pom.xml
+++ b/providers/flagd/pom.xml
@@ -26,6 +26,16 @@
     </developers>
 
     <dependencies>
+
+        <!-- override parent definition -->
+        <dependency>
+            <groupId>dev.openfeature</groupId>
+            <artifactId>sdk</artifactId>
+            <version>[1.4,2.0)</version>
+            <scope>provided</scope>
+        </dependency>
+
+
         <!-- we inherent dev.openfeature.javasdk and the test dependencies from the parent pom -->
 
         <dependency>
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/EventStreamObserver.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/EventStreamObserver.java
@@ -1,8 +1,10 @@
 package dev.openfeature.contrib.providers.flagd;
 
+import java.util.Map;
 import io.grpc.stub.StreamObserver;
 import lombok.extern.slf4j.Slf4j;
 import dev.openfeature.flagd.grpc.Schema.EventStreamResponse;
+import com.google.protobuf.Value;
 
 /**
  * EventStreamObserver handles events emitted by flagd.
@@ -14,6 +16,7 @@ public class EventStreamObserver implements StreamObserver<EventStreamResponse> 
 
     private static final String configurationChange = "configuration_change";
     private static final String providerReady = "provider_ready";
+    static final String flagsKey = "flags";
 
     EventStreamObserver(FlagdCache cache, EventStreamCallback callback) {
         this.cache = cache;
@@ -24,7 +27,7 @@ public class EventStreamObserver implements StreamObserver<EventStreamResponse> 
     public void onNext(EventStreamResponse value) {
         switch (value.getType()) {
             case configurationChange:
-                this.handleConfigurationChangeEvent();
+                this.handleConfigurationChangeEvent(value);
                 break;
             case providerReady:
                 this.handleProviderReadyEvent();
@@ -58,13 +61,23 @@ public class EventStreamObserver implements StreamObserver<EventStreamResponse> 
         this.callback.setEventStreamAlive(false);
     }
 
-    private void handleConfigurationChangeEvent() {
+    private void handleConfigurationChangeEvent(EventStreamResponse value) {
         this.callback.emitConfigurationChangeEvent();
         if (!this.cache.getEnabled()) {
             return;
         }
-        // Always flush the cache when configuration_change event is received
-        this.cache.clear();
+        Map<String, Value> data = value.getData().getFieldsMap();
+        Value flagsValue = data.get(flagsKey);
+        if (flagsValue == null) {
+            this.cache.clear();
+            return;
+        }
+
+        Map<String, Value> flags = flagsValue.getStructValue().getFieldsMap();
+
+        for (String flagKey : flags.keySet()) {
+            this.cache.remove(flagKey);
+        }
     }
 
     private void handleProviderReadyEvent() {
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/EventStreamObserver.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/EventStreamObserver.java
@@ -1,10 +1,8 @@
 package dev.openfeature.contrib.providers.flagd;
 
-import java.util.Map;
 import io.grpc.stub.StreamObserver;
 import lombok.extern.slf4j.Slf4j;
 import dev.openfeature.flagd.grpc.Schema.EventStreamResponse;
-import com.google.protobuf.Value;
 
 /**
  * EventStreamObserver handles events emitted by flagd.
@@ -16,7 +14,6 @@ public class EventStreamObserver implements StreamObserver<EventStreamResponse> 
 
     private static final String configurationChange = "configuration_change";
     private static final String providerReady = "provider_ready";
-    private static final String flagsKey = "flags";
 
     EventStreamObserver(FlagdCache cache, EventStreamCallback callback) {
         this.cache = cache;
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/EventStreamObserver.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/EventStreamObserver.java
@@ -1,6 +1,8 @@
 package dev.openfeature.contrib.providers.flagd;
 
 import java.util.Map;
+
+import dev.openfeature.sdk.ProviderState;
 import io.grpc.stub.StreamObserver;
 import lombok.extern.slf4j.Slf4j;
 import dev.openfeature.flagd.grpc.Schema.EventStreamResponse;
@@ -44,7 +46,7 @@ public class EventStreamObserver implements StreamObserver<EventStreamResponse> 
         if (this.cache.getEnabled()) {
             this.cache.clear();
         }
-        this.callback.setEventStreamAlive(false);
+        this.callback.setState(ProviderState.ERROR);
         try {
             this.callback.restartEventStream();
             this.callback.emitSuccessReconnectionEvents();
@@ -58,7 +60,7 @@ public class EventStreamObserver implements StreamObserver<EventStreamResponse> 
         if (this.cache.getEnabled()) {
             this.cache.clear();
         }
-        this.callback.setEventStreamAlive(false);
+        this.callback.setState(ProviderState.ERROR);
     }
 
     private void handleConfigurationChangeEvent(EventStreamResponse value) {
@@ -81,7 +83,7 @@ public class EventStreamObserver implements StreamObserver<EventStreamResponse> 
     }
 
     private void handleProviderReadyEvent() {
-        this.callback.setEventStreamAlive(true);
+        this.callback.setState(ProviderState.READY);
         if (this.cache.getEnabled()) {
             this.cache.clear();
         }
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/EventStreamObserver.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/EventStreamObserver.java
@@ -27,7 +27,7 @@ public class EventStreamObserver implements StreamObserver<EventStreamResponse> 
     public void onNext(EventStreamResponse value) {
         switch (value.getType()) {
             case configurationChange:
-                this.handleConfigurationChangeEvent(value);
+                this.handleConfigurationChangeEvent();
                 break;
             case providerReady:
                 this.handleProviderReadyEvent();
@@ -47,6 +47,7 @@ public class EventStreamObserver implements StreamObserver<EventStreamResponse> 
         this.callback.setEventStreamAlive(false);
         try {
             this.callback.restartEventStream();
+            this.callback.emitSuccessReconnectionEvents();
         } catch (Exception e) {
             log.error("restart event stream", e);
         }
@@ -60,23 +61,13 @@ public class EventStreamObserver implements StreamObserver<EventStreamResponse> 
         this.callback.setEventStreamAlive(false);
     }
 
-    private void handleConfigurationChangeEvent(EventStreamResponse value) {
+    private void handleConfigurationChangeEvent() {
+        this.callback.emitConfigurationChangeEvent();
         if (!this.cache.getEnabled()) {
             return;
         }
-
-        Map<String, Value> data = value.getData().getFieldsMap();
-        Value flagsValue = data.get(flagsKey);
-        if (flagsValue == null) {
-            this.cache.clear();
-            return;
-        }
-
-        Map<String, Value> flags = flagsValue.getStructValue().getFieldsMap();
-
-        for (String flagKey : flags.keySet()) {
-            this.cache.remove(flagKey);
-        }
+        // Always flush the cache when configuration_change event is received
+        this.cache.clear();
     }
 
     private void handleProviderReadyEvent() {
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
@@ -24,6 +24,8 @@ import dev.openfeature.sdk.ProviderState;
 import dev.openfeature.sdk.Value;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import io.grpc.ManagedChannel;
+import io.grpc.Status;
+import io.grpc.StatusRuntimeException;
 import io.grpc.netty.GrpcSslContexts;
 import io.grpc.netty.NettyChannelBuilder;
 import io.grpc.stub.StreamObserver;
@@ -36,7 +38,6 @@ import lombok.extern.slf4j.Slf4j;
 import javax.net.ssl.SSLException;
 import java.io.File;
 import java.util.HashMap;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.Executors;
@@ -118,17 +119,20 @@ public class FlagdProvider extends EventProvider implements FeatureProvider, Eve
     }
 
     @Override
-    public void initialize(EvaluationContext evaluationContext) throws Exception {
+    public void initialize(EvaluationContext evaluationContext) throws RuntimeException {
         try {
-            EventStreamRequest r = EventStreamRequest.getDefaultInstance();
-            Iterator<EventStreamResponse> eventStreamResponseIterator = this.serviceBlockingStub
+            // try a dummy request
+            this.serviceBlockingStub
                     .withWaitForReady()
                     .withDeadlineAfter(this.deadline, TimeUnit.MILLISECONDS)
-                    .eventStream(r);
-            if(!eventStreamResponseIterator.hasNext()) {
-                throw new Exception("cannot connect");
+                    .resolveBoolean(ResolveBooleanRequest.newBuilder().setFlagKey("ready?").build());
+        } catch (StatusRuntimeException e) {
+            // only return the exception if we don't meet the deadline
+            if (Status.DEADLINE_EXCEEDED.equals(e.getStatus())) {
+                throw e;
             }
         } finally {
+            // try in background to open the event stream
             Executors.newSingleThreadExecutor().submit(this::handleEvents);
         }
     }
@@ -137,12 +141,16 @@ public class FlagdProvider extends EventProvider implements FeatureProvider, Eve
     public void shutdown() {
         try {
             if (this.channel != null) {
-                this.channel.shutdown().awaitTermination(5, TimeUnit.SECONDS);
+                this.channel.shutdown();
+                this.channel.awaitTermination(5, TimeUnit.SECONDS);
             }
         } catch (InterruptedException e) {
             log.error("Error during shutdown {}", FLAGD_PROVIDER, e);
         } finally {
             this.cache.clear();
+            if (this.channel != null) {
+                this.channel.shutdownNow();
+            }
         }
     }
 
@@ -261,6 +269,7 @@ public class FlagdProvider extends EventProvider implements FeatureProvider, Eve
         return this;
     }
 
+    @Override
     public void setState(ProviderState state) {
         Lock l = this.lock.writeLock();
         try {
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
@@ -118,7 +118,9 @@ public class FlagdProvider extends EventProvider implements FeatureProvider, Eve
     @Override
     public void shutdown() {
         try {
-            this.channel.shutdown().awaitTermination(5, TimeUnit.SECONDS);
+            if(this.channel != null) {
+                this.channel.shutdown().awaitTermination(5, TimeUnit.SECONDS);
+            }
         } catch (InterruptedException e) {
             log.error("Error during shutdown {}", FLAGD_PROVIDER, e);
         } finally {
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
@@ -40,7 +40,6 @@ import java.io.File;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.Executors;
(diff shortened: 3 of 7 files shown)


providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/EventStreamObserver.java
```diff
@@ -60,23 +61,13 @@ public void onCompleted() {
...
-            return;
-        }
-
-        Map<String, Value> flags = flagsValue.getStructValue().getFieldsMap();
-
-        for (String flagKey : flags.keySet()) {
-            this.cache.remove(flagKey);
-        }
+        // Always flush the cache when configuration_change event is received
+        this.cache.clear();
```
thisthat: @toddbaert could you please double-check that I interpreted [the docs](https://github.com/open-feature/flagd/blob/main/docs/other_resources/creating_providers.md#provider-lifecycle-initialization-and-shutdown) correctly: 

> emit PROVIDER_CONFIGURATION_CHANGED event and flush cache when a configuration_change message is received on the streaming connection

toddbaert: No, I think because we know the flags changed, we don't need to flush them all - there's no reason to alter that behavior for this feature. I think I should update the doc to be more precise about this.

In the case of flagd, we know exactly the flags changed, so we can be more efficient by only busting those keys.

thisthat: Thanks! I have updated the PR and tests :) now it is ready for reviews

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
```diff
@@ -109,9 +106,32 @@ public FlagdProvider(final FlagdOptions options) {
...
+    }
+
+    @Override
+    public void initialize(EvaluationContext evaluationContext) {
+        this.serviceBlockingStub = this.serviceBlockingStub
+                .withWaitForReady()
+                .withDeadline(Deadline.after(30, TimeUnit.SECONDS));
+        this.serviceStub = this.serviceStub
+                .withWaitForReady()
+                .withDeadline(Deadline.after(30, TimeUnit.SECONDS));
```
toddbaert: I'm a bit confused with how this works in conjunction with `restartEventStream` below, which is called by the `EventStreamObserver` when there's an error. As `withWaitForReady` explains:

```This may dramatically increase the latency of the RPC, but avoids failing "unnecessarily." The default queues the RPC until an attempt to connect has completed, but fails RPCs without sending them if unable to connect.```

I'm concerned the 30s timeout _may_ be conflicting with our custom retry logic. If I'm understanding everything here correctly, it means that if we're having trouble connecting entirely, we won't see an error until 30s later. I think the way our retry works, it may be better to use the default method, since it will then retry using our backoff (because an error will be thrown).

This brings up another point... which we may want to address in another PR: we don't have integration tests that _fully_ test our retry logic, as far as I can tell. I think that's too much for this change though.

File: providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/EventStreamObserverTest.java [hunk 3]
```diff
@@ -0,0 +1,117 @@
...
+        }
+        @Test
+        public void Ready(){
+            Schema.EventStreamResponse resp = mock(Schema.EventStreamResponse.class);
+            when(resp.getType()).thenReturn("provider_ready");
+            stream.onNext(resp);
+            verify(callback, times(0)).emit(any(ProviderEvent.class), any(ProviderEventDetails.class));
+        }
+        @Test
+        public void Reconnections(){
```
toddbaert:
```suggestion
        }
        
        @Test
        public void Reconnections(){
```

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/EventStreamObserverTest.java
```diff
@@ -0,0 +1,117 @@
...
+            stream.onNext(resp);
+            // we notify that the configuration changed
+            verify(callback, times(1)).emitConfigurationChangeEvent();
+            verify(callback, times(1)).emitProviderConfigurationChanged(any(ProviderEventDetails.class));
+            verify(callback, times(1)).emit(eq(ProviderEvent.PROVIDER_CONFIGURATION_CHANGED), any(ProviderEventDetails.class));
+            // we flush the cache
+            verify(cache, atLeast(1)).clear();
+        }
+        @Test
+        public void Ready(){
```
toddbaert:
```suggestion
        }
        
        @Test
        public void Ready(){
```

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/EventStreamObserverTest.java
```diff
@@ -0,0 +1,117 @@
...
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
+
+import java.util.HashMap;
+
+import static org.mockito.Mockito.*;
+
+class EventStreamObserverTest {
+
```
toddbaert, line 17:
```suggestion
```

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/EventStreamObserverTest.java
```diff
@@ -0,0 +1,117 @@
...
+import dev.openfeature.sdk.ProviderEventDetails;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
+
+import java.util.HashMap;
+
+import static org.mockito.Mockito.*;
+
+class EventStreamObserverTest {
```
toddbaert, line 16: Thanks for adding these!

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
toddbaert: OK so I confirmed this locally. _Without_ the `withWaitForReady` and `withDeadline` stuff, we attempt our normal reconnect protocol immediately if flagd is unreachable. _With it_, we wait 30s until we do. I think we should remove those.

toddbaert:
We need to implement the `getState` method, which tells the SDK the providers state, so it knows to call `initialize`. As the spec dictates, providers without this are assumed to be in a `READY` state. You'll need to set the state to READY/ERROR where appropriate, and add a accessor method like as below:

```java
    @Override
    public ProviderState getState() {
        return this.state; // add this private member, and initialize it to `NOT_READY`.
    }
```

This is an easy one for people to miss if their provider needs to have `initialize` called. The reason it's easy to miss is that the `FeatureProvider` interface has a default implementation (always `READY`) to make everything backwards compatible. Our goal is to always prioritize ease-of-use, interoperability and experience for `application authors` over `provider authors`, and this is an example of a place where it's a bit easy for `provider authors` to misstep.

Edit: Oh, I didn't realize this, but it's possible to `@Override` a default method on an interface in an abstract class... so we can force authors of `EventProviders` (the abstract) to implement this method, even though the `FeatureProvider` has a default. That would technically be a breaking change in the SDK, but it would only impact events, which are experimental and only released very recently. It would add some guardrails here for `EventProvider` implementations, which are much more likely to need their `initialize()` called. Very interested in if you think this is a good change for the SDK.

toddbaert (changes requested):
Great job! You did the hard parts, but there's 3 small, necessary changes:

1: https://github.com/open-feature/java-sdk-contrib/pull/361#issuecomment-1645932475 
2: https://github.com/open-feature/java-sdk-contrib/pull/361/files#r1270813223
3: We need to require a 1.4.0 minimum SDK peer for this change, or the provider won't work as expected. You will need to override the required java SDK version in the `pom.xml` for this module. @Kavindu-Dodan did something similar in the OTel hooks.

I think `#3` makes this a breaking change, which is fine... but you'll need to add a `!` to the title.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
```diff
@@ -13,13 +13,9 @@
 import dev.openfeature.flagd.grpc.ServiceGrpc;
 import dev.openfeature.flagd.grpc.ServiceGrpc.ServiceBlockingStub;
 import dev.openfeature.flagd.grpc.ServiceGrpc.ServiceStub;
-import dev.openfeature.sdk.EvaluationContext;
-import dev.openfeature.sdk.FeatureProvider;
-import dev.openfeature.sdk.Metadata;
-import dev.openfeature.sdk.MutableStructure;
-import dev.openfeature.sdk.ProviderEvaluation;
-import dev.openfeature.sdk.Value;
+import dev.openfeature.sdk.*;
```
Kavindu-Dodan: minor - I would avoid wildcard imports and expand them

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
Kavindu-Dodan: Agree with the wait time argument

Also, for flagd provider, connection with backend can only be detected with the event stream connectivity. So the provider readiness can be blocked till we observe a notification on `eventStreamAliveSync` (ex: `this.eventStreamAliveSync.wait();`)

However, we might need to consider running event handling concurrently 🤔  right now it's done in the same thread and also have thread sleeps [^1].

[^1]: https://github.com/open-feature/java-sdk-contrib/blob/main/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java#L123

toddbaert: @Kavindu-Dodan might have a good point here.

thisthat: I added a deadline because I was following this documentation of gRPC: https://grpc.io/blog/deadlines/
The `withWaitForReady` delegates the gRPC library to do the heavy lifting in making sure the connection is established instead of having to do it by ourselves. 

The code needs some refactoring since the data- and control-flow are quite hard to follow. If we want to implement a blocking initialize method, I can only think about two quick solutions without many code changes:
- we set the deadline in an exp way to match the backoff that we have (removing the thread sleep)
- we create a thread that makes the connection and waits for the `this.eventStreamAliveSync.wait();` and we join on it 

I think the more clean implementation is the first option, but I am open to your input @toddbaert and @Kavindu-Dodan  :)

thisthat:
> We need to implement the `getState` method, which tells the SDK the providers state, so it knows to call `initialize`. As the spec dictates, providers without this are assumed to be in a `READY` state. You'll need to set the state to READY/ERROR where appropriate, and add a accessor method like as below:
> 
> ```java
>     @Override
>     public ProviderState getState() {
>         return this.state; // add this private member, and initialize it to `NOT_READY`.
>     }
> ```
> 
> This is an easy one for people to miss if their provider needs to have `initialize` called. The reason it's easy to miss is that the `FeatureProvider` interface has a default implementation (always `READY`) to make everything backwards compatible. Our goal is to always prioritize ease-of-use, interoperability and experience for `application authors` over `provider authors`, and this is an example of a place where it's a bit easy for `provider authors` to misstep.
> 
> Edit: Oh, I didn't realize this, but it's possible to `@Override` a default method on an interface in an abstract class... so we can force authors of `EventProviders` (the abstract) to implement this method, even though the `FeatureProvider` has a default. That would technically be a breaking change in the SDK, but it would only impact events, which are experimental and only released very recently. It would add some guardrails here for `EventProvider` implementations, which are much more likely to need their `initialize()` called. Very interested in if you think this is a good change for the SDK.

oh thanks for pointing that out! I totally missed that point 😅 
I am in favor of forcing provider implementers to think about when they are ready to perform evaluations 👍

thisthat:
@toddbaert @Kavindu-Dodan I've been address your suggestions, thanks for the review. I also created https://github.com/open-feature/java-sdk-contrib/issues/365 for collecting some thoughts about refactoring the Flagd Provider a bit. Please, let me know your thoughts on that :)

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
toddbaert: Just to be clear, the SDK itself runs the `initialize` in a thread it manages. This was intentional to prevent initialization from ever blocking the whole application. That said, it still might be worth not blocking this thread. Not sure what you guys think about that.

> - we set the deadline in an exp way to match the backoff that we have (removing the thread sleep)
> - we create a thread that makes the connection and waits for the this.eventStreamAliveSync.wait(); and we join on it

Personally I'm completely fine with either of these, as long as we conform to the requirements (exponential back-off with a max-delay and max attempts). It _may_ be possible to implement the reconnect stuff more generically using the second model, but I'm not sure that is too important because I don't expect we will ever have a an additional non-gRPC implementation.

providers/flagd/pom.xml
```diff
@@ -26,6 +26,16 @@
     </developers>
 
     <dependencies>
+
+        <!-- override parent definition -->
+        <dependency>
+            <groupId>dev.openfeature</groupId>
+            <artifactId>sdk</artifactId>
+            <version>[1.4,2.0)</version>
+            <scope>provided</scope>
```
toddbaert, line 35: ty!

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
```diff
@@ -205,12 +250,12 @@ public FlagdProvider setDeadline(long deadlineMs) {
     }
 
     @Override
-    public void setEventStreamAlive(boolean alive) {
+    public void setState(ProviderState state) {
```
toddbaert, line 273: I don't think we need this method. In fact, I'm not sure why we aren't see an error in the CI with the `@Override` annotation, since this doesn't actually override anything.

The state is something the provider should manage internally. It's only to signal to the SDK that the provider is ready (the interface only defines a getter). Basically you can just set a private backing field to READY/ERROR after init, and with disconnects, etc.

EDIT: Oh, it's because we have `setState` defined in the other interface.

thisthat: The `public` modifier comes from implementing an interface method of the [EventStreamCallback](https://github.com/open-feature/java-sdk-contrib/blob/main/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/EventStreamCallback.java#L7) interface. 
I planned to work on this details as part of #365 (second point in the list) since it requires quite some rework on the code

toddbaert:
@thisthat you're reconnect logic seems to be working as expected. I see a reconnect back-off in the console in my little test app if flagd isn't running. The bad news is, even if flagd IS running, I see the same back-off. This is new since your last changes - the ones I tested on Friday didn't have this issue. Something in the recent commits seems to have broken the streaming altogether. :grimacing:

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
toddbaert: Ya, I see that now. My local IDE somehow hand't caught up and was showing a compilation error here.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
```diff
@@ -392,7 +437,9 @@ private static ManagedChannel nettyChannel(final FlagdOptions options) {
 
     private void handleEvents() {
         StreamObserver<EventStreamResponse> responseObserver = new EventStreamObserver(this.cache, this);
-        this.serviceStub.eventStream(EventStreamRequest.getDefaultInstance(), responseObserver);
+        this.serviceStub.withWaitForReady()
+                .withDeadline(Deadline.after(this.eventStreamRetryBackoff, TimeUnit.MILLISECONDS))
+                .eventStream(EventStreamRequest.getDefaultInstance(), responseObserver);
```
toddbaert: I think what you're written here means:

`start the event stream, and if I _ever_ (even after the initial message) don't get a message for eventStreamRetryBackoff seconds, throw`. Which is the cause of [this](https://github.com/open-feature/java-sdk-contrib/pull/361#issuecomment-1647943587).

thisthat:
> @thisthat you're reconnect logic seems to be working as expected. I see a reconnect back-off in the console in my little test app if flagd isn't running. The bad news is, even if flagd IS running, I see the same back-off. This is new since your last changes - the ones I tested on Friday didn't have this issue. Something in the recent commits seems to have broken the streaming altogether. 😬

I think we should add a smoke test with [testcontainers](https://testcontainers.com/) to run the latest FlagD image to exactly pick this issues up :)

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
toddbaert: https://github.com/grpc/grpc-java/issues/5498#issuecomment-476299936

toddbaert:
> > @thisthat you're reconnect logic seems to be working as expected. I see a reconnect back-off in the console in my little test app if flagd isn't running. The bad news is, even if flagd IS running, I see the same back-off. This is new since your last changes - the ones I tested on Friday didn't have this issue. Something in the recent commits seems to have broken the streaming altogether. grimacing
> 
> I think we should add a smoke test with [testcontainers](https://testcontainers.com/) to run the latest FlagD image to exactly pick this issues up :)

I 100% agree, and in fact we exactly have this in the SDK repo (a fully integration test using the SDK, gherkin suite, and flagd). We soon can migrate that one to use the in-memory provider, in line with @Kavindu-Dodan 's [proposal](https://github.com/open-feature/spec/pull/199). Then we should move the flagd version here for an e2e test, IMO. I've been thinking about creating an issue for this. I can do so if you agree.

thisthat:
> > > @thisthat you're reconnect logic seems to be working as expected. I see a reconnect back-off in the console in my little test app if flagd isn't running. The bad news is, even if flagd IS running, I see the same back-off. This is new since your last changes - the ones I tested on Friday didn't have this issue. Something in the recent commits seems to have broken the streaming altogether. grimacing
> > 
> > 
> > I think we should add a smoke test with [testcontainers](https://testcontainers.com/) to run the latest FlagD image to exactly pick this issues up :)
> 
> I 100% agree, and in fact we exactly have this in the SDK repo (a fully integration test using the SDK, gherkin suite, and flagd). We soon can migrate that one to use the in-memory provider, in line with @Kavindu-Dodan 's [proposal](https://github.com/open-feature/spec/pull/199). Then we should move the flagd version here for an e2e test, IMO. I've been thinking about creating an issue for this. I can do so if you agree.

Absolutely 👍

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
thisthat: I went through the effort of finally learning a bit more about the inner details of gRPC.
gRPC completely handles in an async fashion the communication channel. It has a retry logic built-in with exp backoff ([docs](https://github.com/grpc/grpc/blob/master/doc/connection-backoff.md)). There is a method on the channel class that could be used to see its status ([docs](https://grpc.github.io/grpc-java/javadoc/io/grpc/ManagedChannel.html#getState-boolean-)), but it is marked as experimental and they say that's not fully representative and become outdated soon. Therefore, I suggest running a dummy flag evaluation and checking the returned error to determine the connection state. 
As part of #365, I suggest removing this custom logic and offloading to the netty channel the reconnecting logic.

toddbaert: I'm aware of the gRPC reconnection logic, and it's definitely helpful. The problem is, the event stream and the underlying connection are not the same thing. When the connection is lost, the stream needs to be reconnected, and in all implementations I'm aware of that needs to be done manually, which is why in the end, we need a stream (not a connection) reconnect protocol. See [here](https://stackoverflow.com/questions/66353603/correct-way-to-perform-a-reconnect-with-grpc-client) for a supporting stack overflow.

I could be wrong, I would love to find out that the event stream can be automatically re-established without writing such code!

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
```diff
@@ -109,22 +115,80 @@ public FlagdProvider(final FlagdOptions options) {
...
+                    .withDeadlineAfter(this.deadline, TimeUnit.MILLISECONDS)
+                    .resolveBoolean(ResolveBooleanRequest.newBuilder().setFlagKey("ready?").build());
+        } catch (StatusRuntimeException e) {
+            // only return the exception if we don't meet the deadline
+            if (Status.DEADLINE_EXCEEDED.equals(e.getStatus())) {
+                throw e;
+            }
+        } finally {
+            // try in background to open the event stream
+            this.handleEvents();
```
toddbaert, line 135: I suppose this works, but I think that the main purpose of `initialize` is to connect the stream. I imagine there could be other reasons the stream could fail besides the connection? Perhaps due to some network/fw restrictions? I think the safest way to test the initialize would be to actually open the stream. Is doing that in a blocking way not possible?

I could be misunderstanding something here. Let me know!

thisthat: gRPC builds on top of channels, if there is no connection with the channel, the RPC won't be scheduled/executed. Since both stubs use the same channel, I used the blocking one to perform an evaluation which ensures that we can connect to FlagD. I also preferred using the blocking stub because this is the critical path for communication with FlagD since FF evaluations run over it. The nonblocking stub is used to remove keys from the cache. Do you prefer to check that both of them are set up correctly? 

> Perhaps due to some network/fw restrictions?

do you mean if there is a level 7 firewall? I think running the evaluation would cover this issue since we transmit data over the wire.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
thisthat: You're totally correct! Thanks a lot @toddbaert for the explanation! Here's also another thread about the topic https://github.com/grpc/grpc-java/issues/8177#issuecomment-842461759

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
toddbaert: > do you mean if there is a level 7 firewall? I think running the evaluation would cover this issue since we transmit data over the wire.

You're probably right. TBH I don't feel too strongly about this. As long as we init properly I think that's fine.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/EventStreamObserver.java
```diff
@@ -57,14 +60,14 @@ public void onCompleted() {
         if (this.cache.getEnabled()) {
             this.cache.clear();
         }
-        this.callback.setEventStreamAlive(false);
+        this.callback.setState(ProviderState.ERROR);
```
toddbaert, line 63: I have a question about this, as to whether or not we should consider this an error state, since I believe this state can only be arrived at by graceful shutdown. I have the same question in the JS impl. I think you could argue that this this should be `NOT_READY` or `ERROR`, or neither. Interested in your opinion.

toddbaert (approved):
Approved with a question (one that also impacts other implementations): https://github.com/open-feature/java-sdk-contrib/pull/361#discussion_r1273808701

beeme1mr:
Why is this a breaking change?

thisthat:
> Why is this a breaking change?

Because it bumps the minimum supported version of the Java SDK (see https://github.com/open-feature/java-sdk-contrib/pull/361#pullrequestreview-1541468446)

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/EventStreamObserver.java
toddbaert: Keeping this as is after discussion with @thisthat . More work may be required to perfect this, but it's beyond the scope of this PR and not a blocker.

toddbaert:
I'm going to do one last smoke test on this (using it in the SDK e2e) and then merge.

toddbaert:
Looks great in the e2e tests. Thanks for all your help.

PR #369, opened 2023-07-25
feat: ISSUE-658 go-feature-flag sdk - add cache

liran2000:
## This PR
Implementing [issue 658](https://github.com/thomaspoignant/go-feature-flag/issues/658): (feature) Implement open-feature provider cache for the Java provider.

- add provider cache
- publish data collector events

### Related Issues
 [issue 658](https://github.com/thomaspoignant/go-feature-flag/issues/658)

### How to test
- Run unit tests - tested.
- Can test via go-feature-flag instance, while observing logs, and checking collect data API requests - not tested.


providers/go-feature-flag/pom.xml
```diff
@@ -50,5 +50,24 @@
...
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+            <version>2.0.7</version>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.logging.log4j</groupId>
+            <artifactId>log4j-slf4j2-impl</artifactId>
+            <version>2.20.0</version>
```
thomaspoignant, line 69: I see that log4j is used only for the tests, so I will suggest to scope of this dependency to tests only.

```suggestion
            <version>2.20.0</version>
            <scope>test</test>
```

providers/go-feature-flag/src/main/java/dev/openfeature/contrib/providers/gofeatureflag/bean/BeanUtils.java
```diff
@@ -0,0 +1,20 @@
...
+/**
+ * Bean utils.
+ */
+public class BeanUtils {
+
+    private static ObjectMapper objectMapper = new ObjectMapper();
+
+    private BeanUtils() {
+
+    }
```
thomaspoignant: Can you remove this and use the Lombok annotation instead?

You can add this 
```java
@NoArgsConstructor(access = AccessLevel.PRIVATE)
public class BeanUtils {
// ...
```

It will make the code clearer.

providers/go-feature-flag/src/main/java/dev/openfeature/contrib/providers/gofeatureflag/bean/BeanUtils.java
```diff
@@ -0,0 +1,20 @@
...
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+/**
+ * Bean utils.
+ */
+public class BeanUtils {
+
+    private static ObjectMapper objectMapper = new ObjectMapper();
```
thomaspoignant:
```suggestion
    private static final ObjectMapper objectMapper = new ObjectMapper();
```

We are not re-affecting the object, so `objectMapper` may be `final`.

providers/go-feature-flag/src/main/java/dev/openfeature/contrib/providers/gofeatureflag/GoFeatureFlagProvider.java
```diff
@@ -34,28 +41,48 @@
...
  * GoFeatureFlagProvider is the JAVA provider implementation for the feature flag solution GO Feature Flag.
  */
+@Slf4j
 public class GoFeatureFlagProvider implements FeatureProvider {
     private static final String NAME = "GO Feature Flag Provider";
     private static final ObjectMapper requestMapper = new ObjectMapper();
     private static final ObjectMapper responseMapper = new ObjectMapper()
             .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
+
+    public static final int DEFAULT_CACHE_TTL_MINUTES = 10;
```
thomaspoignant: To stay consistent with other providers can we have this in milliseconds rather than minutes? [_(see in the node provider)_](https://github.com/open-feature/js-sdk-contrib/blob/f7c54179b97c7b54685e35a16705a5242919ccf2/libs/providers/go-feature-flag/src/lib/model.ts#L51)

And can we use 1 minute as a default value.

providers/go-feature-flag/src/main/java/dev/openfeature/contrib/providers/gofeatureflag/GoFeatureFlagProviderOptions.java
```diff
@@ -46,6 +44,26 @@ public class GoFeatureFlagProviderOptions {
...
+     */
+    private Boolean enableCache;
+
+    /**
+     * (optional) interval time we publish statistics collection data to the proxy.
+     * The parameter is used only if the cache is enabled, otherwise the collection of the data is done directly
+     * when calling the evaluation API.
+     * default: 1 minute
+     */
+    private Long flushIntervalMinues;
```
thomaspoignant: To stay consistent with other providers I will suggest to use milliseconds here rather than minutes _[(see node provider)](https://github.com/open-feature/js-sdk-contrib/blob/f7c54179b97c7b54685e35a16705a5242919ccf2/libs/providers/go-feature-flag/src/lib/model.ts#L75)_.

providers/go-feature-flag/src/main/java/dev/openfeature/contrib/providers/gofeatureflag/GoFeatureFlagProvider.java
```diff
@@ -34,28 +41,48 @@
...
+import java.time.Duration;
 import java.time.Instant;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Consumer;
 import java.util.stream.Collectors;
 
+import static java.net.HttpURLConnection.*;
+import static org.apache.commons.lang3.BooleanUtils.isFalse;
```
thomaspoignant: I see that you are using `apache commons` but it is not imported in the `pom.xml`.

I guess it is available from a sub-dependency. 
I would recommend not using it here because it can disappear at some point if the sub-dependency stop using it.

We can wether import it in the `pom.xml` or what I would prefer, copy/paste the method `isFalse` somewhere in our code since this function is so simple that we should not import all `apache commons` just for this.

Bellow the function from apache commons sources:
```java
public static boolean isFalse(Boolean bool) {
    return Boolean.FALSE.equals(bool);
}
```

providers/go-feature-flag/src/main/java/dev/openfeature/contrib/providers/gofeatureflag/GoFeatureFlagProvider.java
```diff
@@ -34,28 +41,48 @@
...
 public class GoFeatureFlagProvider implements FeatureProvider {
     private static final String NAME = "GO Feature Flag Provider";
     private static final ObjectMapper requestMapper = new ObjectMapper();
     private static final ObjectMapper responseMapper = new ObjectMapper()
             .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
+
+    public static final int DEFAULT_CACHE_TTL_MINUTES = 10;
+    public static final int DEFAULT_CACHE_CONCURRENCY_LEVEL = 1;
+    public static final int DEFAULT_CACHE_INITIAL_CAPACITY = 100;
+    public static final int DEFAULT_CACHE_MAXIMUM_SIZE = 1000;
```
thomaspoignant:
```suggestion
    public static final int DEFAULT_CACHE_MAXIMUM_SIZE = 10000;
```

For consistency with the [`node` provider](https://github.com/open-feature/js-sdk-contrib/blob/f7c54179b97c7b54685e35a16705a5242919ccf2/libs/providers/go-feature-flag/src/lib/model.ts#L63).

providers/go-feature-flag/src/main/java/dev/openfeature/contrib/providers/gofeatureflag/GoFeatureFlagProvider.java
```diff
@@ -322,4 +424,45 @@ private Structure mapToStructure(Map<String, Object> map) {
...
+                requestMapper.writeValueAsBytes(events),
+                MediaType.get("application/json; charset=utf-8")));
+
+        if (this.apiKey != null && !"".equals(this.apiKey)) {
+            reqBuilder.addHeader("Authorization", "Bearer " + this.apiKey);
+        }
+
+        try (Response response = this.httpClient.newCall(reqBuilder.build()).execute()) {
+            if (response.code() == HTTP_UNAUTHORIZED) {
+                throw new GeneralError("Unauthorized");
```
thomaspoignant: I am not sure we should throw in case the data collection is not working.
What I would prefer is that we are keeping the data in memory and we retry. A log can still be welcome for this.

What do you think about it?

providers/go-feature-flag/src/main/java/dev/openfeature/contrib/providers/gofeatureflag/GoFeatureFlagProvider.java
```diff
@@ -129,52 +183,100 @@ public List<Hook> getProviderHooks() {
...
+
+    private <T> ProviderEvaluation getProviderEvaluationWithCheckCache(
+            String key, T defaultValue, Class<?> expectedType, GoFeatureFlagUser user) {
+        ProviderEvaluation res = null;
+        try {
+            String cacheKey = buildCacheKey(key, BeanUtils.buildKey(user));
+            res = cache.getIfPresent(cacheKey);
+            if (res == null) {
+                res = resolveEvaluationGoFeatureFlagProxy(key, defaultValue, user, expectedType);
+                cache.put(cacheKey, res);
```
thomaspoignant: Not all the flags are cacheable.
We added a new field in the GO Feature Flag API response to know if we can or not cache the evaluation result.

What we should do here is add a new property in the file `providers/go-feature-flag/src/main/java/dev/openfeature/contrib/providers/gofeatureflag/bean/GoFeatureFlagResponse.java` called `cacheable` that will tell us if we can cache the flag or not.

And here we need to check the field before being able to add it to the cache.
[You can see similar behavior in the GO provider](https://github.com/open-feature/go-sdk-contrib/blob/18d01eff5038c086ca6f2b54956513882e9e189d/providers/go-feature-flag/pkg/provider.go#L479).

thomaspoignant (changes requested):
@liran2000 thanks a lot for your contribution, this is really awesome, and I can't wait to have the cache in the JAVA provider 🙌.

I've done a first review pass and found some points that can be improved, let me know if you need any further information about them.

providers/go-feature-flag/pom.xml
liran2000: Used logging also at non-test GoFeatureFlagProvider, with Slf4j annotation and _log_ field.

providers/go-feature-flag/src/main/java/dev/openfeature/contrib/providers/gofeatureflag/GoFeatureFlagProvider.java
liran2000: Revisited the flow, such that if it fails, it will log the error, and will retry on the next intervals.
For not overflowing the in-memory collection on long failures, limit the collection size to a quite large size limit.

Data collection events: the data is used only for statistics at the server side ?

providers/go-feature-flag/pom.xml
thomaspoignant: Yes but your logger comes from slf4j not from log4j (except in the test).
In the provider itself the [dependency to `org.slf4j`](https://github.com/liran2000/java-sdk-contrib/blob/a4892214f3ca435593fb3bc4db38d0b63fcd62cf/providers/go-feature-flag/pom.xml#L60C9-L64) is enough, no need to specify an implementation.

providers/go-feature-flag/src/main/java/dev/openfeature/contrib/providers/gofeatureflag/GoFeatureFlagProvider.java
thomaspoignant: Yes, the data are here only for statistics on the server side nothing else.

providers/go-feature-flag/src/main/java/dev/openfeature/contrib/providers/gofeatureflag/GoFeatureFlagProvider.java
```diff
@@ -113,6 +154,28 @@ private void initializeProvider(GoFeatureFlagProviderOptions options) throws Inv
...
+            long flushIntervalMs = options.getFlushIntervalMs() == null
+                ? DEFAULT_FLUSH_INTERVAL_MS : options.getFlushIntervalMs();
+            int maxPendingEvents = options.getMaxPendingEvents() == null
+                ? DEFAULT_MAX_PENDING_EVENTS : options.getMaxPendingEvents();
+            Consumer<List<Event>> publisher = events -> publishEvents(events);
+            eventsPublisher = new EventsPublisher(publisher, flushIntervalMs, maxPendingEvents);
+        }
+    }
+
+    private Cache buildDefaultCache() {
```
thomaspoignant:
```suggestion
    private Cache<String, ProviderEvaluation<?>> buildDefaultCache() {
```

providers/go-feature-flag/src/main/java/dev/openfeature/contrib/providers/gofeatureflag/GoFeatureFlagProviderOptions.java
```diff
@@ -46,6 +44,33 @@ public class GoFeatureFlagProviderOptions {
...
      */
-    @Getter
     private String apiKey;
+
+    /**
+     *  (optional) If cache custom configuration is wanted, you should provide
+     *  a cache builder.
+     *  Default: null
+     */
+    private CacheBuilder cacheBuilder;
```
thomaspoignant:
```suggestion
    private CacheBuilder<String, ProviderEvaluation<?>> cacheBuilder;
```

providers/go-feature-flag/src/main/java/dev/openfeature/contrib/providers/gofeatureflag/GoFeatureFlagProvider.java
```diff
@@ -113,6 +154,28 @@ private void initializeProvider(GoFeatureFlagProviderOptions options) throws Inv
...
+                this.cache = options.getCacheBuilder().build();
+            } else {
+                this.cache = buildDefaultCache();
+            }
+            long flushIntervalMs = options.getFlushIntervalMs() == null
+                ? DEFAULT_FLUSH_INTERVAL_MS : options.getFlushIntervalMs();
+            int maxPendingEvents = options.getMaxPendingEvents() == null
+                ? DEFAULT_MAX_PENDING_EVENTS : options.getMaxPendingEvents();
+            Consumer<List<Event>> publisher = events -> publishEvents(events);
+            eventsPublisher = new EventsPublisher(publisher, flushIntervalMs, maxPendingEvents);
```
thomaspoignant:
```suggestion
            eventsPublisher = new EventsPublisher<>(publisher, flushIntervalMs, maxPendingEvents);
```

providers/go-feature-flag/src/main/java/dev/openfeature/contrib/providers/gofeatureflag/GoFeatureFlagProvider.java
```diff
@@ -113,6 +154,28 @@ private void initializeProvider(GoFeatureFlagProviderOptions options) throws Inv
...
+            if (options.getCacheBuilder() != null) {
+                this.cache = options.getCacheBuilder().build();
+            } else {
+                this.cache = buildDefaultCache();
+            }
+            long flushIntervalMs = options.getFlushIntervalMs() == null
+                ? DEFAULT_FLUSH_INTERVAL_MS : options.getFlushIntervalMs();
+            int maxPendingEvents = options.getMaxPendingEvents() == null
+                ? DEFAULT_MAX_PENDING_EVENTS : options.getMaxPendingEvents();
+            Consumer<List<Event>> publisher = events -> publishEvents(events);
```
thomaspoignant:
```suggestion
            Consumer<List<Event>> publisher = this::publishEvents;
```

providers/go-feature-flag/src/main/java/dev/openfeature/contrib/providers/gofeatureflag/GoFeatureFlagProvider.java
```diff
@@ -129,52 +192,106 @@ public List<Hook> getProviderHooks() {
...
+    }
+
+    @Override
+    public void initialize(EvaluationContext evaluationContext) throws Exception {
+        FeatureProvider.super.initialize(evaluationContext);
+    }
+
+    private <T> ProviderEvaluation<T> getEvaluation(
+            String key, T defaultValue, EvaluationContext evaluationContext, Class<?> expectedType) {
+        ProviderEvaluation res = null;
```
thomaspoignant:
```suggestion
        ProviderEvaluation<T> res = null;
```

providers/go-feature-flag/src/main/java/dev/openfeature/contrib/providers/gofeatureflag/GoFeatureFlagProvider.java
```diff
@@ -129,52 +192,106 @@ public List<Hook> getProviderHooks() {
...
+                .creationDate(System.currentTimeMillis())
+                .build()
+            );
+        }
+        return res;
+    }
+
+    private <T> ProviderEvaluation getProviderEvaluationWithCheckCache(
+            String key, T defaultValue, Class<?> expectedType, GoFeatureFlagUser user) {
+        ProviderEvaluation res = null;
```
thomaspoignant:
```suggestion
        ProviderEvaluation<?> res = null;
```

providers/go-feature-flag/src/main/java/dev/openfeature/contrib/providers/gofeatureflag/GoFeatureFlagProvider.java
```diff
@@ -129,52 +192,106 @@ public List<Hook> getProviderHooks() {
...
+                    key, defaultValue, user, expectedType);
+                if (Boolean.TRUE.equals(proxyRes.getCachable())) {
+                    cache.put(cacheKey, proxyRes.getProviderEvaluation());
+                }
+                res = proxyRes.getProviderEvaluation();
+            } else {
+                res.setReason(CACHED_REASON);
+            }
+        } catch (JsonProcessingException e) {
+            log.error("Error building key for user", user);
```
thomaspoignant:
```suggestion
            log.error("Error building key for user: {}", user);
```

thomaspoignant (changes requested):
Hey, I've spotted some minor things here and there to fix some warnings.

The main thing we should tackle is that we should probably override the method [`initialize`](https://github.com/open-feature/java-sdk/blob/fc034c328483735e7dbd973d2897af6f2ce8ef4a/src/main/java/dev/openfeature/sdk/FeatureProvider.java#L39) to initialize the collection in the background and to use fully the SDK capabilities.

For this what we may do is move the code from `initializeProvider` to `initialize` and remove the call from the constructor.

If we move the code there we will also have to implement the `getStatus()` method and have the provider status lifecycle ready in the provider.
I've done something similar in the node provider here if you want to have a look: https://github.com/open-feature/js-sdk-contrib/pull/493

If you prefer, I'll be happy to help you with that myself and to collaborate on this pull request, but this up to you.

providers/go-feature-flag/src/main/java/dev/openfeature/contrib/providers/gofeatureflag/GoFeatureFlagProvider.java
```diff
@@ -247,53 +264,49 @@ private <T> ProviderEvaluation<T> getEvaluation(
...
+                    .defaultValue(defaultValue)
+                    .variation(cacheResponse.getVariant())
+                    .value(cacheResponse.getValue())
+                    .userKey(user.getKey())
+                    .creationDate(System.currentTimeMillis())
+                    .build()
+            );
+            return (ProviderEvaluation<T>) cacheResponse;
+        } catch (Exception e) {
+            if (!(e instanceof NotPresentInCache)) {
```
liran2000: it makes normal positive flow of first evaluation to get to exception - feels less natural and efficient, don't you think ?

providers/go-feature-flag/src/main/java/dev/openfeature/contrib/providers/gofeatureflag/GoFeatureFlagProvider.java
```diff
@@ -247,53 +264,49 @@ private <T> ProviderEvaluation<T> getEvaluation(
...
-                res = proxyRes.getProviderEvaluation();
-            } else {
-                res.setReason(CACHED_REASON);
+            cacheKey = buildCacheKey(key, BeanUtils.buildKey(user));
+            ProviderEvaluation<?> cacheResponse = cache.getIfPresent(cacheKey);
+            if (cacheResponse == null) {
+                throw new NotPresentInCache(cacheKey);
+            }
+            cacheResponse.setReason(CACHED_REASON);
+            eventsPublisher.add(Event.builder()
```
liran2000: why skip the event when value is evaluated from proxy ?

providers/go-feature-flag/src/test/java/dev/openfeature/contrib/providers/gofeatureflag/GoFeatureFlagProviderTest.java
```diff
@@ -514,7 +514,7 @@ void should_publish_events() {
         assertEquals(0, g.getEventsPublisher().publish(), "first attempt expected to fail");
 
         // simulate publish on next interval
-        assertEquals(3, g.getEventsPublisher().publish(), "expected to publish all events after retry");
+        assertEquals(2, g.getEventsPublisher().publish(), "expected to publish all events after retry");
```
liran2000, line 517: why ? there are 3 evaluations before this, why expect 2 and not 3 events ?
See also comment above for skipping the event.

liran2000 (commented):
@thomaspoignant thanks for reviewing, please see my comments on the [commit](https://github.com/open-feature/java-sdk-contrib/pull/369/commits/a9d27b5211a93c3f01857db7a2beb2695cd0a089), as it affects the flow.

providers/go-feature-flag/src/test/java/dev/openfeature/contrib/providers/gofeatureflag/GoFeatureFlagProviderTest.java
thomaspoignant: The first evaluation is done remotely by the relay proxy, it will be collected directly in the server, so we should not take care of this evaluation.
We are sending to the endpoint `/data/collector` only the evaluation done by the cache.

See this diagram on how it works:
```mermaid
sequenceDiagram
    provider ->> relay proxy: evalute flag X
    relay proxy -->> data exporter: call the exporter
    relay proxy -->> provider: flag evaluation details
    provider ->> provider: evalute flag X (use cache)
    provider ->> provider: evalute flag X (use cache)
    provider -->> relay proxy: call /data/collector endpoint
    relay proxy -->> data exporter: call the exporter
    relay proxy -->> provider: ack
```

providers/go-feature-flag/src/main/java/dev/openfeature/contrib/providers/gofeatureflag/GoFeatureFlagProvider.java
thomaspoignant: see https://github.com/open-feature/java-sdk-contrib/pull/369#discussion_r1284440475

providers/go-feature-flag/src/main/java/dev/openfeature/contrib/providers/gofeatureflag/GoFeatureFlagProvider.java
thomaspoignant: I don't really see the problem here, an `Exception` does not mean an error every time.

providers/go-feature-flag/src/test/java/dev/openfeature/contrib/providers/gofeatureflag/GoFeatureFlagProviderTest.java
liran2000: Cool. It was not clear from the issue, can you possibly update the [requirements](https://gofeatureflag.org/specification/provider-cache) documentation for it?

providers/go-feature-flag/src/main/java/dev/openfeature/contrib/providers/gofeatureflag/GoFeatureFlagProvider.java
liran2000: In my opinion, previous flow of if condition was more readable: 
```
if value not in cache
    fetch proxy
```
[Here](https://softwareengineering.stackexchange.com/questions/189222/are-exceptions-as-control-flow-considered-a-serious-antipattern-if-so-why) is some reference.

providers/go-feature-flag/src/test/java/dev/openfeature/contrib/providers/gofeatureflag/GoFeatureFlagProviderTest.java
thomaspoignant: Yes I will update it !

thomaspoignant: Done, waiting for the website to be deployed.

providers/go-feature-flag/src/main/java/dev/openfeature/contrib/providers/gofeatureflag/GoFeatureFlagProvider.java
thomaspoignant: yes that’s a good point, what I would like is to have all the flow in a single method and avoid as much as possible to duplicate code.

But I can have a look after the week end if you want

thomaspoignant (approved):

PR #378, opened 2023-07-28
chore(flagd): refactoring codebase into multiple classes

### LINKED ISSUE #365: [FlagD] Refactor package structure and offload some responsibilities from FlagdProvider
### Problems

- Everything is bundled together into a single namespace which makes it hard to understand the responsibilities of each class. 
- The [current main commit](https://github.com/open-feature/java-sdk-contrib/blob/7b12a97dcf59bef42e8a53cc9d3b4523d106123f/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java#L208) allows for external entities to change the internal state of the communication of Flagd Provider. IMHO we should remove this from the API surface. 
- Unit and Integration tests are hard to write
- Review opportunities to add `log.debug` statements

### Suggestion

Refactor into modules as proposed: 
image

thisthat:
## This PR

- refactor the codebase into more packages and classes

### Related Issues

Fixes #365


providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/strategy/ResolveFactory.java
```diff
@@ -0,0 +1,12 @@
+package dev.openfeature.contrib.providers.flagd.strategy;
```
Kavindu-Dodan, line 1: Minor - this can also fit into a `telemetry` package

providers/flagd/src/test/java/smoketest/FFEvaluationTest.java
```diff
@@ -0,0 +1,52 @@
...
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+@Testcontainers
+class FFEvaluationTest {
+
+    private final String ffKey = "myFlag";
+
+    private final String flagsPath = getClass().getClassLoader().getResource("flags.json").getPath();
+    @Container
+    private final GenericContainer<?> flagd = new GenericContainer<>("ghcr.io/open-feature/flagd:v0.6.1")
```
Kavindu-Dodan: minor - we can extract flagd version to a const and use renovate to bump the version

providers/flagd/src/test/java/smoketest/FFEvaluationTest.java
```diff
@@ -0,0 +1,52 @@
+package smoketest;
```
Kavindu-Dodan: +1 for tests

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/GrpcConnector.java
```diff
@@ -0,0 +1,173 @@
...
+    private final FlagdCache cache;
+    private final Consumer<ProviderState> setState;
+
+    public GrpcConnector(final FlagdOptions options, final FlagdCache cache, Consumer<ProviderState> setState) {
+        this.channel = nettyChannel(options);
+        this.serviceStub = ServiceGrpc.newStub(channel);
+        this.serviceBlockingStub = ServiceGrpc.newBlockingStub(channel);
+        this.maxEventStreamRetries = options.getMaxEventStreamRetries();
+        this.startEventStreamRetryBackoff = options.getRetryBackoffMs();
+        this.eventStreamRetryBackoff = this.startEventStreamRetryBackoff;
```
Kavindu-Dodan:
```suggestion
        this.startEventStreamRetryBackoff = options.getRetryBackoffMs();
        this.eventStreamRetryBackoff = options.getRetryBackoffMs();
```

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/GrpcConnector.java
```diff
@@ -0,0 +1,173 @@
...
+            // only return the exception if we don't meet the deadline
+            if (Status.DEADLINE_EXCEEDED.equals(e.getStatus())) {
+                throw e;
+            }
+        } finally {
+            // try in background to open the event stream
+            this.startEventStream();
+        }
+        /**
+         * Evaluate if the following makes sense
```
Kavindu-Dodan: I went through this code and IMO the initial dummy evaluation is a good enough strategy

I don't think we should fail in the init method. Another alternative is to use a timer and set error state if event stream connectivity fails

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/GrpcConnector.java
```diff
@@ -0,0 +1,173 @@
...
+    private final ManagedChannel channel;
+    private final int maxEventStreamRetries;
+
+    private int eventStreamAttempt = 1;
+    private int eventStreamRetryBackoff;
+    private final int startEventStreamRetryBackoff;
+    private final long deadline;
+
+    private final FlagdCache cache;
+    private final Consumer<ProviderState> setState;
```
Kavindu-Dodan:
```suggestion
    private final Consumer<ProviderState> stateConsumer;
```

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/GrpcConnector.java
```diff
@@ -0,0 +1,173 @@
...
+        this.serviceBlockingStub = ServiceGrpc.newBlockingStub(channel);
+        this.maxEventStreamRetries = options.getMaxEventStreamRetries();
+        this.startEventStreamRetryBackoff = options.getRetryBackoffMs();
+        this.eventStreamRetryBackoff = this.startEventStreamRetryBackoff;
+        this.deadline = options.getDeadline();
+        this.cache = cache;
+        this.setState = setState;
+    }
+
+    public void initialize(EvaluationContext evaluationContext) throws RuntimeException {
```
Kavindu-Dodan: can omit `EvaluationContext` param through this refactoring

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/GrpcConnector.java
```diff
@@ -0,0 +1,173 @@
...
+            return;
+        }
+        this.eventStreamRetryBackoff = 2 * this.eventStreamRetryBackoff;
+        try {
+            Thread.sleep(this.eventStreamRetryBackoff);
+        } catch (InterruptedException e) {
+            log.debug("Failed to sleep while restarting gRPC Event Stream");
+        }
+        this.startEventStream();
+    }
```
Kavindu-Dodan: I would love to see little bit of cleanup on this logic (if possible). 

For example, 

- see if we can reuse `EventStreamObserver`
- renaming methods to make this logic easy to understand 
- Make sure thread usage is clear in loigic 

The main confusion here is `restartEventStream` calls  `startEventStream` which internally refers `restartEventStream` (circular dep)

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/strategy/ResolveFactory.java
thisthat: I didn't want to change the behavior with this PR, but we should always provide an instrumented version IMHO.
If in the App code, the SDK is not registered, then it will be a no-op implementation.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/GrpcConnector.java
thisthat: I wanted to keep it simpler and don't move around too much critical things in a single PR. 
The circular dep will always be there to an extent since Streams don't try to reconnect by themselves. 
I planned to move the reconnect trigger from the Event Stream into the state handling of the provider. This way, the `EventStreamObserver` is responsible for only notifying that an error in the stream occurred and the provider handles the re-connection. However, I'd suggest addressing this logic in a follow-up given the complexity and here, improve the naming, WDYT @Kavindu-Dodan?

providers/flagd/src/test/java/smoketest/FFEvaluationTest.java
toddbaert: In the long term I created [this](https://github.com/open-feature/java-sdk-contrib/issues/380).

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/GrpcConnector.java
toddbaert: I think a there's already a lot going on here so I'm in favor of creating another issue for this. We could even label it with `help-wanted`, etc to encourage new contributors to help with it.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/GrpcConnector.java
toddbaert: > Another alternative is to fail the init method on purpose and let the status recover once we are connected

I could be misunderstanding, but if you throw in `init` it will cause the SDK to send an error event, which we don't want if things are functioning normally.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagResolution.java
```diff
@@ -0,0 +1,229 @@
...
+
+import static dev.openfeature.contrib.providers.flagd.Config.CACHED_REASON;
+import static dev.openfeature.contrib.providers.flagd.Config.CONTEXT_FIELD;
+import static dev.openfeature.contrib.providers.flagd.Config.FLAG_KEY_FIELD;
+import static dev.openfeature.contrib.providers.flagd.Config.REASON_FIELD;
+import static dev.openfeature.contrib.providers.flagd.Config.STATIC_REASON;
+import static dev.openfeature.contrib.providers.flagd.Config.VALUE_FIELD;
+import static dev.openfeature.contrib.providers.flagd.Config.VARIANT_FIELD;
+
+final class FlagResolution {
```
toddbaert: A bit of a nit, but since this has protobuf stuff in it, I think it should potentially be under the `grpc/` dir.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
```diff
@@ -1,83 +1,43 @@
...
     private final ReadWriteLock lock = new ReentrantReadWriteLock();
 
-    private ServiceBlockingStub serviceBlockingStub;
-    private ServiceStub serviceStub;
-    private ManagedChannel channel;
-    private final int maxEventStreamRetries;
-    private final Object eventStreamAliveSync;
     private final FlagdCache cache;
     private final ResolveStrategy strategy;
+    private final GrpcConnector grpc;
```
toddbaert, line 40: Another nit, not a requirement, but in JS we created a generic abstraction (`service`), which `grpc-service` implements. It's a nice way to keep anything explicitly `grpc` related conceptually out of the provider: https://github.com/open-feature/js-sdk-contrib/tree/main/libs/providers/flagd/src/lib/service

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/EventStreamObserver.java
```diff
@@ -1,28 +1,32 @@
...
 @Slf4j
 public class EventStreamObserver implements StreamObserver<EventStreamResponse> {
-    private final EventStreamCallback callback;
+    private final Consumer<ProviderState> setState;
+    private final Runnable reconnectEventStream;
     private final FlagdCache cache;
 
     private static final String configurationChange = "configuration_change";
     private static final String providerReady = "provider_ready";
     static final String flagsKey = "flags";
```
toddbaert: Can you use SCREAMING_SNAKE_CASE for these? (not something you wrote, I know).

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/GrpcConnector.java
```diff
@@ -0,0 +1,173 @@
...
+        this.setState = setState;
+    }
+
+    public void initialize(EvaluationContext evaluationContext) throws RuntimeException {
+        try {
+            // try a dummy request
+            this.serviceBlockingStub
+                    .withWaitForReady()
+                    .withDeadlineAfter(this.deadline, TimeUnit.MILLISECONDS)
+                    .resolveBoolean(Schema.ResolveBooleanRequest.newBuilder().setFlagKey("ready?").build());
```
toddbaert, line 75: Another way of doing this entire thing would be a countdown latch. It might go something like this:

- add functionality to the `EventStreamObserver` implementation which decrements a countdown latch on initial connection (on_ready)
- start non-blocking event stream observer (`this.serviceStub.eventStream...`), passing the above `EventStreamObserver` instance
- await the above mentioned countdown latch in `initialize`

This results in a implementation that blocks the initialize until the first read message is received, doesn't create extra connection or do unnecessary evaluations, etc.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/GrpcConnector.java
Kavindu-Dodan: +1 for a follow-up pr

providers/flagd/src/test/java/smoketest/FFEvaluationTest.java
toddbaert: @thisthat I created https://github.com/open-feature/java-sdk-contrib/pull/387. This should help you and I hope shouldn't cause any conflicts for you with this PR.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/GrpcConnector.java
thisthat: That's a nice idea @toddbaert 👍

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/EventStreamObserver.java
```diff
@@ -46,25 +55,20 @@ public void onError(Throwable t) {
         if (this.cache.getEnabled()) {
             this.cache.clear();
```
thisthat, line 58: I believe we should not bust the cache in case of error and instead use the latest known values until the stream is reconnected. This line basically prevents to do ANY flag evaluation if we cannot connect to flagd. IMHO we should still be able to resolve flags according to what we have in the cache instead of failing the gRPC resolution and use the default value.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/GrpcConnector.java
thisthat: Added to https://github.com/open-feature/java-sdk-contrib/issues/390

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
thisthat: Added to https://github.com/open-feature/java-sdk-contrib/issues/390

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/GrpcConnector.java
thisthat: Added to https://github.com/open-feature/java-sdk-contrib/issues/390

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/strategy/ResolveFactory.java
thisthat: Added to https://github.com/open-feature/java-sdk-contrib/issues/390

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/EventStreamObserver.java
thisthat: Added to https://github.com/open-feature/java-sdk-contrib/issues/390

providers/flagd/src/test/java/smoketest/FFEvaluationTest.java
thisthat: I removed the smoketest package and the dependencies on testcontainers :)

pom.xml
```diff
@@ -47,7 +47,14 @@
         <junit.jupiter.version>5.10.0</junit.jupiter.version>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
-
+<<<<<<< HEAD
```
Kavindu-Dodan: lint - merge conlfict

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/EventStreamObserver.java
Kavindu-Dodan: The idea here is that a connection lost results in stale flags. And stale flags could result in inconsistent behavior. So I think we should keep this. 

However, another option is to make this a configurable option

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/FlagResolution.java
```diff
@@ -0,0 +1,242 @@
...
+        return message.getDescriptorForType().findFieldByName(name);
+    }
+
+    /**
+     * A converter lambda.
+     */
+    @FunctionalInterface
+    public interface Convert<OutT extends Object, InT extends Object> {
+        OutT convert(InT value);
+    }
```
Kavindu-Dodan: To support Java way of things, this is best extracted to its own interface class

Kavindu-Dodan (approved):
Approving with stated comments and fixes

pom.xml
thisthat: I've forgot to ctrl+s 🤦‍♂️

toddbaert: Happens to the best of us. :wink:

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/EventStreamObserver.java
toddbaert: The problem is we don't cache all values, we only cache static flags (flags without rules).

This puts us in a situation where we really can't evaluate all flags. If we were evaluating flags in memory, we could just have "STALE" flags, but with the RPC design, we simply can't evaluate dynamic flags at all.

toddbaert (approved):

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/EventStreamObserver.java
thisthat: This screams for a configurable option as @Kavindu-Dodan suggests 🤔 
I'll try to explain my thoughts: since values are static, I want to "remotely rewrite" the default value in the code. Hence, saving the last known values would allow me to evaluate things as the last point in time I had a connection to flagd. 
For dynamic flags, failing and defaulting to the default value is fine since I cannot pass the ctx to flagd to evaluate the flags according to the rules.

toddbaert:
Thanks @thisthat ! Looks much better IMO.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/EventStreamObserver.java
toddbaert: I can see how some people might be interested in this option. It's worth discussing, but it can't be done in isolation. We can talk about it (perhaps in a flagd discussion?) and add it to the doc so it can be implemented/scheduled in all flagd RPC providers.

Issue #380, opened 2023-07-31
[flagd] Create e2e test for flagd provider


toddbaert:
This should be similar to what's in the java-sdk now.

Definition of done:
- run e2e tests based on test-harness [ghekin](https://github.com/open-feature/test-harness/blob/main/features/evaluation.feature) suite and flagd test image. A lot of the code can be taken from the [java SDK impl](https://github.com/open-feature/java-sdk/blob/main/src/test/java/dev/openfeature/sdk/e2e/StepDefinitions.java).
- [run in CI](https://github.com/open-feature/java-sdk/blob/main/.github/workflows/pullrequest.yml#L11-L15)

PR #386, opened 2023-08-02
chore: fix failing tests

Kavindu-Dodan:
## This PR

Noticed flagd class loading issues in tests after updating `netty-transport-native-epoll` to version `4.1.96.Final` [^1]

This PR downgrade the version to `4.1.95.Final` and update some of the dependencies.

Addendum 

The root cause was a change introduced with 4.1.96.Final release. Opened an issue to discuss with netty community on a possible fix - https://github.com/netty/netty/issues/13523 

[^1]: https://github.com/open-feature/java-sdk-contrib/pull/372

### DIFF
--- a/providers/flagd/pom.xml
+++ b/providers/flagd/pom.xml
@@ -41,14 +41,14 @@
         <dependency>
             <groupId>io.grpc</groupId>
             <artifactId>grpc-netty</artifactId>
-            <version>1.57.0</version>
+            <version>1.57.1</version>
         </dependency>
 
         <dependency>
             <!-- we only support unix sockets on linux, via epoll native lib -->
             <groupId>io.netty</groupId>
             <artifactId>netty-transport-native-epoll</artifactId>
-            <version>4.1.96.Final</version>
+            <version>4.1.95.Final</version>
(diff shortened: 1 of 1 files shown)


providers/flagd/pom.xml
```diff
@@ -41,22 +41,22 @@
...
-            <version>1.57.0</version>
+            <version>1.57.1</version>
         </dependency>
 
         <dependency>
             <!-- we only support unix sockets on linux, via epoll native lib -->
             <groupId>io.netty</groupId>
             <artifactId>netty-transport-native-epoll</artifactId>
-            <version>4.1.96.Final</version>
+            <version>4.1.95.Final</version>
```
toddbaert, line 51: if you think `.96` is bugged, let's be careful updating it again.

I will prioritize this [next](https://github.com/open-feature/java-sdk-contrib/issues/380), so we can have some e2e tests here.

toddbaert (approved):

providers/flagd/pom.xml
Kavindu-Dodan: Root cause was introduction of a static block validation. I opened an issue for netty - https://github.com/netty/netty/issues/13523

PR #389, opened 2023-08-02
feat: flagd flag evaluation metadata

Kavindu-Dodan:
## This PR

Related to flagd change at https://github.com/open-feature/flagd/pull/761

Handles evaluation metadata from flagd and fills required OF SDK field

### DIFF
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/Config.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/Config.java
@@ -30,7 +30,7 @@ public final class Config {
     public static final String VARIANT_FIELD = "variant";
     public static final String VALUE_FIELD = "value";
     public static final String REASON_FIELD = "reason";
-    static final String METADATA_FIELD = "metadata";
+    public static final String METADATA_FIELD = "metadata";
 
     public static final String LRU_CACHE = "lru";
     static final String DEFAULT_CACHE = LRU_CACHE;
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/Config.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/Config.java
@@ -30,6 +30,7 @@ public final class Config {
     public static final String VARIANT_FIELD = "variant";
     public static final String VALUE_FIELD = "value";
     public static final String REASON_FIELD = "reason";
+    static final String METADATA_FIELD = "metadata";
 
     public static final String LRU_CACHE = "lru";
     static final String DEFAULT_CACHE = LRU_CACHE;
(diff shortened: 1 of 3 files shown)


providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
```diff
@@ -377,7 +380,7 @@ private Value convertList(com.google.protobuf.ListValue protobuf) {
     /**
      * Convert openfeature value to protobuf value.
      */
-    private com.google.protobuf.Value convertPrimitive(Value value) {
+    private static com.google.protobuf.Value convertPrimitive(Value value) {
```
thisthat: I'd push `com.google.protobuf` into the import section like it is for Struct

thisthat (approved):

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
```diff
@@ -527,6 +535,30 @@ private static FieldDescriptor getFieldDescriptor(Message message, String name)
...
+
+        for (Map.Entry<String, com.google.protobuf.Value> entry : struct.getFieldsMap().entrySet()) {
+            if (entry.getValue().hasStringValue()) {
+                builder.addString(entry.getKey(), entry.getValue().getStringValue());
+            } else if (entry.getValue().hasBoolValue()) {
+                builder.addBoolean(entry.getKey(), entry.getValue().getBoolValue());
+            } else if (entry.getValue().hasNumberValue()) {
+                builder.addDouble(entry.getKey(), entry.getValue().getNumberValue());
+            }
+        }
```
toddbaert: I like streams but that's personal preference (and probably slower?)

toddbaert (approved):

beeme1mr (approved):

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
Kavindu-Dodan: IMO streaming here is not needed as we simply wants to iterate over the entries. However, if there's some filtering, then streams would help.

I don't think there's any performance impact for smaller collection from either approach

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
Kavindu-Dodan: Unfortunately no, we have conflict between `com.google.protobuf.Value` & `dev.openfeature.sdk.Value` . So the protobuf value need fully qualified name

Kavindu-Dodan:
Merging as all conversations and suggestions are applied :)

PR #391, opened 2023-08-04
feat: hook builder options for dimension extraction

### LINKED ISSUE #381: [otel hooks] add functional means of adding custom attributes, custom attributes on traces
Currently it's possible to map flag metadata values to custom OTel attributes (see [here](https://github.com/open-feature/java-sdk-contrib/tree/main/hooks/open-telemetry#recording-custom-dimensions)). However this approach has limitations, as it can only map values 1:1.

We should implement something similar to the [JS implementation](https://github.com/open-feature/js-sdk-contrib/tree/main/libs/hooks/open-telemetry#custom-metric-attributes), which allows consumers to provide a function that takes the flag metadata and returns OTel attributes.

We should add the same functionality to the Tracing hook.

Kavindu-Dodan:
## This PR

Fixes https://github.com/open-feature/java-sdk-contrib/issues/381 

Adds callback option to extract custom dimensions from flag evaluation metadata. 

This is similar to what we do in JS SDK - https://github.com/open-feature/js-sdk-contrib/pull/520 

```Java
MetricHookOptions hookOptions = MetricHookOptions.builder()
        //Configure a callback which accepts ImmutableMetadata
        .attributeSetter(metadata -> Attributes.builder()
            .put("boolean", metadata.getBoolean("boolean"))
            .put("integer", metadata.getInteger("integer"))
            .build())
        .build();
```

This change is backward compatible and does not change the current beavhior.

### DIFF
--- a/hooks/open-telemetry/README.md
+++ b/hooks/open-telemetry/README.md
@@ -111,7 +111,7 @@ OpenFeatureAPI api = OpenFeatureAPI.getInstance();
 api.addHooks(new MetricsHook(openTelemetry, customDimensions));
 ```
 
-Alternatively, you can wrtie your own extraction logic against [flag evaluation metadata](https://github.com/open-feature/spec/blob/main/specification/types.md#flag-metadata) by providing a callback to `TracesHookOptions.dimensionExtractor`.
+Alternatively, you can wrtie your own extraction logic against [flag evaluation metadata](https://github.com/open-feature/spec/blob/main/specification/types.md#flag-metadata) by providing a callback to `TracesHookOptions.attributeSetter`.
 
 ```java
 final OpenTelemetry openTelemetry = ... // OpenTelemetry API instance
--- a/hooks/open-telemetry/README.md
+++ b/hooks/open-telemetry/README.md
@@ -33,7 +33,22 @@ The hook implementation is attached to `after`and `error` [hook stages](https://
 Both successful and failed flag evaluations will add a span event named `feature_flag` with evaluation details such as flag key, provider name and variant.
 Failed evaluations can be allowed to set span status to `ERROR`. You can configure this behavior through`TracesHookOptions`.
 
-Consider the following code example for usage,
+Further, you can write your own logic to extract custom dimensions from [flag evaluation metadata](https://github.com/open-feature/spec/blob/main/specification/types.md#flag-metadata) by setting a callback to `TracesHookOptions.dimensionExtractor`.
+Extracted dimensions will be added to successful falg evaluation spans.
+
+```java
+ TracesHookOptions options = TracesHookOptions.builder()
+             // configure a callback
+            .dimensionExtractor(metadata -> Attributes.builder()
+                    .put("boolean", metadata.getBoolean("boolean"))
+                    .put("integer", metadata.getInteger("integer"))
+                    .build()
+            ).build();
+
+    TracesHook tracesHook = new TracesHook(options);
+```
+
+Consider the following code example for a complete usage,
 
 ```java
 final Tracer tracer=... // derive Tracer from OpenTelemetry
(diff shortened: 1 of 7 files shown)


hooks/open-telemetry/src/main/java/dev/openfeature/contrib/hooks/otel/MetricHookOptions.java
```diff
@@ -0,0 +1,33 @@
...
+     * Custom handler to derive {@link Attributes} from flag evaluation metadata represented with
+     * {@link ImmutableMetadata}.
+     */
+    private Function<ImmutableMetadata, Attributes> attributeSetter;
+
+    /**
+     * List of {@link DimensionDescription} to be extracted from flag evaluation metadata.
+     */
+    @Builder.Default
+    private final List<DimensionDescription> setDimensions = Collections.emptyList();
```
thisthat, line 32: why do we need two different methods?

Kavindu-Dodan: This is to be backward compatible with the current release [^1] where we allow a list of Dimension descriptions to be extracted from metadata. 

With the proposed change of this PR, we are trying to provide call-back support aligning with OTel hook of JS SDK

[^1]: https://github.com/open-feature/java-sdk-contrib/blob/dev.openfeature.contrib.hooks.otel-v2.0.0/hooks/open-telemetry/src/main/java/dev/openfeature/contrib/hooks/otel/MetricsHook.java#L53-L59

thisthat: I see 👍  thanks for the explanation. Then, I'd suggest adding the `@Deprecated` annotation here to the old method 🤔

hooks/open-telemetry/README.md
```diff
@@ -95,3 +110,19 @@ OpenFeatureAPI api = OpenFeatureAPI.getInstance();
 // Register MetricsHook with custom dimensions
 api.addHooks(new MetricsHook(openTelemetry, customDimensions));
 ```
+
+Alternatively, you can wrtie your own extraction logic against [flag evaluation metadata](https://github.com/open-feature/spec/blob/main/specification/types.md#flag-metadata) by providing a callback to `TracesHookOptions.dimensionExtractor`.
```
thisthat: nit: here we say dimensionExtractor but the code down uses attributeSetter

hooks/open-telemetry/src/main/java/dev/openfeature/contrib/hooks/otel/MetricHookOptions.java
Kavindu-Dodan: hmm, I am not sure whether we should deprecate this option. IMO it can continue to coexist alongside the callback option. The `DimensionDescription` approach is minimalistic but restrictive where as the callback option gives flexibility to end user.

hooks/open-telemetry/README.md
Kavindu-Dodan: good catch, updated with the latest commit

hooks/open-telemetry/src/main/java/dev/openfeature/contrib/hooks/otel/MetricHookOptions.java
thiyagu06: Hmmm. The old method also converts the `DimensionDescription` to attributesBuilder from flagmeta data. I would prefer to expose one single method in SDK to create atrributes and not to confuse the user. So if an user want to set an atrributes, they will use lambda approach only.

Kavindu-Dodan: @thiyagu06 yes that's correct. However, they can be considered as two approaches (one for shallow extraction and the other for customized extractions).

I am open to any opinion here. If we want to remove, then this should be done in two steps - first deprecate and then remove to avoid usage interruptions.

thiyagu06: Hmm. I agree. In that case, i would prefer to  keep the lambda based approach which provides both option of extracting attributes. Also, its one less concept for (`DimensionDescription` ) SDK users to understand the attribute builder for otel metrics.

Lets hear other thoughts as well.

Kavindu-Dodan: @thisthat @toddbaert @beeme1mr appreciate your feedback on this :)

toddbaert: I think my slight preference is to keep just the one option, but I tend to agree with @Kavindu-Dodan that marking for deprecation is a better first step.

Honestly I don't feel very strongly either way on this one.

toddbaert (approved):

hooks/open-telemetry/src/main/java/dev/openfeature/contrib/hooks/otel/MetricHookOptions.java
Kavindu-Dodan: @toddbaert thank you for the feedback

I added deprecation notice for the dimension description construct - https://github.com/open-feature/java-sdk-contrib/pull/391/commits/396ab41d5da3d036600d28606d6dd7c12376bfa5

We still have this through `MetricHookOptions` but eventually we can deprecate that too.

thiyagu06 (approved):

Issue #390, opened 2023-08-03
[flagd] follow-up refactoring


thisthat:
```[tasklist]
### Tasks
- [x] Improve retry-logic for gRPC stream ([ctx](https://github.com/open-feature/java-sdk-contrib/pull/378#discussion_r1279580804))
- [x] Abstract gPRC into a generic service ([ctx](https://github.com/open-feature/java-sdk-contrib/pull/378#discussion_r1281078271))
- [x] Improve `initialize(EvaluationContext)` strategy ([ctx](https://github.com/open-feature/java-sdk-contrib/pull/378#discussion_r1281102577))
- [ ] Keep the latest known cache status in case of error ([ctx](https://github.com/open-feature/java-sdk-contrib/pull/378/files#r1282810265))
- [x] Use always OTel interceptor ([ctx](https://github.com/open-feature/java-sdk-contrib/pull/378#discussion_r1280580832))
```

toddbaert:
See my comment [here](https://github.com/open-feature/java-sdk-contrib/pull/378#discussion_r1283687965), I'm not sure we can really do much about the 4th point.

The rest looks good.

Kavindu-Dodan:
I will work on this :)

Kavindu-Dodan:
@thisthat except for [this](https://github.com/open-feature/java-sdk-contrib/pull/378/files#r1282810265), we have completed all other tasks. And this last one is not necessary for the current operation of flagd. Further, in-process provider proposed through https://github.com/open-feature/java-sdk-contrib/issues/411 will provider more options to use flagd. 

I am voting to close this as complete

thisthat:
I think also the last point (always use OTel) is not yet done. 
In the code ([link](https://github.com/open-feature/java-sdk-contrib/blob/main/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/grpc/strategy/ResolveFactory.java#L15C21-L15C37)), we instrument the gRPC streams only if we pass an SDK instance instead of acting as a library and get the default tracer all the time. OTel docs: https://opentelemetry.io/docs/concepts/instrumentation/libraries/

thiyagu06:
[...]

PR #398, opened 2023-08-15
chore: refactor flagd event handling 

Kavindu-Dodan:
## This PR

Completes some of the tasks mentioned at #390 

Introduce clean contracts for flagd evaluator.

- Introduce abstraction layer between flagd OF implementation & GRPC connection handling - `Resolver` interface
- Remove test focused constructors and rely on reflection to inject mocks
- Retry logic is now handled by a dedicated thread & utilize wait notify mechanism. Dummy calls were removed
- Shutdown is propagated to the retry logic handler (thread) to allow proper resource celanup 

Best reviewed either checking out the code OR visiting branch directly on my fork - https://github.com/Kavindu-Dodan/java-sdk-contrib/tree/feat/refactor-flagd/providers/flagd

### DIFF
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/EventStreamCallback.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/EventStreamCallback.java
@@ -1,9 +1,0 @@
-package dev.openfeature.contrib.providers.flagd.grpc;
-
-/**
- * Defines behaviour required of event stream callbacks.
- */
-public interface EventStreamCallback {
-
-
-}
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/EventStreamObserver.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/EventStreamObserver.java
@@ -18,7 +18,7 @@ import java.util.function.Consumer;
 @SuppressFBWarnings(justification = "cache needs to be read and write by multiple objects")
 class EventStreamObserver implements StreamObserver<EventStreamResponse> {
     private final Consumer<ProviderState> stateConsumer;
-    private final Runnable reconnectEventStream;
+    private final Object sync;
     private final Cache cache;
 
     private static final String CONFIGURATION_CHANGE = "configuration_change";
@@ -32,10 +32,10 @@ class EventStreamObserver implements StreamObserver<EventStreamResponse> {
      * @param stateConsumer        lambda to call for setting the state
      * @param reconnectEventStream callback for trying to recreate the stream
      */
-    EventStreamObserver(Cache cache, Consumer<ProviderState> stateConsumer, Runnable reconnectEventStream) {
+    EventStreamObserver(Object sync ,Cache cache, Consumer<ProviderState> stateConsumer) {
+        this.sync = sync;
         this.cache = cache;
         this.stateConsumer = stateConsumer;
-        this.reconnectEventStream = reconnectEventStream;
     }
 
     @Override
@@ -59,7 +59,9 @@ class EventStreamObserver implements StreamObserver<EventStreamResponse> {
             this.cache.clear();
         }
         this.stateConsumer.accept(ProviderState.ERROR);
-        this.reconnectEventStream.run();
+        synchronized (this.sync) {
+            this.sync.notify();
+        }
     }
 
     @Override
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/EventStreamObserver.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/EventStreamObserver.java
@@ -32,7 +32,7 @@ class EventStreamObserver implements StreamObserver<EventStreamResponse> {
      * @param stateConsumer        lambda to call for setting the state
      * @param reconnectEventStream callback for trying to recreate the stream
      */
-    EventStreamObserver(Object sync ,Cache cache, Consumer<ProviderState> stateConsumer) {
+    EventStreamObserver(Object sync, Cache cache, Consumer<ProviderState> stateConsumer) {
         this.sync = sync;
         this.cache = cache;
         this.stateConsumer = stateConsumer;
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/EventStreamObserver.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/EventStreamObserver.java
@@ -28,9 +28,9 @@ class EventStreamObserver implements StreamObserver<EventStreamResponse> {
     /**
      * Create a gRPC stream that get notified about flag changes.
      *
+     * @param sync                 synchronization object from caller
      * @param cache                cache to update
      * @param stateConsumer        lambda to call for setting the state
-     * @param reconnectEventStream callback for trying to recreate the stream
      */
     EventStreamObserver(Object sync, Cache cache, Consumer<ProviderState> stateConsumer) {
         this.sync = sync;
@@ -59,9 +59,9 @@ class EventStreamObserver implements StreamObserver<EventStreamResponse> {
             this.cache.clear();
         }
         this.stateConsumer.accept(ProviderState.ERROR);
-        synchronized (this.sync) {
-            this.sync.notify();
-        }
+
+        // handle last call of this stream
+        handleEndOfStream();
     }
 
     @Override
@@ -70,6 +70,9 @@ class EventStreamObserver implements StreamObserver<EventStreamResponse> {
             this.cache.clear();
         }
         this.stateConsumer.accept(ProviderState.ERROR);
+
+        // handle last call of this stream
+        handleEndOfStream();
     }
 
     private void handleConfigurationChangeEvent(EventStreamResponse value) {
@@ -94,6 +97,12 @@ class EventStreamObserver implements StreamObserver<EventStreamResponse> {
         this.stateConsumer.accept(ProviderState.READY);
         if (this.cache.getEnabled()) {
             this.cache.clear();
+        }
+    }
+
+    private void handleEndOfStream() {
+        synchronized (this.sync) {
+            this.sync.notify();
         }
     }
 }
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/EventStreamObserver.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/EventStreamObserver.java
@@ -16,7 +16,7 @@ import java.util.function.Consumer;
  */
 @Slf4j
 @SuppressFBWarnings(justification = "cache needs to be read and write by multiple objects")
-public class EventStreamObserver implements StreamObserver<EventStreamResponse> {
+class EventStreamObserver implements StreamObserver<EventStreamResponse> {
     private final Consumer<ProviderState> stateConsumer;
     private final Runnable reconnectEventStream;
     private final Cache cache;
@@ -27,11 +27,12 @@ public class EventStreamObserver implements StreamObserver<EventStreamResponse> 
 
     /**
      * Create a gRPC stream that get notified about flag changes.
-     * @param cache cache to update
-     * @param stateConsumer lambda to call for setting the state
+     *
+     * @param cache                cache to update
+     * @param stateConsumer        lambda to call for setting the state
      * @param reconnectEventStream callback for trying to recreate the stream
      */
-    public EventStreamObserver(Cache cache, Consumer<ProviderState> stateConsumer, Runnable reconnectEventStream) {
+    EventStreamObserver(Cache cache, Consumer<ProviderState> stateConsumer, Runnable reconnectEventStream) {
         this.cache = cache;
         this.stateConsumer = stateConsumer;
         this.reconnectEventStream = reconnectEventStream;
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/GrpcConnector.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/GrpcConnector.java
@@ -60,8 +60,7 @@ public class GrpcConnector {
         this.cache = cache;
         this.stateConsumer = stateConsumer;
     }
-
-
+    
     /**
      * Initialize the gRPC stream.
      * @throws RuntimeException if the connection cannot be established.
@@ -103,19 +102,19 @@ public class GrpcConnector {
     }
 
     /**
-     * Resets internal state for retrying establishing the stream.
-     */
-    public void resetRetryConnection() {
-        this.eventStreamAttempt = 1;
-        this.eventStreamRetryBackoff = this.startEventStreamRetryBackoff;
-    }
-
-    /**
      * Provide the object that can be used to resolve Feature Flag values.
      * @return a {@link ServiceGrpc.ServiceBlockingStub} for running FF resolution.
      */
     public ServiceGrpc.ServiceBlockingStub getResolver() {
         return serviceBlockingStub.withDeadlineAfter(this.deadline, TimeUnit.MILLISECONDS);
+    }
+
+    /**
+     * Resets internal state for retrying establishing the stream.
+     */
+    private void resetRetryConnection() {
+        this.eventStreamAttempt = 1;
+        this.eventStreamRetryBackoff = this.startEventStreamRetryBackoff;
     }
 
     private void startEventStream() {
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/GrpcConnector.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/GrpcConnector.java
@@ -115,11 +115,7 @@ public class GrpcConnector {
         while (this.eventStreamAttempt <= this.maxEventStreamRetries) {
             final StreamObserver<Schema.EventStreamResponse> responseObserver =
                     new EventStreamObserver(sync, this.cache, this::grpcStateConsumer);
-
-            // Event listener with wait for ready but fail with timeouts
-            this.serviceStub
-                    .withWaitForReady()
-                    .eventStream(Schema.EventStreamRequest.getDefaultInstance(), responseObserver);
+            this.serviceStub.eventStream(Schema.EventStreamRequest.getDefaultInstance(), responseObserver);
 
             try {
                 synchronized (sync) {
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/GrpcConnector.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/GrpcConnector.java
@@ -18,6 +18,7 @@ import lombok.extern.slf4j.Slf4j;
 
 import javax.net.ssl.SSLException;
 import java.io.File;
+import java.util.Random;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Consumer;
@@ -30,6 +31,8 @@ import java.util.function.Consumer;
 public class GrpcConnector {
     private final Object sync = new Object();
     private final AtomicBoolean connected = new AtomicBoolean(false);
+    private final Random random = new Random();
+    private final int maxJitter = 100;
 
     private final ServiceGrpc.ServiceBlockingStub serviceBlockingStub;
     private final ServiceGrpc.ServiceStub serviceStub;
@@ -133,7 +136,8 @@ public class GrpcConnector {
             }
 
             this.eventStreamAttempt++;
-            this.eventStreamRetryBackoff = 2 * this.eventStreamRetryBackoff;
+            // backoff with a jitter
+            this.eventStreamRetryBackoff = 2 * this.eventStreamRetryBackoff + random.nextInt(maxJitter);
 
             try {
                 Thread.sleep(this.eventStreamRetryBackoff);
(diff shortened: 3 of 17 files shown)


providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/EventStreamCallback.java
```diff
@@ -1,9 +0,0 @@
-package dev.openfeature.contrib.providers.flagd.grpc;
```
Kavindu-Dodan, line 1: Removed as this was unused

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/EventStreamObserver.java
```diff
@@ -58,7 +59,9 @@ public void onError(Throwable t) {
             this.cache.clear();
         }
         this.stateConsumer.accept(ProviderState.ERROR);
-        this.reconnectEventStream.run();
+        synchronized (this.sync) {
```
thisthat: I think we should also add this to the `onCompleted()` method

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/GrpcConnector.java
```diff
@@ -102,43 +105,54 @@ public void shutdown() throws Exception {
...
-    private void restartEventStream() {
-        this.eventStreamAttempt++;
-        if (this.eventStreamAttempt > this.maxEventStreamRetries) {
-            log.error("failed to connect to event stream, exhausted retries");
-            this.stateConsumer.accept(ProviderState.ERROR);
-            return;
+            try {
+                Thread.sleep(this.eventStreamRetryBackoff);
+            } catch (InterruptedException e) {
+                log.debug("failed to block while restarting gRPC Event Stream");
```
thisthat: so we don't confuse this with the wait call reading the logs :) 
```suggestion
                log.debug("failed to sleep while restarting gRPC Event Stream");
```

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/GrpcConnector.java
```diff
@@ -102,43 +105,54 @@ public void shutdown() throws Exception {
...
+                return;
```
thisthat, line 147: I have mixed feelings about what to do here. I think it might be worth having a `continue` so we retry until we exhaust the retries

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/GrpcConnector.java
```diff
@@ -80,12 +82,13 @@ public void initialize() throws RuntimeException {
             }
         } finally {
             // try in background to open the event stream
-            this.startEventStream();
+            new Thread(this::observeEventStream).start();
```
liran2000: Perhaps can use a single threaded (or other) executor, and then shut it down gracefully at shutdown() method ?

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/GrpcConnector.java
```diff
@@ -102,43 +105,54 @@ public void shutdown() throws Exception {
...
-    }
+    private void observeEventStream() {
+        while (this.eventStreamAttempt <= this.maxEventStreamRetries) {
+            StreamObserver<Schema.EventStreamResponse> responseObserver =
+                    new EventStreamObserver(sync, this.cache, this::grpcStateConsumer);
+            this.serviceStub.eventStream(Schema.EventStreamRequest.getDefaultInstance(), responseObserver);
+
+            try {
+                synchronized (sync) {
+                    sync.wait();
```
liran2000, line 129: For common flow of app shutdown, I guess it can be one of the cases:
1. a graceful shutdown of the thread, it is commonly expected to be interrupted if it is still running - in this case it will print a warning to log if the wait() is interrupted, but is expected as normal positive flow of app shutdown if I understand correctly ?
2. notify() on the object - if I understand correctly, it will proceed to an error ?

There are optional [alternatives](https://stackoverflow.com/questions/4912165/whats-the-best-alternative-to-wait-notify-for-low-level-synchronization#answer-4916691) to wait/notify to consider.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/GrpcConnector.java
```diff
@@ -61,9 +63,9 @@ public GrpcConnector(final FlagdOptions options, final Cache cache, Consumer<Pro
         this.stateConsumer = stateConsumer;
     }
 
-
     /**
      * Initialize the gRPC stream.
+     *
      * @throws RuntimeException if the connection cannot be established.
      */
     public void initialize() throws RuntimeException {
```
toddbaert: I'd still like to see something like this done, if possible. I think that was the meaning of the 3rd checkbox [here](https://github.com/open-feature/java-sdk-contrib/issues/390).

thisthat:
Please, change the wording in the PR description from `Fixes` to `Part of` or something similar. Otherwise the original issue will be closed when not all the points are addressed

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/EventStreamObserver.java
liran2000: If this can be called from multiple threads, and the other thread which wait on this object is during wait(), then this here may be stuck on synchronized ?
Is it possible ?

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/GrpcConnector.java
```diff
@@ -102,43 +105,54 @@ public void shutdown() throws Exception {
...
+            this.eventStreamRetryBackoff = 2 * this.eventStreamRetryBackoff;
 
-    private void restartEventStream() {
-        this.eventStreamAttempt++;
-        if (this.eventStreamAttempt > this.maxEventStreamRetries) {
-            log.error("failed to connect to event stream, exhausted retries");
-            this.stateConsumer.accept(ProviderState.ERROR);
-            return;
+            try {
+                Thread.sleep(this.eventStreamRetryBackoff);
```
liran2000, line 143: Can consider adding [Jitter](https://aws.amazon.com/builders-library/timeouts-retries-and-backoff-with-jitter/).

liran2000 (commented):
Hi, added some comments, hope it's ok.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/EventStreamObserver.java
Kavindu-Dodan: @thisthat good point. `onComplete` is a last call similar to `onError`. So we must reinitialize the listener

Kavindu-Dodan: @liran2000 we do not have multi-thread model here. Let me try to explain, 

We start a thread to handle the stream observer [^1], this is the only extra thread that we create. This thread handles a retry mechanism [^2] and executes an asynchronous `EventStreamObserver`.

`EventStreamObserver` is an implementation of grpc `StreamObserver`. `StreamObserver` implementation is such that `onError` & `onCompleted` calls are called once, and `no further calls to any method are allowed` [^3]

We `notify` waiting caller when dealing with `onError` & `onCompleted`. Given these are the last calls, only the freshly initialized  `EventStreamObserver` can call the `notify` if that too face with an error. 

[^1]: https://github.com/Kavindu-Dodan/java-sdk-contrib/blob/feat/refactor-flagd/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/GrpcConnector.java#L83

[^2]: https://github.com/Kavindu-Dodan/java-sdk-contrib/blob/feat/refactor-flagd/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/GrpcConnector.java#L119

[^3]: https://grpc.github.io/grpc-java/javadoc/io/grpc/stub/StreamObserver.html#onError(java.lang.Throwable)

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/GrpcConnector.java
Kavindu-Dodan: fair point 🤔  I will update this to continue

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/GrpcConnector.java
toddbaert: > Perhaps can use a single threaded (or other) executor, and then shut it down gracefully at shutdown() method ?

Ya, I generally like to use single thread executors instead of threads directly.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/GrpcConnector.java
```diff
@@ -102,43 +103,57 @@ public void shutdown() throws Exception {
...
+        final Object sync = new Object();
+
+        while (this.eventStreamAttempt <= this.maxEventStreamRetries) {
+            StreamObserver<Schema.EventStreamResponse> responseObserver =
+                    new EventStreamObserver(sync, this.cache, this::grpcStateConsumer);
+            this.serviceStub.eventStream(Schema.EventStreamRequest.getDefaultInstance(), responseObserver);
+
+            try {
+                synchronized (sync) {
+                    sync.wait();
```
toddbaert, line 129: I think this wait/notify can be expanded to prevent us from having to do a dummy evaluation, as I described [here](https://github.com/open-feature/java-sdk-contrib/pull/378#discussion_r1281102577).

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/GrpcConnector.java
Kavindu-Dodan: Updated with the latest commit

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/GrpcConnector.java
Kavindu-Dodan: I am revisiting the loigc. will see if we can get rid of dummy call

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/GrpcConnector.java
Kavindu-Dodan: Unfortunately, we need some sort of a concurrent mechanism to execute the connection retry logic [^1]. This is to support, busy wait (sleep) and the loop for retry counter.

However, I agree with the point of graceful shutdown (thread termination on shutdown invocation). I will work on this improvement 

[^1]: https://github.com/Kavindu-Dodan/java-sdk-contrib/blob/feat/refactor-flagd/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/GrpcConnector.java#L115-L144

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/GrpcConnector.java
Kavindu-Dodan: I had to revert this with graceful shutdown support [^1] . This is needed to clean up this thread for provider shutdown. 

[^1]: https://github.com/Kavindu-Dodan/java-sdk-contrib/blob/feat/refactor-flagd/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/GrpcConnector.java#L84-L85

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/GrpcConnector.java
Kavindu-Dodan: @liran2000  good points.

I have added thread cleanup backed by `thread.interrupt` [^1]. And downgraded the logs to debug level. From OpenFeature perspective, API provides shutdown support, which flagd implements. And the new logic respects this shutdown call by accepting an interrupt (then log and exit the thread).

 Once the thread is interrupted and lead to a shutdown, `notify` is no-op and shall not cause any error. Then the GC should take care of the rest of the cleanup  



[^1]: https://github.com/Kavindu-Dodan/java-sdk-contrib/blob/feat/refactor-flagd/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/GrpcConnector.java#L84-L85

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/GrpcConnector.java
Kavindu-Dodan: Thanks for the follow-up. Completed the cleanup with this commit - https://github.com/open-feature/java-sdk-contrib/pull/398/commits/cbaa368a0867f5de6ee23d089aeeae6b792e3eb4

We no longer rely on the dummy call

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/GrpcConnector.java
Kavindu-Dodan: > If this can be called from multiple threads, and the other thread which wait on this object is during wait(), then this here may be stuck on synchronized ?
Is it possible ?

As explained here [^1], we do not use multiple threads. We only have 1-1 relationship from caller to async stream handler.

Given sync block is the standard way to deal with `wait` & `notify` [^2], we shall not face with deadlocks. 

Regarding adding jitter, I think this can be considered a follow-up 

[^1]: https://github.com/open-feature/java-sdk-contrib/pull/398#discussion_r1296076334
[^2]: https://docs.oracle.com/javase/8/docs/api/java/lang/Object.html#notify--

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/GrpcConnector.java
liran2000: thanks, what about the case of handling notify from onCompleted, is this a valid scenario ? as I see in this case, wait() returns without exception, the code then proceeds to sleep() - not sure it this will cause immediate interrupted exception  and return or will finish the sleep and will proceed to unexpected error.
Maybe add log info/debug and return immediately after wait() can make it clear ? what do you think ?
If this specific case is covered by unit test it can be helpful.

Kavindu-Dodan:
> Please, change the wording in the PR description from `Fixes` to `Part of` or something similar. Otherwise the original issue will be closed when not all the points are addressed

Done. However, we can mark `Use always OTel interceptor` can be considered done with the new packaging structure

Kavindu-Dodan:
@thisthat @toddbaert @liran2000  thank you for the reviews 🙌 

I addressed all your concerns. As summary,

- Retry logic now solely rely on the eventing mechanism (no dummy calls)
- Graceful shutdowns for event listener implemented - proper resource cleanup 
- Justification on thread usage along with proper log levels and comments where necessary

liran2000:
@Kavindu-Dodan thanks :) had one [question](https://github.com/open-feature/java-sdk-contrib/pull/398#discussion_r1296273468) above.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/GrpcConnector.java
Kavindu-Dodan: `onCompleted` means a successful stream completion. If flagd provide stream closure without error, then this will handle this scenario. 

We need to handle this as this is also a final call similar to `onError`

When this (either `onError` or `onComplete`) happens, we disconnect from flagd. This can be resulted from a restart (ex:- new version rollout), OOMKILL or similar. From SDK point of view, we attempt to reconnect with backoff. This is why we have a sleep and continue with a new stream watcher connection

Kavindu-Dodan: I will try to add some unit tests to cover this scenario :)

toddbaert:
> @thisthat @toddbaert @liran2000 thank you for the reviews 🙌
> 
> I addressed all your concerns. As summary,
> 
> * Retry logic now solely rely on the eventing mechanism (no dummy calls)
> * Graceful shutdowns for event listener implemented - proper resource cleanup
> * Justification on thread usage along with proper log levels and comments where necessary

Amazing! I will re-review tomorrow.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/GrpcConnector.java
Kavindu-Dodan: Added unit tests to verify retry attempts - https://github.com/open-feature/java-sdk-contrib/pull/398/commits/1af6f4ce52f326bb12482da906957013acade150#diff-0dff792882d8d7962486094010e2c59c4210efa38a4b0f7ee4d5f964934eb3c4

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/GrpcConnector.java
```diff
@@ -61,34 +65,27 @@ public GrpcConnector(final FlagdOptions options, final Cache cache, Consumer<Pro
...
-            if (Status.DEADLINE_EXCEEDED.equals(e.getStatus())) {
-                throw e;
-            }
-        } finally {
-            // try in background to open the event stream
-            this.startEventStream();
-        }
+    public void initialize() {
+        eventObserverThread = new Thread(this::observeEventStream);
+        eventObserverThread.start();
```
toddbaert, line 79: This reads much simpler, but we aren't blocking until we actually get a good connection, like I mentioned [here](https://github.com/open-feature/java-sdk-contrib/pull/378#discussion_r1281102577).

This means that flagd can't be reached, we'll see a READY from the SDK (because initialize returns immediately) and then after that get an error from the thread. Unless I'm missing something, this could be fixed simply by waiting on the `sync` here and notifying on successful connect to unblock.

Kavindu-Dodan: > This means that flagd can't be reached, we'll see a READY from the SDK

This is incorrect. Let me try to explain.

Flagd provider is an `EventProvider` with a default status of `NOT_READY`. WIth the current refactoring, status is fully maintained by event subscriptions, observed concurrently through this logic [^1]. The actual event subscriptions are handled by `EventStreamObserver` [^2].

Regarding initialization returning immediately, given that we have start status of `NOT_READY`, we are sticking to this status till we perform a callback from the event observer. And this happens fast as soon as event observer attempt to connect with flagd. And there are two possibilities that happens at `EventStreamObserver`

- Successful connection - results in a `ProviderReady` sent by flagd as the initial handshake

This makes `EventStreamObserver` to callback with ready status, making provider ready

- Connection failure - results in invoking `onError`

This makes `EventStreamObserver` to callback with`Error` status, making provider status to `ERROR`

Hope this helps to clarify your doubt. But please let me know if I am not seeing something here :) 


[^1]: https://github.com/open-feature/java-sdk-contrib/blob/1af6f4ce52f326bb12482da906957013acade150/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/GrpcConnector.java#L114-L144
[^2]: https://github.com/open-feature/java-sdk-contrib/blob/main/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/EventStreamObserver.java#L19

toddbaert: > Regarding initialization returning immediately, given that we have start status of NOT_READY, we are sticking to this status till we perform a callback from the event observer. And this happens fast as soon as event observer attempt to connect with flagd. And there are two possibilities that happens at EventStreamObserver

Regardless of this, when this `initialize` returns successfully, the SDK fires a READY event. This is what I mean when I said "we'll see a READY from the SDK".

The SDK expects the provider to be in a usable state immediately after the return of `initialize`.

Kavindu-Dodan: Addressed this with the latest commit - https://github.com/open-feature/java-sdk-contrib/pull/398/commits/a21363ead7e869ab9550d180c385dcc36744ea0b

Now we perform a busy wait for a successful connection. This is done with a countdown similar to what you proposed - https://github.com/open-feature/java-sdk-contrib/pull/398/commits/a21363ead7e869ab9550d180c385dcc36744ea0b#diff-67d657d2b09b8caa955560fe7640113ec2bb5528db86e0f96dc15b83e86c9627R172-R182

please check and let me know your thoughts :)

toddbaert: looks good!

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/grpc/GrpcConnector.java
Kavindu-Dodan: Given that adding jitter is a simple change, I added this to our logic. See commit - https://github.com/open-feature/java-sdk-contrib/pull/398/commits/d28bdcd679fef903566697677b7d5a8797417c18

toddbaert (approved):
Manually tested this (though we do have the e2e tests now :sunglasses:). LGTM now. Thanks for the improvements.

Kavindu-Dodan:
> Manually tested this (though we do have the e2e tests now 😎). LGTM now. Thanks for the improvements.

Thank you 🙏 I also did several manual validations against a local flagd instance. But yeah, e2e tests to helps

PR #403, opened 2023-08-17
chore: go feature flags: use sdk functionality

liran2000:
## This PR

- use recently added sdk functionality instead of each provider implementation.
- [issue 405](https://github.com/open-feature/java-sdk-contrib/issues/405) - missing targeting key


liran2000:
@thomaspoignant to review when available.

@toddbaert to see how some recent SDK functionality can help providers by using it instead of needing to implement for each provider.

Kavindu-Dodan (dismissed):

providers/go-feature-flag/src/main/java/dev/openfeature/contrib/providers/gofeatureflag/GoFeatureFlagProvider.java
```diff
@@ -262,15 +260,10 @@ private <T> ProviderEvaluation<T> getEvaluation(
...
-                    errorCode = ErrorCode.GENERAL;
+                if (ProviderState.NOT_READY.equals(state)) {
+                    throw new ProviderNotReadyError("provider not initialized yet");
                 }
-                return ProviderEvaluation.<T>builder()
-                        .errorCode(errorCode)
-                        .reason(errorCode.name())
-                        .value(defaultValue)
-                        .build();
+                throw new GeneralError("unknown error, provider state: " + state);
```
Kavindu-Dodan: I think @thomaspoignant needs to approve the usage of evaluation response with error vs throwing an exception

liran2000: see [this](https://github.com/open-feature/java-sdk/pull/561#discussion_r1295058795) comment by @toddbaert :

> throwing like this ends in the same result, but it's a bit easier to trace, and also simpler to write. This is how we generally recommend it's done. There is a small performance cost in serialization of the exception, but I think it's worth it.

When is ProviderNotReadyError thrown here, it should be handled by the SDK framework, ErrorCode.PROVIDER_NOT_READY and default value should be returned when evaluated via the client.

I just added a [test](https://github.com/open-feature/java-sdk-contrib/pull/403/files#diff-3d0ec86c96fe8d2f9224111448f23a4ccad41ee847a904fb50fd15921623b7faR187) step to enforce it.

As it is not always clear and trivial that throwing ProviderNotReadyError is handled by the framework, if you have any suggestion on how to make it clear in documentation/spec/provider guidelines, you are welcome to share your thoughts.

Kavindu-Dodan: > As it is not always clear and trivial that throwing ProviderNotReadyError is handled by the framework..

Yeah, that's true. It's not trivial. Also, we have an error logging for exception [^1] which makes this slightly different.

I will wait for @thomaspoignant approval on this :) Otherwise, change looks good 

[^1]: https://github.com/open-feature/java-sdk/blob/v1.5.0/src/main/java/dev/openfeature/sdk/OpenFeatureClient.java#L140

providers/go-feature-flag/README.md
```diff
@@ -16,7 +16,21 @@ FeatureProvider provider = new GoFeatureFlagProvider(
...
   .timeout(1000)
   .build());
+
+OpenFeatureAPI.getInstance().setProviderAndWait(providerName);
+
+// ...
+
+Client client = OpenFeatureAPI.getInstance().getClient(providerName);
+
+// targetingKey is mandatory for each evaluation
```
liran2000, line 26: @thomaspoignant if you can confirm that this is correct

providers/go-feature-flag/src/main/java/dev/openfeature/contrib/providers/gofeatureflag/GoFeatureFlagProvider.java
```diff
@@ -476,52 +474,6 @@ private <T> T convertValue(Object value, Class<?> expectedType) {
         return (T) objectToValue(value);
```
thomaspoignant, line 474: Good catch, I missed that the SDK was implementing that directly.

providers/go-feature-flag/src/test/java/dev/openfeature/contrib/providers/gofeatureflag/GoFeatureFlagProviderTest.java
```diff
@@ -157,10 +164,57 @@ void constructor_options_valid_endpoint() {
...
+    void client_test() {
         GoFeatureFlagProvider g = new GoFeatureFlagProvider(GoFeatureFlagProviderOptions.builder().endpoint(this.baseUrl.toString()).timeout(1000).build());
-        assertEquals(ErrorCode.PROVIDER_NOT_READY, g.getBooleanEvaluation("fail_not_initialized", false, this.evaluationContext).getErrorCode());
+
+        String providerName = "clientTest";
+        OpenFeatureAPI.getInstance().setProvider(providerName, g);
+
+        /*
+         // TODO replace to setProviderAndWait and remove this when
+         https://github.com/open-feature/java-sdk/pull/563 is merged and released
```
thomaspoignant: The PR is merged now, I guess we can refactor this.

thomaspoignant (approved):
Super nice contribution, thanks a lot this makes the code way more readable.

providers/go-feature-flag/src/main/java/dev/openfeature/contrib/providers/gofeatureflag/GoFeatureFlagProvider.java
liran2000: It was just recently added.

liran2000:
@toddbaert can this be merged?

thomaspoignant:
@liran2000 I've merged it, thanks a lot again for your contribution 🙏

PR #404, opened 2023-08-17
chore: improve netty usage

### LINKED ISSUE #402: [flagd-provider] add guard before loading Epoll classes
As per the recommendation from netty, we should [add a guard](https://github.com/netty/netty/issues/13523#issuecomment-1680917605) before loading the Epoll classes. It seems like they will revert the change that caused our issue, but using this guard will also help us to give helpful error messages if this native lib is missing.

cc @Kavindu-Dodan

Kavindu-Dodan:
## This PR

Fixes https://github.com/open-feature/java-sdk-contrib/issues/402

Improves netty usage by verifying epoll availability.  <del>Added a test to validate the check</del> Unfortunately, this is not testable. 

@toddbaert I had to remove two tests, which are not compatible with non-linux systems.

### DIFF
--- a/providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/grpc/GrpcConnectorTest.java
+++ b/providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/grpc/GrpcConnectorTest.java
@@ -7,9 +7,15 @@ import dev.openfeature.flagd.grpc.ServiceGrpc;
 import io.grpc.Channel;
 import io.grpc.netty.NettyChannelBuilder;
 import io.netty.channel.EventLoopGroup;
+import io.netty.channel.epoll.EpollEventLoopGroup;
+import io.netty.channel.unix.DomainSocketAddress;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.condition.EnabledIf;
+import org.junit.jupiter.api.condition.EnabledOnOs;
+import org.junit.jupiter.api.condition.OS;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.ValueSource;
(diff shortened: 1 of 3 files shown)


providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/grpc/GrpcConnectorTest.java
```diff
@@ -123,80 +117,6 @@ void initialization_fail_with_timeout() throws Exception {
         assertThrows(RuntimeException.class, connector::initialize);
     }
 
-    @Test
```
toddbaert, line 126: @Kavindu-Dodan maybe this is a dumb question - could we keep these tests, but run them conditionally if epoll IS available?

toddbaert: We could perhaps use: https://www.baeldung.com/junit-5-conditional-test-execution#2-conditions

Kavindu-Dodan: 🤔 hmm yeah that's possible. let me try to reenable and check the build

Kavindu-Dodan: @toddbaert  added them with conditional execution `@EnabledOnOs(OS.LINUX)` - https://github.com/open-feature/java-sdk-contrib/pull/404/commits/35d293bbea9d60b13bd82c9e8483388629fc4d46

This is ideal solution and thanks for recommending this :)

toddbaert (approved):

PR #407, opened 2023-08-23
chore: generate automatic module names

### LINKED ISSUE #261: Add automatic module names to all packages
See [here](https://github.com/open-feature/java-sdk/pull/363), and the linked issue for details and background. [This blog post](http://branchandbound.net/blog/java/2017/12/automatic-module-name/) is also helpful, specifically for library maintainers.

In each module in this project, we should add this maven step (we could also add it in the parent POM and somehow grab the name dynamically). We should be careful that the module name matches the "highest level" package in the module.

toddbaert:
See title.

Tested this locally with `mvn install`, then opened the locally cached module. This is what the MANIFEST.MF looks like for the flagd provider:

```
Manifest-Version: 1.0
Automatic-Module-Name: dev.openfeature.contrib.providers.flagd
Build-Jdk-Spec: 1.8
Created-By: Maven JAR Plugin 3.3.0
```

Note: I had to "make up" some module names for a few providers that would otherwise have invalid ones based on the module name (apparently "-" is illegal).

:warning: Note: http://branchandbound.net/blog/java/2017/12/automatic-module-name/ and https://blog.joda.org/2017/04/java-se-9-jpms-module-naming.html _strongly_ recommend that module names should match the `root package` of the jar. I've therefore used the root package for the "-" cases.

Fixes: https://github.com/open-feature/java-sdk-contrib/issues/261

### DIFF
--- a/providers/env-var/pom.xml
+++ b/providers/env-var/pom.xml
@@ -14,7 +14,7 @@
 
     <properties>
         <!-- override module name defined in parent ("-" is not allowed) -->
-        <module-name>${groupId}.env</module-name>
+        <module-name>${groupId}.envvar</module-name>
     </properties>
 
     <name>env-var</name>
--- a/providers/env-var/pom.xml
+++ b/providers/env-var/pom.xml
@@ -12,6 +12,11 @@
     <artifactId>env-var</artifactId>
     <version>0.0.4</version> <!--x-release-please-version -->
 
+    <properties>
+        <!-- override module name defined in paraent ("-" is not allowed) -->
+        <module-name>${groupId}.env</module-name>
+    </properties>
+
     <name>env-var</name>
     <description>Environment Variables provider for Java</description>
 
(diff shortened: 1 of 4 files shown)


madhead (approved):

beeme1mr (approved):

Kavindu-Dodan (approved):

providers/env-var/pom.xml
```diff
@@ -12,6 +12,11 @@
     <artifactId>env-var</artifactId>
     <version>0.0.4</version> <!--x-release-please-version -->
 
+    <properties>
+        <!-- override module name defined in parent ("-" is not allowed) -->
+        <module-name>${groupId}.envvar</module-name>
```
Kavindu-Dodan, line 17: +1 for name suggestion

However, another alternative is to use underscore `_` , but this is also fine

toddbaert: I think if we'd do that, we'd be breaking the "root package" == module name" rule (unless we renamed the packages).

Issue #411, opened 2023-08-29
[flagd] Intial version of in-process provider


Kavindu-Dodan:
As mentioned in flagd documentation of  **Creating an in-process flagd provider** [^1], in-process provider comes with unique advantages. This issue focuses on creating the first version of flagd in-process provider in Java.

**Requirements**

- Fulfill requirements outlined by flagd in-process guideline [^1]
- Minimal external dependencies 
- Updated/New ReadMe & usage guide
- Unit & Integration tests 

**Technical requirements**

- Use the same repository directory structure (`providers/flagd`)
- Build both remote and in-process providers into a single dependency (To be decided after the POC)
- Reuse existing interfaces and implementations where possible 

**Tasks** 

- [x] POC implementation and build to detect packaging requirements (no evaluations needed)
- [x] In-process provider with gRPC flag sync capability on basic flags
- [x] Support for JsonLogic evaluators (complex flagd flag configurations)
- [x] Support for flagd custom JsonLogic evaluators 

[^1]: https://github.com/open-feature/flagd/blob/main/docs/other_resources/in-process-providers/specification.md

PR #412, opened 2023-09-01
feat: flagd in-process provider

Kavindu-Dodan:
**This PR**

Completes first set of tasks proposed through https://github.com/open-feature/java-sdk-contrib/issues/411 

Please review with checkout as this contains refactoring + implementation of the in-process resolver for flagd.

**Design**

Consider the following diagram for architecture,

image

**Highlights**

- No API changes, hence no breaking changes
- Resolver selection through builder option,

```Java
       // This builds a flagd provider with in-memory resolving
        new FlagdProvider(
                FlagdOptions.builder()
                        .resolverType(Config.ResolverType.inProcess)
                        .build());
```
- Extensibility through layering and contracts 
- Feature parity with grpc evaluator (i.e - Eventing, targeting rules and matching evaluation payloads)


providers/flagd/pom.xml
```diff
@@ -10,7 +10,7 @@
     </parent>
     <groupId>dev.openfeature.contrib.providers</groupId>
     <artifactId>flagd</artifactId>
-    <version>0.6.1</version> <!--x-release-please-version -->
+    <version>999-snapshot</version> <!--x-release-please-version -->
```
Kavindu-Dodan: Todo - fix me

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/Config.java
```diff
@@ -49,4 +50,19 @@ static int fallBackToEnvOrDefault(String key, int defaultValue) {
...
     }
+
+    /**
+     * flagd resolving type.
+     */
+    public enum ResolverType {
+        /**
+         * This is the default resolver type, which connects to flagd instance with flag evaluation gRPC contract.
+         */
+        flagd,
```
thisthat: usually enum are capitalized (see `dev.openfeature.contrib.providers.flagd.cache.CacheType`)

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/FlagStore.java
```diff
@@ -0,0 +1,101 @@
...
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.logging.Level;
+
+/**
+ * Feature flag storage.
+ */
+@Log
+public class FlagStore implements Storage {
+    private final Object sync = new Object();
```
thisthat: I think a `ReentrantReadWriteLock` is better here since we will do many more reads than writes

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/FlagStore.java
```diff
@@ -0,0 +1,101 @@
...
+     */
+    public void shutdown() {
+        shutdown.set(true);
+        connector.shutdown();
+    }
+
+    /**
+     * Retrieve flag for the given key.
+     */
+    public FeatureFlag getFLag(final String key) {
```
thisthat:
```suggestion
    public FeatureFlag getFlag(final String key) {
```

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolver.java
```diff
@@ -0,0 +1,199 @@
...
+                final Object jsonResolved = jsonLogicHandler.apply(flag.getTargeting(), ctx.asObjectMap());
+                if (jsonResolved == null) {
+                    resolvedVariant = flag.getDefaultVariant();
+                    reason = Reason.DEFAULT.toString();
+                } else {
+                    resolvedVariant = jsonResolved;
+                    reason = Reason.TARGETING_MATCH.toString();
+                }
+            } catch (JsonLogicException e) {
+                log.log(Level.FINE, "Error evaluating targeting rule", e);
```
thisthat, line 172: I think we should be more verbose telling the users that something is off. Maybe this could be a possible configuration option (no need to tackle as part of this PR)
```suggestion
                log.log(Level.WARNING, "Error evaluating targeting rule", e);
```

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolver.java
```diff
@@ -0,0 +1,199 @@
...
+                return ProviderEvaluation.<T>builder()
+                        .value(defaultValue)
+                        .reason(Reason.ERROR.toString())
+                        .errorCode(ErrorCode.PARSE_ERROR)
+                        .errorMessage(String.format("error parsing targeting rule: %s", key))
```
thisthat:
```suggestion
                        .errorMessage(String.format("error parsing targeting rule for flag: %s", key))
```
or 
```suggestion
                        .errorMessage(String.format("error parsing targeting rule: %s", flag.getTargeting()))
```

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/model/FeatureFlag.java
```diff
@@ -0,0 +1,40 @@
...
+        this.defaultVariant = defaultVariant;
+        this.variants = variants;
+        this.targeting = targeting;
+    }
+
+    /**
+     * Get targeting rule of the flag.
+     */
+    public String getTargeting() {
+        return this.targeting == null ? "{}" : this.targeting;
```
thisthat: maybe let's extract `{}` into a public static final class field so it can be used also in [InProcessResolver.java](https://github.com/open-feature/java-sdk-contrib/pull/412/files#diff-27f34c6acd23d899adf4c27d00f4d5b9e730090bb2a88efc9a286f3da0bd9dcc)

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/model/FlagParser.java
```diff
@@ -0,0 +1,81 @@
...
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.regex.Pattern;
+
+/**
+ * flagd feature flag configuration parser.
+ */
+public class FlagParser {
```
thisthat, line 31: have you checked if we could use the schema available here? https://github.com/open-feature/playground/blob/main/schemas/flag.schema.json

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/FlagStore.java
```diff
@@ -0,0 +1,101 @@
...
+        });
+        streamer.setDaemon(true);
+        streamer.start();
+    }
+
+    /**
+     * Shutdown storage layer.
+     */
+    public void shutdown() {
+        shutdown.set(true);
```
thisthat: since we have the info, shall we protect the method from being called multiple times?
```suggestion
     if(!shutdown.getAndSet(true)){
```

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/grpc/GrpcStreamConnector.java
```diff
@@ -0,0 +1,129 @@
...
+ */
+@Log
+public class GrpcStreamConnector implements Connector {
+    private static final Random RANDOM = new Random();
+
+    private static final int INIT_BACK_OFF = 2 * 1000;
+    private static final int MAX_BACK_OFF = 120 * 1000;
+
+    private final AtomicBoolean shutdown = new AtomicBoolean(false);
+    private final BlockingQueue<StreamPayload> blockingQueue = new LinkedBlockingQueue<>(5);
```
thisthat: I wonder what could be a good experiment to run for identifying the right magic number for the queue capacity 🤔 do you have any idea? @toddbaert  @Kavindu-Dodan

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/grpc/GrpcStreamConnector.java
```diff
@@ -0,0 +1,129 @@
...
+                    case SYNC_STATE_ALL:
+                        if (!writeTo.offer(
+                                new StreamPayload(StreamPayloadType.Data, flagsResponse.getFlagConfiguration()))) {
+                            log.log(Level.WARNING, "Stream writing failed");
+                        }
+                        break;
+                    case SYNC_STATE_UNSPECIFIED:
+                    case SYNC_STATE_ADD:
+                    case SYNC_STATE_UPDATE:
+                    case SYNC_STATE_DELETE:
```
thisthat, line 125: I believe we should cover this, otherwise single flag changes will not be propagated - maybe we can add a TODO or a log line for the moment

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
```diff
@@ -42,7 +44,15 @@ public FlagdProvider() {
...
-        this.flagResolver = new GrpcResolution(options, CacheFactory.getCache(options), this::getState, this::setState);
+        switch (options.getResolverType()) {
+            case inProcess:
+                this.flagResolver = new InProcessResolver(options, this::setState);
+                break;
+            case flagd:
+            default:
+                this.flagResolver =
+                        new GrpcResolver(options, CacheFactory.getCache(options), this::getState, this::setState);
+        }
```
odubajDT, line 55: I would add a default case returning error: "invalid resolver type"

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
```diff
@@ -42,7 +44,15 @@ public FlagdProvider() {
      * @param options {@link FlagdOptions} with
      */
     public FlagdProvider(final FlagdOptions options) {
-        this.flagResolver = new GrpcResolution(options, CacheFactory.getCache(options), this::getState, this::setState);
+        switch (options.getResolverType()) {
+            case inProcess:
+                this.flagResolver = new InProcessResolver(options, this::setState);
+                break;
+            case flagd:
+            default:
```
odubajDT: I would say the default path should return an error: invalid resolved type

thisthat: we should never fail in the library IMHO. However, a log line would be nice 🤔

odubajDT: also a posibility :+1:

odubajDT: here it's done nicely https://github.com/open-feature/java-sdk-contrib/pull/412/files#diff-27f34c6acd23d899adf4c27d00f4d5b9e730090bb2a88efc9a286f3da0bd9dccR58

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolver.java
```diff
@@ -0,0 +1,199 @@
...
+                        .build();
+            }
+        }
+
+        // check variant existence
+        Object value = flag.getVariants().get(resolvedVariant);
+        if (value == null) {
+            log.log(Level.FINE, String.format("variant %s not found in flag with key %s", resolvedVariant, key));
```
odubajDT, line 185:
```suggestion
            log.log(Level.WARNING, String.format("variant %s not found in flag with key %s", resolvedVariant, key));
```

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolver.java
```diff
@@ -0,0 +1,199 @@
...
+            return ProviderEvaluation.<T>builder()
+                    .value(defaultValue)
+                    .reason(Reason.ERROR.toString())
+                    .errorCode(ErrorCode.TYPE_MISMATCH)
+                    .errorMessage(String.format("requested flag is not of the evaluation type: %s", type.getName()))
+                    .build();
+        }
+
+        if (!value.getClass().isAssignableFrom(type) || !(resolvedVariant instanceof String)) {
+            log.log(Level.FINE, String.format("returning default variant for flagKey: %s, type not valid", key));
```
odubajDT, line 195:
```suggestion
            log.log(Level.WARNING, String.format("returning default variant for flagKey: %s, type not valid", key));
```

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/FlagStore.java
Kavindu-Dodan: good point, yes will update this :)

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolver.java
Kavindu-Dodan: I will update this to warning

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/FlagStore.java
Kavindu-Dodan: Good one, will update

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/grpc/GrpcStreamConnector.java
Kavindu-Dodan: I don't think we need to worry much about this. Readers are all about parsing payload and writing flags to a map. This is fast enough compared to gRPC stream coming through wire at write end.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/grpc/GrpcStreamConnector.java
Kavindu-Dodan: @toddbaert and I discussed these states. While these payloads do show benefits, I don't know of any specific implementation that uses these. I think someday we could move to a V2 spec without these states and completely rely on SYNC_ALL (full flag configuration) only.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
Kavindu-Dodan: I will add a log line :)

Kavindu-Dodan: ohh actually we can throw an error given this is a configuration step. OF Spec allows this behavior.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolver.java
Kavindu-Dodan: on second thought, incorrect flag configurations could lead to log overflow. hence I will keep this as `Level.FINE` (which is basically Debug level logs)

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolver.java
Kavindu-Dodan: As highlighted at https://github.com/open-feature/java-sdk-contrib/pull/412#discussion_r1316201469 , this could lead to log overflow with incorrect flag configurations/API invocations

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolver.java
Kavindu-Dodan: https://github.com/open-feature/java-sdk-contrib/pull/412#discussion_r1316202124

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolver.java
thisthat: Good point! We can iterate over with a throttle logger in the future if we see the needs :)

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/Config.java
Kavindu-Dodan: Done with the latest changes

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/FlagStore.java
Kavindu-Dodan: Introduced `ReentrantReadWriteLock `

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolver.java
Kavindu-Dodan: Corrected :)

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/FlagStore.java
Kavindu-Dodan: Done :)

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/model/FlagParser.java
Kavindu-Dodan: We now use networknt json-schema-validator and validate the payload before accepting

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/model/FlagParser.java
```diff
@@ -0,0 +1,131 @@
...
+     * Parse {@link String} for feature flags.
+     */
+    public static Map<String, FeatureFlag> parseString(final String configuration) throws IOException {
+        if (SCHEMA_VALIDATOR != null) {
+            try (JsonParser parser = MAPPER.createParser(configuration)) {
+                Set<ValidationMessage> validationMessages = SCHEMA_VALIDATOR.validate(parser.readValueAsTree());
+
+                if (!validationMessages.isEmpty()) {
+                    throw new IllegalArgumentException(
+                            String.format("Failed to parse configurations. %d validation " + "error(s) reported.",
```
thisthat:
```suggestion
                            String.format("Failed to parse configurations. %d validation error(s) reported.",
```

thisthat (commented):
The implementation LGTM, great work @Kavindu-Dodan 👏 
I suggest adding some unit tests for the gRPC Stream Handler/Connector and the FlagStore.
Please, also adapt the main readme with the new usage options :) I suggest also posting there (or in another doc) the nice diagram that you attached to this PR

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
```diff
@@ -42,7 +44,18 @@ public FlagdProvider() {
...
+            case IN_PROCESS:
+                this.flagResolver = new InProcessResolver(options, this::setState);
+                break;
+            case FLAGD:
+                this.flagResolver =
+                        new GrpcResolver(options, CacheFactory.getCache(options), this::getState, this::setState);
+                break;
+            default:
+                throw new IllegalStateException(
+                        String.format("Requested unsupported resolver type of %s", options.getResolverType()));
```
odubajDT, line 57: didn't we agreed only on logging the error here ?

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/grpc/GrpcStreamConnector.java
```diff
@@ -0,0 +1,137 @@
...
+
+    /**
+     * Contains blocking calls, to be used concurrently.
+     */
+    static void observeEventStream(final BlockingQueue<StreamPayload> writeTo,
+                                   final AtomicBoolean shutdown,
+                                   final FlagSyncServiceGrpc.FlagSyncServiceStub serviceStub)
+            throws InterruptedException {
+
+        final BlockingQueue<GrpcResponseModel> blockingQueue = new LinkedBlockingQueue<>(5);
```
odubajDT: we use number 5 on multiple places, is it worth to introduce a constant and use that instead ?

providers/flagd/pom.xml
```diff
@@ -164,60 +203,61 @@
...
-                    </arguments>
-                    </configuration>
-                </execution>
-                </executions>
-            </plugin>
-            </plugins>
-        </build>
-        </profile>
-    </profiles>
+    <!-- Disabled till Gherkin test suites are ready -->
```
toddbaert: I think the `evaluation.feature` should still work. I could be wrong, but I would guess we can re-enable these for the plain old RPC provider and it will work fine.

I can add the flagd specific tests (`flagd.feature`) later.

providers/flagd/pom.xml
```diff
@@ -121,12 +145,12 @@
                                 <argument>submodule</argument>
                                 <argument>update</argument>
                                 <argument>--init</argument>
-                                <argument>schemas</argument>
+                                <argument>--remote</argument>
```
toddbaert: This change makes the build non-deterministic. Every time it runs it will attempt to update the submodule, regardless of the previously set tag. It effectively removes the "versioning" here. If you need a more recent version, you should `git pull`/`git checkout` in the submodule, and commit the result.

See https://git-scm.com/book/en/v2/Git-Tools-Submodules

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/common/ChannelBuilder.java
```diff
@@ -0,0 +1,71 @@
...
+import io.netty.channel.unix.DomainSocketAddress;
+import io.netty.handler.ssl.SslContextBuilder;
+
+import javax.net.ssl.SSLException;
+import java.io.File;
+
+/**
+ * gRPC channel builder helper.
+ */
+public class ChannelBuilder {
```
toddbaert, line 20: Does this mean we can (hypothetically) use unix sockets for the sync connection too?

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolver.java
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    /**
+     * Resolve an object flag.
+     */
+    public ProviderEvaluation<Value> objectEvaluation(String key, Value defaultValue, EvaluationContext ctx) {
+        return resolveGeneric(Value.class, key, defaultValue, ctx);
+    }
+
+    private <T> ProviderEvaluation<T> resolveGeneric(Class<T> type, String key, T defaultValue,
```
toddbaert:
```suggestion
    private <T> ProviderEvaluation<T> resolve(Class<T> type, String key, T defaultValue,
```

nit, but I think the fact this is generic is communicated by the signature.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/model/FlagParser.java
```diff
@@ -0,0 +1,131 @@
...
+            while (it.hasNext()) {
+                final String key = it.next();
+                flagMap.put(key, MAPPER.treeToValue(flagNode.get(key), FeatureFlag.class));
+            }
+        }
+
+        return flagMap;
+    }
+
+    private static String transposeEvaluators(final String configuration) throws IOException {
```
toddbaert, line 100: This reminds me that I should add shared evaluators to the e2e test harness: https://github.com/open-feature/flagd-testbed/pull/54. I didn't add those.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolver.java
```diff
@@ -0,0 +1,203 @@
...
+     */
+    public void init() throws Exception {
+        flagStore.init();
+        final Thread stateWatcher = new Thread(() -> {
+            try {
+                while (true) {
+                    final StorageState storageState = flagStore.getStateQueue().take();
+                    switch (storageState) {
+                        case OK:
+                            stateConsumer.accept(ProviderState.READY);
```
toddbaert, line 56: If I'm understanding correctly, if we are already `READY`, this will emit a PROVIDER_CONFIGURATION_CHANGED event due to [this](https://github.com/open-feature/java-sdk-contrib/pull/412/files#diff-7ac82836e0450dd1f6581536670fba629b6fa21a8ccb4a5ffcdce54a53dd57e3R137-R143).

Can you confirm my understanding?

providers/flagd/src/main/resources/json-schema-v0.1.1
```diff
@@ -0,0 +1,169 @@
+{
```
toddbaert: You can copy this from the `schemas` submodule just like the protos: https://github.com/open-feature/java-sdk-contrib/pull/412#diff-604480b711e56c4eb6ae627a3596fae16ae30ef31e7da7b71be05cdc8fbd38d8R167-R181

toddbaert (changes requested):
This looks really good! I added a few suggestions/nits.

The things I think need changing are [this](https://github.com/open-feature/java-sdk-contrib/pull/412#discussion_r1317484595) and [this](https://github.com/open-feature/java-sdk-contrib/pull/412/files#r1317434395).

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolver.java
toddbaert: Do you think DEBUG is better? I honestly never think of `fine` when I think of log levels.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/model/FeatureFlag.java
Kavindu-Dodan: this is done :)

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolver.java
Kavindu-Dodan: @toddbaert there's no DEBUG type exposed. FINE is the closest level [^1]  we have.

[^1]: https://docs.oracle.com/javase/8/docs/api/java/util/logging/class-use/Level.html

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
Kavindu-Dodan: This aligns with the OF spec [^1]

`... Exceptions include functions or methods for the purposes for configuration or setup.`

[^1]: https://openfeature.dev/specification/sections/flag-evaluation#requirement-149

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/common/ChannelBuilder.java
Kavindu-Dodan: yeah, this is possible. I reused our channel builder so that we have the option if needed.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolver.java
Kavindu-Dodan: Yup, this is correct. READY -> READY means there is a fresh configuration update.

providers/flagd/src/main/resources/json-schema-v0.1.1
Kavindu-Dodan: This is a really good one. Will try to get this done.

providers/flagd/pom.xml
Kavindu-Dodan: yeah, this is true. I had trouble fetching the latest schema updates and used this to force an update. I will try to fix this

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/grpc/GrpcStreamConnector.java
Kavindu-Dodan: done :)

providers/flagd/pom.xml
Kavindu-Dodan: I fixed the test harness submodule for the previous release. Now the e2e tests are back :)

providers/flagd/pom.xml
Kavindu-Dodan: Fixed and reverted :)

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolver.java
Kavindu-Dodan: Done with the latest commit

providers/flagd/src/main/resources/json-schema-v0.1.1
Kavindu-Dodan: Done, thanks for this great suggestion !!

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/model/FlagParser.java
Kavindu-Dodan: Yup, we need $ref rules and validate them in e2e tests :)

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/FlagStoreTest.java
```diff
@@ -0,0 +1,74 @@
...
+
+        // Shutdown handling
+        store.shutdown();
+
+        assertTimeoutPreemptively(Duration.ofMillis(maxDelay), ()-> {
+            assertEquals(StorageState.ERROR,  states.take());
+        });
+    }
+
+}
```
thisthat: nit
```suggestion
}

```

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/grpc/GrpcStreamConnectorTest.java
```diff
@@ -0,0 +1,132 @@
...
+
+        final FlagSyncServiceGrpc.FlagSyncServiceStub stubMock =
+                Mockito.mock(FlagSyncServiceGrpc.FlagSyncServiceStub.class);
+
+        serviceStubField.set(connector, stubMock);
+
+        return stubMock;
+    }
+
+}
```
thisthat: nit
```suggestion
}

```

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/resolver/process/model/FlagParserTest.java
```diff
@@ -0,0 +1,66 @@
...
+        });
+    }
+
+    @Test
+    public void invalidConfigurationsThrowsError() {
+        assertThrows(IllegalArgumentException.class, () -> {
+            FlagParser.parseString(getFlagsFromResource(INVALID_CFG));
+        });
+    }
+}
```
thisthat: nit
```suggestion
}

```

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolverTest.java
```diff
@@ -0,0 +1,267 @@
...
+        Field flagStore = InProcessResolver.class.getDeclaredField("flagStore");
+        flagStore.setAccessible(true);
+
+        InProcessResolver resolver = new InProcessResolver(FlagdOptions.builder().build(), stateConsumer);
+        flagStore.set(resolver, storage);
+
+        return resolver;
+    }
+
+}
```
thisthat: nit
```suggestion
}

```

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolverTest.java
```diff
@@ -0,0 +1,267 @@
...
+            Assertions.assertEquals(ProviderState.READY, receiver.take());
+        });
+
+        assertTimeoutPreemptively(Duration.ofMillis(200), () -> {
+            Assertions.assertEquals(ProviderState.ERROR, receiver.take());
+        });
+    }
+
+    @Test
+    public void simpleResolving() throws Exception {
```
thisthat: maybe worth testing also for numeric values: int and double

thisthat (approved):
LGTM, I only left few nits for you to consider

odubajDT (approved):

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolver.java
toddbaert: That's FINE then. :stuck_out_tongue:

providers/flagd/README.md
```diff
@@ -27,26 +32,43 @@ FlagdProvider flagd = new FlagdProvider();
 OpenFeatureAPI.getInstance().setProvider(flagd);
 ```
 
+### In-process resolver
+
+This mode perform flag evaluations locally(in-process). Flag configurations for evaluations are obtained via gRPC protocol using [sync protocol buffer](https://buf.build/open-feature/flagd/file/main:sync/v1/sync_service.proto) service definition.
```
```suggestion
This mode perform flag evaluations locally(in-process). Flag configurations for evaluations are obtained via gRPC protocol using [sync protobuf schema](https://buf.build/open-feature/flagd/file/main:sync/v1/sync_service.proto) service definition.
```

toddbaert (approved):
Great work @Kavindu-Dodan. There's a lot here, so I would be surprised if we didn't miss _something_, but I can't see any issues!

Once we merge this I will happily attempt to implement the test suite [here](https://github.com/open-feature/flagd-testbed/pull/54).

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/grpc/GrpcStreamConnector.java
Kavindu-Dodan: I will resolve this conversation and follow-up if we see the need

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolverTest.java
Kavindu-Dodan: Added them in the latest commit :)

PR #415, opened 2023-09-05
feat: add Togglz provider

liran2000:
Add Togglz provider.

### Related Issues
Togglz [issue](https://github.com/togglz/togglz/issues/970).

### Notes

- Unofficial provider implementation for Togglz.
- Tested by unit test using Togglz functionalities.


liran2000:
Hi @bennetelli, @hennr,

you can have a look and share your thoughts.

Thanks

providers/togglz/src/main/java/dev/openfeature/contrib/providers/togglz/TogglzProvider.java
```diff
@@ -0,0 +1,99 @@
...
+            throw new GeneralError("unknown error");
+        }
+        Feature feature = features.get(key);
+        if (feature == null) {
+            throw new FlagNotFoundError("flag " + key + " not found");
+        }
+        boolean featureBooleanValue = feature.isActive();
+        return ProviderEvaluation.<Boolean>builder()
+            .value(featureBooleanValue)
+            .reason(Reason.TARGETING_MATCH.name())
```
beeme1mr, line 75: You may want to use the reason `STATIC` since it doesn't look like evaluation context is being used.

https://openfeature.dev/specification/types#resolution-details

providers/togglz/README.md
```diff
@@ -0,0 +1,7 @@
+# Unofficial Togglz OpenFeature Provider for Java
+
+## Usage
+See [TogglzProviderTest.java](./src/test/java/dev/openfeature/contrib/providers/togglz/TogglzProviderTest.java)
```
beeme1mr: It can be addressed in a follow-up PR, but please add more information to the readme. Please include information such as install instructions, basic usage, limitations (only booleans are supported), etc.

providers/togglz/src/main/java/dev/openfeature/contrib/providers/togglz/TogglzProvider.java
liran2000: Not sure that STATIC is correct here, since [Activation ](https://www.togglz.org/documentation/activation-strategies)can be done by user even without passing context if I understand correctly.
@bennetelli, @hennr can share their thoughts.

beeme1mr: Wouldn't you need to include `FeatureUser` when calling `isActive`?

bennetelli:
@liran2000 will have a look at it tomorrow. Thanks for your PR :) it's has been on my list for a while as well

toddbaert:
@liran2000 I'll give this a proper review soon from a general OpenFeature perspective.

In the meantime, could you add yourself (and anyone else relevant, maybe @bennetelli ?) to the [component owners](https://github.com/open-feature/java-sdk-contrib/blob/main/.github/component_owners.yml)? We're trying to identify owners for all new components. It functions somewhat like CODEOWNERS but just adds you as assignee when these files change.

providers/togglz/src/main/java/dev/openfeature/contrib/providers/togglz/TogglzProvider.java
toddbaert: @beeme1mr I don't think so. It seems like Togglz uses pre-configured features/beans to get these values contextually. For example, to get the user name  you configure one of these: https://www.togglz.org/documentation/authentication, which take the user-name from the servlet context, etc.

This is actually configured in the `feature` (see the constructor) if I understand correctly.

I'm honestly not sure how to properly determine the reason with this provider.

liran2000:
hi @bennetelli, can you review this from Togglz point of view ?

PR #424, opened 2023-09-07
feat: add Unleash provider

liran2000:
add Unleash provider.

See [Readme ](https://github.com/open-feature/java-sdk-contrib/compare/main...liran2000:java-sdk-contrib:feature/unleash?expand=1#diff-33b37a3e9b9bd14c4d0853c582e47b468d3927d41da53b47900a8c60a90b1133) for details.

[Discussed](https://github.com/orgs/Unleash/discussions/3896) at Unleash.


liran2000:
hi @pransh15, @chriswk, @ivarconr, @gastonfournier, 

you can have a look and share your thoughts.

Also to be reviewed from OpenFeature perspective.

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
```diff
@@ -0,0 +1,212 @@
...
+        }
+        UnleashContext context = ctx == null ? UnleashContext.builder().build() : transform(ctx);
+        Payload defaultVariantPayload = new Payload("string", String.valueOf(defaultValue));
+        Variant defaultVariant = new Variant("default_fallback", defaultVariantPayload, true);
+        Variant evaluatedVariant = unleash.getVariant(key, context, defaultVariant);
+        Payload evaluatedVariantPayload = evaluatedVariant.getPayload().orElse(defaultVariantPayload);
+        String evaluatedVariantPayloadValue = evaluatedVariantPayload.getValue();
+        return ProviderEvaluation.<String>builder()
+            .value(evaluatedVariantPayloadValue)
+            .reason(Reason.UNKNOWN.name())
```
gastonfournier: Maybe we can avoid setting the reason, apparently it's optional: https://openfeature.dev/docs/reference/concepts/evaluation-api/#evaluation-details-structure-fields

providers/unleash/src/test/java/dev/openfeature/contrib/providers/unleash/UnleashProviderIntegrationTest.java
```diff
@@ -0,0 +1,94 @@
...
+ * To trigger manually only.
+ * To test it, set API_KEY and other values accordingly.
+ */
+@Slf4j
+class UnleashProviderIntegrationTest {
+
+    private static final String FLAG_NAME = "open-feature_flag1";
+    public static final String API_KEY = null;
+    public static final String API_URL = "https://app.unleash-hosted.com/demo/api/";
+    public static final String APP_NAME = "open-feature";
```
gastonfournier: I wonder if we can use docker containers for these tests. It should be easy both for dev and for CI: https://github.com/Unleash/unleash-server-api-go/blob/main/.github/workflows/build.yml#L30-L31

In particular, you can fix the admin token so it's known at runtime: https://github.com/Unleash/unleash-server-api-go/blob/fcd6e223b3a6c33a9e8394c5fbcf27f5995277e3/docker-compose.yml#L19-L31

gastonfournier (commented):
This looks promising! I leave you a few comments to make it better, but I'm glad you're doing this!

It might be worthy adding a comment in https://github.com/orgs/Unleash/discussions/3896 to engage other contributors.

liran2000:
> This looks promising! I leave you a few comments to make it better, but I'm glad you're doing this!
> 
> It might be worthy adding a comment in https://github.com/orgs/Unleash/discussions/3896 to engage other contributors.

Thanks @gastonfournier !
I made some adjustments based on your comments.
Since you probably more familiar with Unleash, I am interested in opinions on the concepts of this provider as described at the [Readme](https://github.com/open-feature/java-sdk-contrib/compare/main...liran2000:java-sdk-contrib:feature/unleash?expand=1#diff-33b37a3e9b9bd14c4d0853c582e47b468d3927d41da53b47900a8c60a90b1133):

## Concepts
* Boolean evaluation gets feature enabled status.
* String evaluation gets feature variant value.

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    @Override
+    public ProviderEvaluation<Boolean> getBooleanEvaluation(String key, Boolean defaultValue, EvaluationContext ctx) {
+        if (!ProviderState.READY.equals(state)) {
+            if (ProviderState.NOT_READY.equals(state)) {
+                throw new ProviderNotReadyError(PROVIDER_NOT_YET_INITIALIZED);
+            }
+            throw new GeneralError(UNKNOWN_ERROR);
+        }
```
ivarconr, line 109: It is an anti-pattern for Unleash SDKs to throw during flag evaluations. It should be safe to check the state of a flag, regardless of the SDK being ready. It leads to a bad coding practice where the user will need to catch runtime exception all over to avoid their application to crash. In most cases a user can rely on default-values and local backups until the SDK has synchronized. If you need your application to not handle load before this has happend you can block on the initialization phase or use the ready event to not make the application ready and added to the LB configuration. 

I am not sure if it is idiomatic to the Open Feature API to throw here, but then it should be taken care of the Core Logic of Open Feature, and not in the wrapper.

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
```diff
@@ -0,0 +1,171 @@
...
+        UnleashContext context = ctx == null ? UnleashContext.builder().build() : ContextTransformer.transform(ctx);
+        boolean featureBooleanValue = unleash.isEnabled(key, context, defaultValue);
+        return ProviderEvaluation.<Boolean>builder()
+            .value(featureBooleanValue)
+            .build();
+    }
+
+    @Override
+    public ProviderEvaluation<String> getStringEvaluation(String key, String defaultValue, EvaluationContext ctx) {
+        if (!ProviderState.READY.equals(state)) {
```
ivarconr, line 175: remove. Se comment above.

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
```diff
@@ -0,0 +1,171 @@
...
+        UnleashContext context = ctx == null ? UnleashContext.builder().build() : ContextTransformer.transform(ctx);
+        Variant evaluatedVariant = unleash.getVariant(key, context);
+        String variantName;
+        String value;
+        if (DISABLED_VARIANT.equals(evaluatedVariant)) {
+            variantName = null;
+            value = defaultValue;
+        } else {
+            variantName = evaluatedVariant.getName();
+            value = evaluatedVariant.getPayload().get().getValue();
```
ivarconr: value is optional in Unleash. Need to have a test to make sure this does not end up in Null-pointer exception.

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
```diff
@@ -0,0 +1,171 @@
...
+        }
+        return ProviderEvaluation.<String>builder()
+            .value(value)
+            .variant(variantName)
+            .build();
+    }
+
+    @Override
+    public ProviderEvaluation<Integer> getIntegerEvaluation(String key, Integer defaultValue, EvaluationContext ctx) {
+        throw new TypeMismatchError(NOT_IMPLEMENTED);
```
ivarconr: Payload of type "number" will come soon in Unleash.

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    @Override
+    public ProviderEvaluation<Double> getDoubleEvaluation(String key, Double defaultValue, EvaluationContext ctx) {
+        throw new TypeMismatchError(NOT_IMPLEMENTED);
```
ivarconr: Payload of type "number" will come soon in Unleash.

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    @Override
+    public ProviderEvaluation<Value> getObjectEvaluation(String s, Value value, EvaluationContext evaluationContext) {
+        throw new TypeMismatchError(NOT_IMPLEMENTED);
```
ivarconr: I do not understand why we are not returning the variant payload here? It is quite complex and support a lot of things (json, CSV, string, numbers, etc).

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
```diff
@@ -0,0 +1,171 @@
...
+import static io.getunleash.Variant.DISABLED_VARIANT;
+
+/**
+ * Provider implementation for Unleash.
+ */
+@Slf4j
+public class UnleashProvider extends EventProvider {
+
+    @Getter
+    private static final String NAME = "Unleash Provider";
```
ivarconr: should not the name simply be "Unleash"?

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
```diff
@@ -0,0 +1,171 @@
...
+    public static final String NOT_IMPLEMENTED =
+        "Not implemented - provider does not support this type. Only boolean is supported.";
```
ivarconr, line 36: This is not true. Unleash support `boolean`, `string`, `json` and `csv`. We also expect to add `number` as variant payload soon.

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
liran2000: To clarify, it is taken care of in the Core Logic of Open Feature by [returning a default value](https://github.com/open-feature/java-sdk/blob/main/src/main/java/dev/openfeature/sdk/OpenFeatureClient.java#L150) in this case when evaluating via the client, not expecting the user to catch runtime exception.
It is verified in [provider test](https://github.com/open-feature/java-sdk-contrib/pull/424/files#diff-4ac6957a0b62e15f9248ed31569e8a32d9a7e86a15486f69015b1a40d84ca1dcR180) and [core spec test](https://github.com/open-feature/java-sdk/blob/68f97c6c08a1223ad49e5358cadc3dab66a1d9ec/src/test/java/dev/openfeature/sdk/FlagEvaluationSpecTest.java#L101).

\* @toddbaert this is raised by several developers when implementing a provider, if you have some idea how to make it more clear that throwing exception leading to returning default value by the framework, can be taken separately from this PR.

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
liran2000: Ok, so this can be updated when a new Unleash version is released and adopted.

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
```diff
@@ -0,0 +1,171 @@
...
+    @Getter
+    private ProviderState state = ProviderState.NOT_READY;
+
+    private AtomicBoolean isInitialized = new AtomicBoolean(false);
+
+    /**
+     * Constructor.
+     * @param unleashOptions UnleashOptions
+     */
+    public UnleashProvider(UnleashOptions unleashOptions) {
```
ivarconr: should accept `UnleashConfig` and not a custom wrapper. This breaks the expectations of Unleash.

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
```diff
@@ -0,0 +1,171 @@
...
+     * @param evaluationContext evaluation context
+     * @throws Exception on error
+     */
+    @Override
+    public void initialize(EvaluationContext evaluationContext) throws Exception {
+        boolean initialized = isInitialized.getAndSet(true);
+        if (initialized) {
+            throw new GeneralError("already initialized");
+        }
+        super.initialize(evaluationContext);
```
ivarconr, line 73: this looks strange to me. What does this line do?

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
liran2000: Can you please clarify on boolean ? in Unleash payload, I don't see an option for boolean type, see screenshot.

What do you think about the following concepts, and if you have any suggestion for this ?
- Boolean evaluation gets feature enabled status.
- String evaluation gets feature variant value.

image

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
liran2000: UnleashOptions contains UnleashConfig. This can be beneficial if additional future parameters are needed. It is also consistent with other OpenFeature providers.

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
```diff
@@ -0,0 +1,171 @@
...
+        UnleashSubscriberWrapper unleashSubscriberWrapper = new UnleashSubscriberWrapper(
+            unleashOptions.getUnleashConfigBuilder().build().getSubscriber(), this);
+        unleashOptions.getUnleashConfigBuilder().subscriber(unleashSubscriberWrapper);
+        UnleashConfig unleashConfig = unleashOptions.getUnleashConfigBuilder().build();
+        unleash = new DefaultUnleash(unleashConfig);
+
+        // else, state will be changed via UnleashSubscriberWrapper events
+        if (unleashConfig.isSynchronousFetchOnInitialisation()) {
```
ivarconr: I do not see the need for this. Unleash is per definition ready after it is initialized. If you want  it to throw on init if you cannot connect to the Unleash API/Edge you control that via you configurations.

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashSubscriberWrapper.java
```diff
@@ -0,0 +1,122 @@
...
+import lombok.extern.slf4j.Slf4j;
+
+import javax.annotation.Nullable;
+
+/**
+ * UnleashSubscriber wrapper for emitting event provider events.
+ */
+@Slf4j
+@Generated
+public class UnleashSubscriberWrapper implements UnleashSubscriber {
```
ivarconr, line 27: I think this wrapper can be dropped entirely, see other comments around "ready" and best practices in Unleash.

providers/unleash/src/test/resources/features.json
```diff
@@ -0,0 +1,4509 @@
+{
+  "version": 1,
+  "features": [
+    {
+      "name": "083db41c-c6c8-427b-815b-f735c3c6da60",
```
ivarconr: this file looks very strange and is very long. It is usually better to a nicely crafted examples with descriptions on the purpose of each configuration examples and what we are testing for.

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
```diff
@@ -0,0 +1,171 @@
...
+
+    @Override
+    public ProviderEvaluation<Boolean> getBooleanEvaluation(String key, Boolean defaultValue, EvaluationContext ctx) {
+        if (!ProviderState.READY.equals(state)) {
+            if (ProviderState.NOT_READY.equals(state)) {
+                throw new ProviderNotReadyError(PROVIDER_NOT_YET_INITIALIZED);
+            }
+            throw new GeneralError(UNKNOWN_ERROR);
+        }
+        UnleashContext context = ctx == null ? UnleashContext.builder().build() : ContextTransformer.transform(ctx);
```
ivarconr, line 110: it would be easier if the `ContextTransformer` could handle a nullable context. How is the OpenFeature API for this?

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashSubscriberWrapper.java
liran2000: How can this wrapper be dropped, and meet the [spec requirement](https://github.com/open-feature/spec/blob/main/specification/sections/05-events.md#requirement-511), for example for emitting PROVIDER_CONFIGURATION_CHANGED event ? "The provider MAY define a mechanism for signaling the occurrence of one of a set of events, including PROVIDER_READY, PROVIDER_ERROR, PROVIDER_CONFIGURATION_CHANGED and PROVIDER_STALE, with a provider event details payload." ?

CC. @toddbaert

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/ContextTransformer.java
```diff
@@ -0,0 +1,75 @@
...
+        });
+        return unleashContextBuilder.build();
+    }
+
+    /**
+     * Transform UnleashContext to EvaluationContext.
+     * @param unleashContext the UnleashContext
+     * @return transformed EvaluationContext
+     */
+    public static EvaluationContext transform(UnleashContext unleashContext) {
```
ivarconr: I think this is only needed for the tests?

providers/unleash/src/test/java/dev/openfeature/contrib/providers/unleash/UnleashProviderTest.java
```diff
@@ -0,0 +1,272 @@
...
+        // ErrorCode.PROVIDER_NOT_READY should be returned when evaluated via the client
+        assertThrows(GeneralError.class, ()-> asyncInitUnleashProvider.getBooleanEvaluation("fail", false, new ImmutableContext()));
+        assertThrows(GeneralError.class, ()-> asyncInitUnleashProvider.getStringEvaluation("fail", "", new ImmutableContext()));
+
+        asyncInitUnleashProvider.shutdown();
+    }
+
+    @SneakyThrows
+    @Test
+    void contextTransformTest() {
```
ivarconr, line 242: I am not sure what this actually tests, as one of the method being tested is only used by this test.

providers/unleash/src/test/resources/features.json
liran2000: It is taken from [Unleash demo](https://github.com/Unleash/unleash-client-go/blob/88862d88f5813e6869f4790da689a1d492b587a6/demo_app_toggles.json#L5). I can make it shorter, yes.

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/ContextTransformer.java
```diff
@@ -0,0 +1,75 @@
...
+                    unleashContextBuilder.environment(String.valueOf(v));
+                    break;
+                case CONTEXT_REMOTE_ADDRESS:
+                    unleashContextBuilder.remoteAddress(String.valueOf(v));
+                    break;
+                case CONTEXT_SESSION_ID:
+                    unleashContextBuilder.sessionId(String.valueOf(v));
+                    break;
+                case CONTEXT_CURRENT_TIME:
+                    unleashContextBuilder.currentTime(ZonedDateTime.parse(String.valueOf(v)));
```
ivarconr, line 40: I would like to see some more tests around this one in particular. How are the user supposed to set the field?

providers/unleash/src/test/java/dev/openfeature/contrib/providers/unleash/UnleashProviderTest.java
```diff
@@ -0,0 +1,272 @@
...
+            unleashProvider.getDoubleEvaluation("test", 1.0, new ImmutableContext());
+        });
+        assertThrows(TypeMismatchError.class, () -> {
+            unleashProvider.getObjectEvaluation("test", new Value(), new ImmutableContext());
+        });
+    }
+
+    @SneakyThrows
+    @Test
+    void shouldThrowIfNotInitialized() {
```
ivarconr, line 211: I disagree with this test. Unleash should only throw as part of initialization if that is what the user wants.

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/ContextTransformer.java
liran2000: Not only for tests.
By [Usage example](https://github.com/open-feature/java-sdk-contrib/blob/8a6316705d8ac7779ad2f00db4b18710120de7d0/providers/unleash/README.md#usage-example):
```
UnleashContext unleashContext = UnleashContext.builder().userId("1").build();
EvaluationContext evaluationContext = EvaluationContext.transform(unleashContext);
featureEnabled = client.getBooleanValue(FLAG_NAME, false, evaluationContext);
```

What do you think ?

ivarconr: why would you use the Open Feture wrapper if you use the Unleash SDK? 

Imh the Open Feature API should use generics for context containers, but they don't.

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/ContextTransformer.java
liran2000: It is included and verified in [test](https://github.com/open-feature/java-sdk-contrib/blob/8a6316705d8ac7779ad2f00db4b18710120de7d0/providers/unleash/src/test/java/dev/openfeature/contrib/providers/unleash/UnleashProviderTest.java#L216).

providers/unleash/src/test/java/dev/openfeature/contrib/providers/unleash/UnleashProviderTest.java
liran2000: Please see previous [comment](https://github.com/open-feature/java-sdk-contrib/pull/424#discussion_r1321126698). End result for the user is not an exception.

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
ivarconr: I agree with your comment for "boolean value". 

In Unleash a toggle can either be enabled or disabled. In the case of being enabled it can also take a variant. The variant may take values in the form of payload, which can take different types. 

If it is disabled the user can specify the fallback variant to be used, which can also take a payload.

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
ivarconr: The challenge is that UnleashOptions is mutable and this can lead to a race condition. If the provider needs to expend with custom options (which I really do not think is the case) it should unwrap the immutable object and create a new configuration based on the the provided object. 

There is a reason the SDK does not accept UnleashOptions in the first place.

ivarconr:
Hi, thanks for doing this. Seems to make some good progress already. I left a bunch of comments. 

PS! There is also a "more options" in Unleash, which allows you to peek more in to the defintions and whether flags exists or not, that might be interesting for some of the unimplemented API methods for this Unleash  Open Feature Provider:

https://github.com/Unleash/unleash-client-java/blob/main/src/main/java/io/getunleash/DefaultUnleash.java#L307-L336

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
ivarconr: I think we probably should just return the entire variant object here. (PS! I have not investigated the "Value" type and what it supports"

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashSubscriberWrapper.java
ivarconr: I missed the `PROVIDER_CONFIGURATION_CHANGED` event. That makes sense I guess.

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
ivarconr: If this is important to Open Feature this should live in the core logic and not in the provider. Unleash does not work like this, and this logic changes the exceptions on how Unleash works.

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
liran2000: I think this is one of the main points in this provider. We can separate to boolean/string cases per my interpertation of the [spec](getStringEvaluation(String key, String defaultValue, EvaluationContext ctx)):

- _getStringEvaluation(String key, String defaultValue, EvaluationContext ctx)_
The method expects defaultValue, not the value of default variant, this is not exactly by what you described. How I see it in the implementation here, is to fit the generic case of evaluation by default value of this method.
- getBooleanEvaluation(String key, Boolean defaultValue, EvaluationContext ctx)
Since there is no explicit "boolean" payload type, I thought to return the "enabled" status of the flag here, and not the payload. This will act like "feature is enabled" flag.

What do you think of the concepts here ?
CC. @toddbaert

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
liran2000: I don't think it is changing the logic, please have a look at this [core spec test](https://github.com/open-feature/java-sdk/blob/68f97c6c08a1223ad49e5358cadc3dab66a1d9ec/src/test/java/dev/openfeature/sdk/FlagEvaluationSpecTest.java#L101):

```
assertEquals(ErrorCode.PROVIDER_NOT_READY, client.getBooleanDetails("return_error_when_not_initialized", false).getErrorCode());
```

The user expected to evaluate a flag via the client, not directly via the provider.
It means that if user evaluates a flag via the client, he gets a response with default value and a matching error code, and not an exception.

liran2000:
@ivarconr Appreciate your feedback !

I did see the "more options" like fetching all flags, but I don't think the OpenFeature provider currently have place for them.

Will address your comments.

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
```diff
@@ -0,0 +1,171 @@
...
+        Variant evaluatedVariant = unleash.getVariant(key, context);
+        String variantName;
+        String value;
+        if (DISABLED_VARIANT.equals(evaluatedVariant)) {
```
ivarconr, line 185: I am a bit unsure about this one. Here you are redefining the disabled variant in Unleash. 

Unleash itself has a concept of a "default" `Variant` you can pass and have that one returned by the SDK. This wrapper takes away the variant name and only keeps the value for the variant. Not sure if this is even possible to represent in Open Feature? It looks like a limitation in the API to only be able to define the value and not the name,

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
ivarconr: Yes, but I still believe this is cluttering the provider code and open for unnecessary bugs. I strongly recommend that this logic belongs to Open Feature Core and not the provider code.

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
liran2000: It is a conceptual question.
_getStringEvaluation(String key, String defaultValue, EvaluationContext ctx)_
The method expects defaultValue, not the value of default variant, this is not exactly by what you described. How I see it in the implementation here, is to fit the generic case of evaluation by default value of this method.

I am also not sure here, would like some opinions from OpenFeature point of view.
@toddbaert what do you think ?

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
liran2000: Ok, I hear you. Created core SDK [PR](https://github.com/open-feature/java-sdk/pull/603) for it, then can be removed from this provider.

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
toddbaert: Yes, reason is optional. If there's the provider has no API for understanding _why_ an evaluation proceeded the way it did, this can be omitted.

providers/unleash/src/test/java/dev/openfeature/contrib/providers/unleash/UnleashProviderIntegrationTest.java
toddbaert: Ya I would recommend a dockerized integration test, and (if possible) that you leverage the standard integration test suite defined in the spec: https://github.com/open-feature/spec/blob/main/specification/assets/gherkin/evaluation.feature

You can omit specific assertions (such as reason) as your provider requires, but this covers a lot of surface area, including context parsing, object parsing, type mapping and error handling.

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
toddbaert: @ivarconr I think you are misunderstanding a point here. An openfeature provider _is not an SDK_ itself. It is an _adapter_ that allows the OpenFeature SDK to talk to a provider certain backend (Unleash in this case). When an OpenFeature _provider_ throws like this _it IS caught by the SDK_ and the user never gets that error. This is basically just an example of the "throw early catch late" philosophy. This is the mechanism by which a provider communicates an error to the SDK. The error is only recorded internally for debugging and integration use-cases.

>Yes, but I still believe this is cluttering the provider code and open for unnecessary bugs. **_I strongly recommend that this logic belongs to Open Feature Core and not the provider code._**

This is exactly how it works. The OpenFeature core SDK code properly handles these errors and returns the default. It also will run any custom error handing code (if attached) transparently. For example, if you configure the [OpenTelemetry hook](https://openfeature.dev/ecosystem?instant_search%5Bquery%5D=OpenTelemetry&instant_search%5BrefinementList%5D%5Btype%5D%5B0%5D=Hook), it will send telemetry about such errors, even though the user-code doesn't see them.

See [here](https://openfeature.dev/docs/reference/concepts/provider#how-should-i-handle-error-conditions) and [here](https://openfeature.dev/specification/sections/providers#requirement-227) for more info.

You could hide all these errors in your provider, but then all sorts of integrations (like the OTel one mentioned above) wouldn't work. In both cases, users will never see the error, but if you throw this error in your provider, they can attach all kind of interesting error integrations.

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
toddbaert: That's what I would recommend.

providers/unleash/src/test/java/dev/openfeature/contrib/providers/unleash/UnleashProviderTest.java
toddbaert: Yep this is not an exception exposed to the user. It's only used for allowing custom error handing and reporting (via hooks), and telemetry, etc.

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
toddbaert: This is going to be the main challenge of this provider (though we do have a minority of other providers that have the same challenge and have done something similar to what you propose above). It's mostly a documentation issue. I think given that Unleash supports string/json/csv payloads in addition to a boolean enabled/disabled state, the best solution is to return the `enabled/disabled` state with boolean resolutions, and the `variant` with all other evaluations.

The other option could be use use [flagMetadata](https://openfeature.dev/specification/types#flag-metadata) to carry the enabled property, but I think that would be a poor user experience.

You could also use the constructor of your provider to allow some customization/fallback of this behavior, at the cost of some complexity.

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
toddbaert: The `Value` type supports almost anything - it can return an array like structure or an object like structure, or even boxed basic types.

I agree that this can be enhanced to return CSV and JSON payloads from Unleash as a Value containing Arrays and Maps, respectively.

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
toddbaert: Correct, OpenFeature does not support passing a string to identify the default variant at evaluation. It only supports passing the default value as a fallback.

I'm not sure if "overriding" the DISABLED variant is the best choice here. As far as I can tell you can just return what the unleash API gives you, unless there's some kind of error (in which case you can throw and the OpenFeature SDK will handle returning the default).

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
liran2000: 'Provider' suffix is consistent with other providers in this repository, although I would also think 'Unleash' is preferred. @toddbaert can we use 'Unleash', even it is inconsistent with other providers ?

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
liran2000: UnleashOptions is not Mutable.
It gets UnleashConfig.Builder, and building new UnleashConfig on initialization.
I think initialization is fine now:

```
UnleashConfig.Builder unleashConfigBuilder =
	UnleashConfig.builder()
		.unleashAPI(new URI(unleashAPI))
		.appName("fakeApp")
		.subscriber(testSubscriber)
		.synchronousFetchOnInitialisation(synchronousFetchOnInitialisation);

UnleashOptions unleashOptions = UnleashOptions.builder()
	.unleashConfigBuilder(unleashConfigBuilder)
	.build();
return new UnleashProvider(unleashOptions);
```

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
liran2000: A practice to call super in case there will be some implementation.

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
liran2000: **This is an important decision here.**

```
unleashProvider.getStringEvaluation("non-existing", "fallback_str", context);
```

What is the expected value here ?

- If we just return what the unleash API gives you, we will get **_null_**, as it returns DISABLED variant with a payload with null value.
- If we use implementation which checks for the DISABLED variant and uses default value, we will get **_fallback_str_**.

If first option is chosen, user can use default value with expecting it to be used in case of some issue like non-existing flag, but will actually get _null_.

An possible solution here is to make the implementation configurable on the provider options, returnDefaultValueOnDisabledVariant = true/false. In this case, what would be the default value ?

@ivarconr @toddbaert what do you think ?

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
toddbaert: It's really just an identifier. I wish we'd used `flagd` instead of `flagd provider` in many cases. We wouldn't expect people to build logic around this string, it's mostly for logging/telemetry.

I'd do whatever you prefer.

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
toddbaert: Ah, I didn't realize the return value would be `null` in that case.

Then I think this implementation is the most idiomatic: 

> If we use implementation which checks for the DISABLED variant and uses default value, we will get fallback_str

You could also throw a FlagNotFoundError in this case, in which case the SDK would default. That would be more expensive because of the error serialization, but for this provider you might consider trying to get the variant in this DISABLED state an exceptional circumstance.

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
ivarconr: It's already on main, and on the demo instance.
image

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
ivarconr: > provider throws like this it IS caught by the SDK and the user never gets that error. This is basically just an example of the "throw early catch late" philosophy. This is the mechanism by which a provider communicates an error to the SDK.

I am really glad Open Feature catches this, and not surface this to the developers. Generally I would be careful on using **exception based control flow**. This tend to scream at me, as I have seen how they play out in large code bases over time. 

> This is the mechanism by which a provider communicates an error to the SDK. The error is only recorded internally for debugging and integration use-cases.

This error is **not theoretically possible** in the Unleash Java SDK, after it has been instantiated. The Unleash Java SDK guarantees that you will get a response from the "isEnabled" call and the "getVariant" call. 

Its up to the developer to decide if he/she want to wait until synchronized (and/or throw) during the initialization phase of Unleash. This is controlled via the Unleash configuration. 

> You could hide all these errors in your provider, but then all sorts of integrations (like the OTel one mentioned above) wouldn't work.

I am not talking about hiding errors, but the implemented error logic here is not relevant for the Unleash Java SDK. 

I also think it needs to be in the concern of OpenFeature core logic to check whether the Provider has been initialized or not as the control logic for that happens outside of the provider code.

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
ivarconr: I think we should return the entire `Variant` object from Unleash as it contain more details than just the payload.

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
ivarconr: This me trying to understand the life-cycle. Why would initialization need "evaluationContext"?

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
ivarconr: Ok, I see your point. It can be confusing to maintain `UnleashConfig` and `UnleashOptions` and explain the difference. 

Maybe we can simply rename to "UnleashProviderConfig" or UnleashProviderOptions"?

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
```diff
@@ -0,0 +1,173 @@
...
+        UnleashContext context = ctx == null ? UnleashContext.builder().build() : ContextTransformer.transform(ctx);
+        Variant evaluatedVariant = unleash.getVariant(key, context);
+        String variantName;
+        Value value;
+        if (DISABLED_VARIANT.equals(evaluatedVariant)) {
+            variantName = null;
+            value = defaultValue;
+        } else {
+            variantName = evaluatedVariant.getName();
+            value = evaluatedVariant.getPayload().map(p -> new Value(p.getValue())).orElse(null);
```
ivarconr, line 190: I think we need to rethink the value part a bit. By just mapping out the payload value we loose a few important details
* payload type (make it possible to parse the value)
* Is the feature enabled

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
ivarconr: When the user actually provide a defaultValue I think the current implementaion is ok for now, as long as we document that the variantName will not be set when the defaultValue is returned. 


But when the use **does not** provide a "defaultValue" it would be better to simply use the "DISBALED_VARIANT" from Unleash.

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
toddbaert: That's understandable.

I was making general comments on the behavior of the system when providers throw. I agree that we don't want exception based control flow, generally. I was just responding to your comments about exceptions reaching users, not this particular exception. One of the core idioms of the project is to never throw during flag evaluations.

> I also think it needs to be in the concern of OpenFeature core logic to check whether the Provider has been initialized or not as the control logic for that happens outside of the provider code.

The SDK does this, but if the initialization involves some IO, it is possible in some circumstances for users to run an evaluation before the SDK has completely initialized the provider. The user would have to be avoiding a few guardrails in that case. The SDK has numerous methods for awaiting the initialization.

If it's impossible for this to be a problem in the unleash provider (because the underlying SDK is ready when instantiated), then we can probably just set the provider to ready immediately.

liran2000: I prefer to leave the check, as it is possible on incorrect usage:

```
UnleashProvider unleashProvider = new UnleashProvider(unleashOptions);
unleashProvider.getBooleanEvaluation... // -> this should get to the NOT_READY error, as unleash not initialized yet.
OpenFeatureAPI.getInstance().setProviderAndWait("unleash", unleashProvider);
```

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
liran2000: Bug...
image

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/ContextTransformer.java
sighphyre: Oh that's interesting, I think I have the same worries as ivarconr - that's not really about this implementation but about the way OpenFeature is intended to be used. If I have to put logic specific to my vendor SDK in the code, that sounds like it would be tough to migrate to another provider, since I'd have to find all these cases and map them to a new provider's content implementation. Totally might have missed something obvious here, since I don't know OpenFeature very well (yet!)

providers/unleash/src/test/java/dev/openfeature/contrib/providers/unleash/UnleashProviderTest.java
```diff
@@ -0,0 +1,293 @@
...
+                .willReturn(
+                    aResponse()
+                        .withStatus(200)
+                        .withHeader("Content-Type", "application/json")
+                        .withBody(backupFileContent)));
+        stubFor(post(urlEqualTo("/api/client/register")).willReturn(aResponse().withStatus(200)));
+    }
+
+    @SneakyThrows
+    private UnleashProvider buildUnleashProvider(boolean synchronousFetchOnInitialisation, String unleashAPI, String backupFileContent, TestSubscriber testSubscriber) {
```
sighphyre: I don't think `backupFileContent` is ever used here. Doesn't look like it affects the tests but it might be worth taking out

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    @Override
+    public void shutdown() {
+        super.shutdown();
+        log.info("shutdown");
+        if (unleash != null) {
+            unleash.shutdown();
+        }
+        state = ProviderState.NOT_READY;
```
sighphyre, line 212: Does OpenFeature ever expect NOT_READY providers to reenter a READY state? If so, it might be worth allowing the GC to reclaim the unleash object, we definitely don't expect to be able to re initialize these, generally we make a new one

providers/unleash/src/test/java/dev/openfeature/contrib/providers/unleash/UnleashProviderTest.java
liran2000: it is needed, similar to tests at Unleash.

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
liran2000: It is not expected to use the provider for evaluations after it is shutdown.

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
liran2000: It is by the [spec requirement](https://github.com/open-feature/spec/blob/main/specification/sections/02-providers.md#requirement-241) to receive that when initialized method is implemented.

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/ContextTransformer.java
liran2000: This is a valid point. How I see it, evaluation context, like _USER_ID_ is something for a specific provider service, which not always used or defined by this key. Then context can be for attributes specific for this provider. Since such context properties are not defined at _OpenFeature_, then I think this part is sort of provider specific. As for using _UnleashContext_, it is not mandatory, user can set attributes map keys manually, but I think as it sort of custom anyway, it can be simpler for the user.

CC @toddbaert

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
liran2000: Added number support

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
liran2000: I added some details to evaluation via [flag metadata](https://github.com/open-feature/java-sdk-contrib/compare/main...liran2000:java-sdk-contrib:feature/unleash?expand=1#diff-33b37a3e9b9bd14c4d0853c582e47b468d3927d41da53b47900a8c60a90b1133R47):

>Additional evaluation data can be received via flag metadata, such as:
  * *enabled* - boolean
  * *variant-stickiness* - string
  * *payload-type* - string, optional

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
liran2000: I added some details to evaluation via [flag metadata](https://github.com/open-feature/java-sdk-contrib/compare/main...liran2000:java-sdk-contrib:feature/unleash?expand=1#diff-33b37a3e9b9bd14c4d0853c582e47b468d3927d41da53b47900a8c60a90b1133R47):

Additional evaluation data can be received via flag metadata, such as:

enabled - boolean
variant-stickiness - string
payload-type - string, optional

liran2000:
hi @ivarconr what do you think about proceeding with this PR and adding Unleash provider here ?
basically most of the comments are addressed.

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
toddbaert: Ya, that would be odd API usage (not using the OpenFeature evaluation API) but I guess you're right.

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/ContextTransformer.java
toddbaert: > why would you use the Open Feture wrapper if you use the Unleash SDK?
> 
> Imh the Open Feature API should use generics for context containers, but they don't.

> Oh that's interesting, I think I have the same worries as ivarconr - that's not really about this implementation but about the way OpenFeature is intended to be used. If I have to put logic specific to my vendor SDK in the code, that sounds like it would be tough to migrate to another provider, since I'd have to find all these cases and map them to a new provider's content implementation. Totally might have missed something obvious here, since I don't know OpenFeature very well (yet!)

Sorry I missed this until now!

@sighphyre and @ivarconr are correct about this. It's an anti-pattern to expect users to interact with an `UnleashContext` directly if they're using open feature.

If the `userId` is a special field that consistently identifies a user in Unleash, it should be mapped from the [targetingKey](https://openfeature.dev/specification/sections/evaluation-context#requirement-311). This is the only rigorously defined entry in the `EvaluationContext`. The transformation should be done internally. If you want to give users some control over it, you could expose a transformer function in your provider constructor to allow customization of this behavior.

If your provider _needs_ the targetingKey set, we have exceptions that indicate this specific scenario.

liran2000: hi @toddbaert,
Unleash context contains several optional parameters, user ID is one of them.
Then instead of:
```
UnleashContext unleashContext = UnleashContext.builder()
	.userId(userIdValue)
	.currentTime(currentTimeValue)
	.sessionId(sessionIdValue)
	.remoteAddress(remoteAddressValue)
	.environment(environmentValue)
	.appName(appNameValue)
	.addProperty(customPropertyKey, customPropertyValue)
	.build();
EvaluationContext evaluationContext = ContextTransformer.transform(unleashContext);
```
Then I can replace the usage by exposing context keys constants, with something like:
```
MutableContext evaluationContext = new MutableContext();
evaluationContext.add(CONTEXT_APP_NAME, appName);
evaluationContext.add(CONTEXT_USER_ID, userId);
evaluationContext.add(CONTEXT_ENVIRONMENT, env);
evaluationContext.add(CONTEXT_SESSION_ID, sessionId);
evaluationContext.add(CONTEXT_REMOTE_ADDRESS, remoteAddress);
evaluationContext.add(CONTEXT_CURRENT_TIME, currentTime);
```
This looks like a better usage, is this what you mean ?
Parameters are optional, so possibly only some are enough for most cases.
It is also will be consistent with some other references:
- [flagd provider](https://github.com/liran2000/java-sdk-contrib/blob/2c1060ca8078329a344c45409d23c44d1b9db074/providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/FlagdProviderTest.java#L349)
- [go feature flag provider](https://github.com/liran2000/java-sdk-contrib/blob/81fc4122cf22ed935790333b0559a8181a5a769e/providers/go-feature-flag/src/test/java/dev/openfeature/contrib/providers/gofeatureflag/GoFeatureFlagProviderTest.java#L109)

As for whether user ID should be mapped to targeting key - can be either parameter like others or targeting key. I think it is not mandatory, then regular context parameter is better fit than targeting key here, but I am not sure. @ivarconr ?

toddbaert: _if_ there is a dedicated field in Unleash that denotes the user of the flag evaluation, that should be what's mapped from `targetingKey`, especially if it's used to maintain stickiness of random assignments, etc. If there's no such value, then there's no need for this mapping.

For the other values, I would recommend you _don't_ use exported constants, because that again creates a dependency in all flag-related application code directly on your provider. Instead, I would perhaps recommend a mapping function (a [Consumer](https://docs.oracle.com/javase/8/docs/api/java/util/function/Consumer.html) perhaps) that can be supplied when you construct the provider. It would take an `EvaluationContext` and return a `UnleashedContext`. Then consumers could map the values exactly as they want by supplying this param. You could have a default that just maps them using the same keys (`currentTimeValue`, `sessionIdValue`, etc).

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
sighphyre: Hey @liran2000, that bug looks super ugly, The dev who worked on this is off today, so I'll check tomorrow but we can't reproduce this. Was this still an issue last you checked?

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/ContextTransformer.java
ivarconr: > As for whether user ID should be mapped to targeting key - can be either parameter like others or targeting key. I think it is not mandatory, then regular context parameter is better fit than targeting key here, but I am not sure. @ivarconr ?

Correct. All fields in the unleash context are optional, but a few might be set for you if you do not set them yourself (environment and appName are examples). In fact a user can fully remove "userId" as a possible target in the Unleash API.

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
ivarconr: you make a good point @liran2000. This should not be allowed, and might be caused by my confusion of not fully understanding the Open Feature API. In my understanding using the  `unleashProvider` directly would be an anti-pattern and not expected, or even allowed.

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
liran2000: Not reproduced for me anymore.

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/ContextTransformer.java
liran2000: I updated the code and usage according to the comments.

ivarconr:
> hi @ivarconr what do you think about proceeding with this PR and adding Unleash provider here ?
basically most of the comments are addressed.

All in favor. Would love to get to test it in an example app, either as a alpha release or if you provide snapshot versions.

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
toddbaert: It's not specified if implementations should allow providers to re-enter NOT_READY/READY, so at this point it's up to the implementation.

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashProvider.java
toddbaert: Thanks for adding this to the readme!

providers/unleash/README.md
```diff
@@ -0,0 +1,62 @@
...
+* Boolean evaluation gets feature enabled status.
+* String evaluation gets feature variant value.
+
+## Usage
+Unleash OpenFeature Provider is using Unleash Java SDK.
+
+### Usage Example
+
+```
+FeatureProvider featureProvider = new UnleashProvider(unleashProviderConfig);
```
ivarconr:
```suggestion
FeatureProvider unleashProvider = new UnleashProvider(unleashProviderConfig);
```

Typo in the example here!

.github/component_owners.yml
```diff
@@ -19,6 +19,8 @@ components:
     - thomaspoignant
   providers/jsonlogic-eval-provider:
     - justinabrahms
+  providers/unleash:
+    - liran2000
```
toddbaert, line 23: @sighphyre @gastonfournier @ivarconr 

Would any of you like to be set as "component owners" here in addition to @liran2000 ? This will add you as assignee/reviewers automatically if changes are made to files under this path in a PR.

We require only one name here, but I thought I'd point it out.

providers/unleash/src/test/java/dev/openfeature/contrib/providers/unleash/UnleashProviderTest.java
toddbaert: I'm also confused about this :sweat_smile:

toddbaert: I pulled the branch and removed this param, and everything seems to work the same.

Can you clarify @liran2000 ? I'd like the tests to be as understandable as possible!

providers/unleash/src/test/java/dev/openfeature/contrib/providers/unleash/UnleashProviderTest.java
toddbaert: Looks like this test is testing a "real" method used at runtime now.

providers/unleash/src/test/java/dev/openfeature/contrib/providers/unleash/UnleashProviderIntegrationTest.java
toddbaert: I don't think this is as important with the way @liran2000 has done it. Mocking HTTP responses seems like a pretty good solution, as long as the mocked response is accurate these tests should be pretty solid.

toddbaert:
@liran2000 the release setup looks good. When we merge this and the associated release PR, it will release a `0.0.2` version which is fine with me. We don't publish snapshots (in the java sense) from this repo, but we could (I believe) make the version number something like `0.0.1-alpha` or `0.0.1-prerelease` if you'd like to have something that seems less "final" for now, which might be a good idea. cc @ivarconr 

@liran2000 Thanks for all your efforts on this. It's been requested by the community more than once!

Please see the 2 unresolved issues above.

.github/component_owners.yml
sighphyre: I'm very open to being added! I spend a lot of time on the Unleash SDKs so it'd be great to have an eye on this too!

liran2000: Added

providers/unleash/src/test/java/dev/openfeature/contrib/providers/unleash/UnleashProviderTest.java
liran2000: Removed

liran2000:
Thanks :)
addressed latest comments.

toddbaert (approved):
I'll merge this before EOD unless I hear objections.

Thanks again all, especially @liran2000

PR #434, opened 2023-09-12
feat: json logic operators for flagd in-process provider

### LINKED ISSUE #411: [flagd] Intial version of in-process provider

Kavindu-Dodan:
## This PR

Fixes https://github.com/open-feature/java-sdk-contrib/issues/411

Introduces and binds the following operators,

- Fractional
- SemVer
- StringComp 

Note that, implementations use following libraries 

- [semver4j](https://github.com/semver4j/semver4j) : for semantic version parsing and validations
- Murmur3 : ~from [Apache Hive ](https://github.com/apache/hive/blob/master/storage-api/src/java/org/apache/hive/common/util/Murmur3.java) extracted with license intact~ from apache common

~Note that, `Fractional` evaluations are not 100% compatible with flagd implementation from GO. This is because of the usage of signed long values and related overflowing happening in respective implementations of Java.~

Fractional evaluations are cross-language compatible thanks to murmur3 32 bit usage across languages.


providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/model/FlagParser.java
```diff
@@ -42,15 +40,18 @@ public class FlagParser {
     private static JsonSchema SCHEMA_VALIDATOR;
 
     static {
-        try {
-            final URL url = FlagParser.class.getClassLoader().getResource(SCHEMA_RESOURCE);
```
Kavindu-Dodan, line 46: This was an issue identified while working with a local build library. Resource loader urls which are relative to jar contains exclamation mark, preventing file loading. This PR contains the fix.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/targeting/Fractional.java
```diff
@@ -0,0 +1,127 @@
...
+
+    public Object evaluate(List arguments, Object data) throws JsonLogicEvaluationException {
+        if (arguments.size() < 2) {
+            return null;
+        }
+
+        final Operator.FlagProperties properties = new Operator.FlagProperties(data);
+
+        // check optional string target in first arg
+        Object arg1 = arguments.get(0);
```
toddbaert, line 31: I don't think we want to support this anymore. This should no longer be a string, and instead be an expression, see here: https://flagd.dev/concepts/#fractional-evaluation

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/resolver/process/targeting/FractionalTest.java
```diff
@@ -0,0 +1,220 @@
...
+
+        final List<Object> bucket2 = new ArrayList<>();
+        bucket2.add("green");
+        bucket2.add(50);
+
+        rule.add(bucket1);
+        rule.add(bucket2);
+
+        Map<String, String> data = new HashMap<>();
+        data.put(FLAG_KEY, "flagA");
```
toddbaert, line 52: I'm confused by this test, it doesn't set a value for `bucketKey` in data... I would expect to see that have different values, and see differential results when it does.

toddbaert:
> Note that, Fractional evaluations are not 100% compatible with flagd implementation from GO. This is because of the usage of signed long values and related overflowing happening in respective implementations of Java.

Unless the difference is _highly nuanced_ and unlikely to be seen in normal usage, I think this is a deal-breaker. We _need_ the evaluators to resolve identically, or it could result in weird behavior if applications have different components written in different languages. If we need to change something about the algorithm (convert numbers to UTF-8 strings in a consistent way, for example) that's fine - we can make that update.

Can you provide details as to exactly what is inconsistent? For example, could I make the [gherkin suite](https://github.com/open-feature/flagd-testbed/blob/main/gherkin/flagd-json-evaluator.feature) pass?

toddbaert:
FYI I see the same failure locally as what appears to be happening in the CI: https://github.com/open-feature/java-sdk-contrib/actions/runs/6165511503/job/16733420790?pr=434#step:6:3109

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/targeting/Fractional.java
toddbaert: Perhaps part of the problem is that this go sample here is out of date: https://github.com/open-feature/flagd/blob/main/docs/other_resources/in-process-providers/evaluators/fractional_evaluation.md. I'm sorry I missed that.

:disappointed: 

Maybe we should remove this sample and just link the real go code in this case. I think the rest of the doc is up to date.

EDIT:

Actually... it may not be out of date. I think by the time this code runs, `{ "var": "someValue" }` will be replaced WITH a string: https://github.com/open-feature/flagd/blob/main/core/pkg/eval/fractional_evaluation.go#L55. If the Java stuff works similarly, you might not have to do too much here... not sure.

Kavindu-Dodan: In case there is an expression, the value we receive at index 0 (argument 1) is already parsed by the JsonLogic library. This is done at `PreEvaluatedArgumentsExpression` interface.

Kavindu-Dodan: And this is identical to flagd implementation - https://github.com/open-feature/flagd/blob/main/core/pkg/eval/fractional_evaluation.go#L55-L63

Kavindu-Dodan: >  I think by the time this code runs, { "var": "someValue" } will be replaced WITH a string

Exactly this .!! 😃

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/resolver/process/targeting/FractionalTest.java
Kavindu-Dodan: This is related to this comment - https://github.com/open-feature/java-sdk-contrib/pull/434#discussion_r1324915298

By the time our implementation receives arguments, JsonLogic expressions will be resolved to their values. The test here only validate the rule it receive (bucketing key set to `bucketKey` and flag key set to `flagA`).

Hope this helps to clarify your doubt ? 

Note - I think I need to add more tests to cover targeting rules as a complete unit (or component) tests

Kavindu-Dodan:
> FYI I see the same failure locally as what appears to be happening in the CI: https://github.com/open-feature/java-sdk-contrib/actions/runs/6165511503/job/16733420790?pr=434#step:6:3109

this is resolved locally. I will push the changes with review changes :)

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/targeting/Fractional.java
toddbaert: Ah OK - so basically, your unit tests CAN'T actually test this - if I understand correctly, because they all assume that this replace will already have happened?

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/resolver/process/targeting/FractionalTest.java
toddbaert: > Note - I think I need to add more tests to cover targeting rules as a complete unit (or component) tests

Honestly, no, I think if we're on the same page, I will just implement the gherkin suite which should test those things.

toddbaert: I think, if anything, you'd want to run this test twice... once with "bucketKeyA" and once with "bucketKeyB" or some other value that resolves in a different random assignment. Does that make sense?

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/targeting/Fractional.java
Kavindu-Dodan: Yes exactly, we receive replaced values. As you observed in flagd [^1], we cannot differentiate whether this value comes through an expression or is replaced by JsonLogic. And the unit tests here only focus on the logical flow proposed in evaluator documentation [^2].

Anyway, good that we discussed this point so it's clear for everyone :) 

[^1]: https://github.com/open-feature/flagd/issues/871 
[^2]: https://github.com/open-feature/flagd/blob/main/docs/other_resources/in-process-providers/evaluators/fractional_evaluation.md

toddbaert: Thanks!

toddbaert:
This looks good to me, pending a better understanding of [this point](https://github.com/open-feature/java-sdk-contrib/pull/434#issuecomment-1718126928) and the additional test I mentioned. If the differences in the fractional evaluation are slim or can easily be avoided or fixed later, I can approve.

Kavindu-Dodan:
> > Note that, Fractional evaluations are not 100% compatible with flagd implementation from GO. This is because of the usage of signed long values and related overflowing happening in respective implementations of Java.
> 
> Unless the difference is _highly nuanced_ and unlikely to be seen in normal usage, I think this is a deal-breaker. We _need_ the evaluators to resolve identically, or it could result in weird behavior if applications have different components written in different languages. If we need to change something about the algorithm (convert numbers to UTF-8 strings in a consistent way, for example) that's fine - we can make that update.
> 
> Can you provide details as to exactly what is inconsistent? For example, could I make the [gherkin suite](https://github.com/open-feature/flagd-testbed/blob/main/gherkin/flagd-json-evaluator.feature) pass?

Sorry for the late response.

I did some tests with Java and Go but I was unable to get same hash generated from both. To start with, flagd use [twmb/murmur3](https://github.com/twmb/murmur3) and it uses unsigned integers [^1]. I copied our usage from apache hive [^2], and this uses Java long, which are signed values. Given that go uses UTF-8 by default and I mandate encoding in Java to UTF-8, this shouldn't be the root cause. However, I assume there's something with signed vs unsigned number usage (unfortunately I am not an expert on hashing ;) )

I will try with Guava [^3] and see if we get same results in Java. 

[^1]: https://github.com/twmb/murmur3/blob/master/murmur64.go#L45-L48
[^2]: https://github.com/apache/hive/blob/master/storage-api/src/test/org/apache/hive/common/util/TestMurmur3.java
[^3]: https://github.com/google/guava/blob/master/guava/src/com/google/common/hash/Murmur3_32HashFunction.java

toddbaert:
> > > Note that, Fractional evaluations are not 100% compatible with flagd implementation from GO. This is because of the usage of signed long values and related overflowing happening in respective implementations of Java.
> > 
> > 
> > Unless the difference is _highly nuanced_ and unlikely to be seen in normal usage, I think this is a deal-breaker. We _need_ the evaluators to resolve identically, or it could result in weird behavior if applications have different components written in different languages. If we need to change something about the algorithm (convert numbers to UTF-8 strings in a consistent way, for example) that's fine - we can make that update.
> > Can you provide details as to exactly what is inconsistent? For example, could I make the [gherkin suite](https://github.com/open-feature/flagd-testbed/blob/main/gherkin/flagd-json-evaluator.feature) pass?
> 
> Sorry for the late response.
> 
> I did some tests with Java and Go but I was unable to get same hash generated from both. To start with, flagd use [twmb/murmur3](https://github.com/twmb/murmur3) and it uses unsigned integers [1](#user-content-fn-1-8aef6b6dd4b61031a51b1170e3393df9). I copied our usage from apache hive [2](#user-content-fn-2-8aef6b6dd4b61031a51b1170e3393df9), and this uses Java long, which are signed values. Given that go uses UTF-8 by default and I mandate encoding in Java to UTF-8, this shouldn't be the root cause. However, I assume there's something with signed vs unsigned number usage (unfortunately I am not an expert on hashing ;) )
> 
> I will try with Guava [3](#user-content-fn-3-8aef6b6dd4b61031a51b1170e3393df9) and see if we get same results in Java.
> 
> ## Footnotes
> 1. https://github.com/twmb/murmur3/blob/master/murmur64.go#L45-L48 [↩](#user-content-fnref-1-8aef6b6dd4b61031a51b1170e3393df9)
> 2. https://github.com/apache/hive/blob/master/storage-api/src/test/org/apache/hive/common/util/TestMurmur3.java [↩](#user-content-fnref-2-8aef6b6dd4b61031a51b1170e3393df9)
> 3. https://github.com/google/guava/blob/master/guava/src/com/google/common/hash/Murmur3_32HashFunction.java [↩](#user-content-fnref-3-8aef6b6dd4b61031a51b1170e3393df9)

I could be misunderstanding your point, but the number representations the implementation of the hash algo uses should be irrelevant. ie: SHA1 implemented in any language with any number representation should be deterministic for any input. My suspicion is that our input bytes are actually not the same in both implementations.

Kavindu-Dodan:
> > > > Note that, Fractional evaluations are not 100% compatible with flagd implementation from GO. This is because of the usage of signed long values and related overflowing happening in respective implementations of Java.
> > > 
> > > 
> > > Unless the difference is _highly nuanced_ and unlikely to be seen in normal usage, I think this is a deal-breaker. We _need_ the evaluators to resolve identically, or it could result in weird behavior if applications have different components written in different languages. If we need to change something about the algorithm (convert numbers to UTF-8 strings in a consistent way, for example) that's fine - we can make that update.
> > > Can you provide details as to exactly what is inconsistent? For example, could I make the [gherkin suite](https://github.com/open-feature/flagd-testbed/blob/main/gherkin/flagd-json-evaluator.feature) pass?
> > 
> > 
> > Sorry for the late response.
> > I did some tests with Java and Go but I was unable to get same hash generated from both. To start with, flagd use [twmb/murmur3](https://github.com/twmb/murmur3) and it uses unsigned integers [1](#user-content-fn-1-8aef6b6dd4b61031a51b1170e3393df9). I copied our usage from apache hive [2](#user-content-fn-2-8aef6b6dd4b61031a51b1170e3393df9), and this uses Java long, which are signed values. Given that go uses UTF-8 by default and I mandate encoding in Java to UTF-8, this shouldn't be the root cause. However, I assume there's something with signed vs unsigned number usage (unfortunately I am not an expert on hashing ;) )
> > I will try with Guava [3](#user-content-fn-3-8aef6b6dd4b61031a51b1170e3393df9) and see if we get same results in Java.
> > ## Footnotes
> > 
> > 1. https://github.com/twmb/murmur3/blob/master/murmur64.go#L45-L48 [↩](#user-content-fnref-1-8aef6b6dd4b61031a51b1170e3393df9)
> > 2. https://github.com/apache/hive/blob/master/storage-api/src/test/org/apache/hive/common/util/TestMurmur3.java [↩](#user-content-fnref-2-8aef6b6dd4b61031a51b1170e3393df9)
> > 3. https://github.com/google/guava/blob/master/guava/src/com/google/common/hash/Murmur3_32HashFunction.java [↩](#user-content-fnref-3-8aef6b6dd4b61031a51b1170e3393df9)
> 
> I could be misunderstanding your point, but the number representations the implementation of the hash algo uses should be irrelevant. ie: SHA1 implemented in any language with any number representation should be deterministic for any input. My suspicion is that our input bytes are actually not the same in both implementations.

Finally found the reason and the fix for this mismatch.

flagd originally used 64-bit version of the murmur3 algorithm. This seems to not come from the original algorithm. I have swithched the algorithm to the 32-bit version [^1] and this matches perfectly with the Java version as well. 

I have updated this PR with the change -https://github.com/open-feature/java-sdk-contrib/pull/434/commits/2386d16c42a670f2f1a83618fe6e19a2d12776ab 

[^1]: https://github.com/open-feature/flagd/pull/913

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/resolver/process/targeting/FractionalTest.java
Kavindu-Dodan: this is done :) and there's variation validation thanks to this.

toddbaert:
> > > > > Note that, Fractional evaluations are not 100% compatible with flagd implementation from GO. This is because of the usage of signed long values and related overflowing happening in respective implementations of Java.
> > > > 
> > > > 
> > > > Unless the difference is _highly nuanced_ and unlikely to be seen in normal usage, I think this is a deal-breaker. We _need_ the evaluators to resolve identically, or it could result in weird behavior if applications have different components written in different languages. If we need to change something about the algorithm (convert numbers to UTF-8 strings in a consistent way, for example) that's fine - we can make that update.
> > > > Can you provide details as to exactly what is inconsistent? For example, could I make the [gherkin suite](https://github.com/open-feature/flagd-testbed/blob/main/gherkin/flagd-json-evaluator.feature) pass?
> > > 
> > > 
> > > Sorry for the late response.
> > > I did some tests with Java and Go but I was unable to get same hash generated from both. To start with, flagd use [twmb/murmur3](https://github.com/twmb/murmur3) and it uses unsigned integers [1](#user-content-fn-1-8aef6b6dd4b61031a51b1170e3393df9). I copied our usage from apache hive [2](#user-content-fn-2-8aef6b6dd4b61031a51b1170e3393df9), and this uses Java long, which are signed values. Given that go uses UTF-8 by default and I mandate encoding in Java to UTF-8, this shouldn't be the root cause. However, I assume there's something with signed vs unsigned number usage (unfortunately I am not an expert on hashing ;) )
> > > I will try with Guava [3](#user-content-fn-3-8aef6b6dd4b61031a51b1170e3393df9) and see if we get same results in Java.
> > > ## Footnotes
> > > 
> > > 1. https://github.com/twmb/murmur3/blob/master/murmur64.go#L45-L48 [↩](#user-content-fnref-1-8aef6b6dd4b61031a51b1170e3393df9)
> > > 2. https://github.com/apache/hive/blob/master/storage-api/src/test/org/apache/hive/common/util/TestMurmur3.java [↩](#user-content-fnref-2-8aef6b6dd4b61031a51b1170e3393df9)
> > > 3. https://github.com/google/guava/blob/master/guava/src/com/google/common/hash/Murmur3_32HashFunction.java [↩](#user-content-fnref-3-8aef6b6dd4b61031a51b1170e3393df9)
> > 
> > 
> > I could be misunderstanding your point, but the number representations the implementation of the hash algo uses should be irrelevant. ie: SHA1 implemented in any language with any number representation should be deterministic for any input. My suspicion is that our input bytes are actually not the same in both implementations.
> 
> Finally found the reason and the fix for this mismatch.
> 
> flagd originally used 64-bit version of the murmur3 algorithm. This seems to not come from the original algorithm. I have swithched the algorithm to the 32-bit version [1](#user-content-fn-1-00f9f1be7339fb899eb6503c349a1600) and this matches perfectly with the Java version as well.
> 
> I have updated this PR with the change -[2386d16](https://github.com/open-feature/java-sdk-contrib/commit/2386d16c42a670f2f1a83618fe6e19a2d12776ab)
> 
> ## Footnotes
> 1. https://github.com/open-feature/flagd/pull/913 [↩](#user-content-fnref-1-00f9f1be7339fb899eb6503c349a1600)

Great job. That makes sense. The murmur3 spec doesn't even include a 64bit version, and the 128bit is unstable across architectures. 32 is the right choice!

toddbaert (approved):

PR #436, opened 2023-09-15
fix: blocking in-process init, e2e tests

toddbaert:
* adds blocking init to flagd-in-process (made this common)
* throws in provider instead of returning error evals
* minor change to enum naming (`RPC` instead of `FLAGD`)
* adds `evaluation.feature` tests to both flagd-rpc and flagd-in-process


providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/common/Util.java
```diff
@@ -0,0 +1,23 @@
...
+     *
+     * @param deadline number of milliseconds to block
+     * @param check    {@link AtomicBoolean} to check for status true
+     */
+    public static void busyWaitAndCheck(final Long deadline, final AtomicBoolean check) throws InterruptedException {
+        long start = System.currentTimeMillis();
+
+        do {
+            if (deadline <= System.currentTimeMillis() - start) {
+                throw new RuntimeException(String.format("Initialization not complete after %d ms", deadline));
```
Kavindu-Dodan: Minor - can change this exception to say something like `Deadline exceeded. Condition did not complete within the deadline`

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolver.java
```diff
@@ -70,13 +79,17 @@ public void init() throws Exception {
         });
         stateWatcher.setDaemon(true);
         stateWatcher.start();
+
+        // block till ready
+        Util.busyWaitAndCheck(this.deadline, this.connected);
```
Kavindu-Dodan, line 84: +1 for detecting this and the implementation :)

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/common/Util.java
toddbaert: fixed!

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolver.java
```diff
@@ -133,22 +146,12 @@ private <T> ProviderEvaluation<T> resolve(Class<T> type, String key, T defaultVa
 
         // missing flag
         if (flag == null) {
-            return ProviderEvaluation.<T>builder()
-                    .value(defaultValue)
-                    .reason(Reason.ERROR.toString())
-                    .errorCode(ErrorCode.FLAG_NOT_FOUND)
-                    .errorMessage(String.format("requested flag could not be found: %s", key))
-                    .build();
+            throw new FlagNotFoundError("flag: " + key + " not found");
```
Kavindu-Dodan, line 149: This IMO is one of the hidden features of the SDK. While throwing exceptions is not my first choice, using standard exceptions from SDK is acceptable.

toddbaert: Ya there's a bit of performance cost, but I think that these are truly "Exceptional" situations: ie: not parts of "normal" control flow. If you are evaluating a missing flag or a flag of the wrong type, it's probably an error you should fix.

Kavindu-Dodan: There's one aspect we should remember, SDK log the error [^1]. If a flag configuration is incorrect (invalid targeting rule), then exception usage can result in multiple log entries 🤔 

[^1]: https://github.com/open-feature/java-sdk/blob/main/src/main/java/dev/openfeature/sdk/OpenFeatureClient.java#L140

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/BaseStepDefinitions.java
```diff
@@ -1,29 +1,22 @@
...
-import io.cucumber.java.en.Given;
-import io.cucumber.java.en.Then;
-import io.cucumber.java.en.When;
 
-import java.util.HashMap;
-import java.util.Map;
-
-import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertTrue;
+public class BaseStepDefinitions {
```
Kavindu-Dodan:
```suggestion
public abstract class BaseStepDefinitions {
```

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/BaseStepDefinitions.java
```diff
@@ -1,29 +1,22 @@
...
 
-public class StepDefinitions {
+    protected static Client client;
```
Kavindu-Dodan: suggestion

Shall we do this through a setting 

```java
    public void setClient(Client client) {
        this.client = client;
    }
```

Which then can be called from before all of the sub classes ?

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolver.java
toddbaert: That's a good point.

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/BaseStepDefinitions.java
toddbaert: Fixed.

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/RunFlagdInProcessCucumberTest.java
```diff
@@ -12,8 +12,7 @@
 @IncludeEngines("cucumber")
 @SelectClasspathResource("features")
 @ConfigurationParameter(key = PLUGIN_PROPERTY_NAME, value = "pretty")
-@ConfigurationParameter(key = GLUE_PROPERTY_NAME, value = "dev.openfeature.contrib.providers.flagd.e2e.rpc")
 @ConfigurationParameter(key = GLUE_PROPERTY_NAME, value = "dev.openfeature.contrib.providers.flagd.e2e.process")
-public class RunCucumberTest {
+public class RunFlagdInProcessCucumberTest {
```
toddbaert, line 16: Having 2 of these is how we run 2 suites @Kavindu-Dodan . (see the annotation above which filers on package name). 

The other is `RunFlagdRpcCucumberTest`.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/grpc/GrpcResolver.java
```diff
@@ -347,4 +360,21 @@ private static ImmutableMetadata metadataFromResponse(Message response) {
 
         return builder.build();
     }
+
+    private OpenFeatureError mapError(Exception e) {
```
Kavindu-Dodan, line 365: +1 for capturing and parsing this error

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/grpc/GrpcResolver.java
```diff
@@ -160,28 +165,33 @@ private <ValT, ReqT extends Message, ResT extends Message> ProviderEvaluation<Va
...
+            // cache if cache enabled
+            if (this.isEvaluationCacheable(result)) {
+                this.cache.put(key, result);
+            }
 
-        // cache if cache enabled
-        if (this.isEvaluationCacheable(result)) {
-            this.cache.put(key, result);
+            return result;
+        } catch (Exception e) {
```
Kavindu-Dodan: minor refactoring - shall we simplify catch to be closer to where the exception is thrown 

```Java

        try {
            Message response = strategy.resolve(resolverRef, req, key);
         }catch(){
          //   ... handling 
         }   
         
        // rest of the logic
```            

in my view this helps to understand the logic better

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolver.java
Kavindu-Dodan: we could decide on the strategy we choose here - throw exception and let sdk handle it (convenient, less control) OR generate the response by ourselves (manual handling, more control) :)

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/RunFlagdInProcessCucumberTest.java
Kavindu-Dodan: Nice 🙌 this is a good finding :)

Kavindu-Dodan:
Overall a great addition to our test suite. Left few comments. We can merge once they are finalized

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolver.java
toddbaert: I think the error types are clear enough without logging in the provider. We can just use `log.error` for stuff that only the provider understands (ie: grpc connection lost).

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/grpc/GrpcResolver.java
toddbaert: We could, but then we'd have to declare the `response` var outside the try block, which I tend not to like because it introduces more mutability.

On the other hand, it might be more readable. Do you think it's better to have a null/mutated `response` var and keep the try/catch smaller?

Kavindu-Dodan: We can make `Message response` effectively final, 

```java
        final Message response;
        try {
            response = strategy.resolve(resolverRef, req, key);
        } catch (Exception e) {
            OpenFeatureError openFeatureError = mapError(e);
            throw openFeatureError;
        }
```

personally I think it makes code cleaner and more readable :) anyway let's not worry much about this.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolver.java
Kavindu-Dodan: Sure, let's do that. Will resolve the conversation with this agreemnet

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/StepDefinitions.java
```diff
@@ -48,15 +51,26 @@ public class StepDefinitions {
     private int typeErrorDefaultValue;
     private FlagEvaluationDetails<Integer> typeErrorDetails;
 
+    /**
+     * Injects the client to use for this test.
+     * Tests run one at a time, but just in case, a lock is used to make sure the client is not updated mid-test.
+     * 
+     * @param client client to inject into test.
+     */
+    public static void setClient(Client client) {
```
toddbaert, line 60: @Kavindu-Dodan this is the only somewhat interesting thing about this implementation. To use the same test suite with confidence with different setups, I've done a sort of poor-man's dependency injection here, where we inject the client from the setup file into the test suite.

You can use actual DI frameworks with cucumber for this, but I think it will be much too heavy for our needs.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/grpc/GrpcResolver.java
toddbaert: I will make this change.

toddbaert: Done!

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/StepDefinitions.java
Kavindu-Dodan: yeah this is fine. we now only have a single gherkin test suite to maintain, which is a plus :)

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/BaseStepDefinitions.java
toddbaert: No longer applicable, I've removed the double test suites.

Kavindu-Dodan (approved):

PR #438, opened 2023-09-18
chore: implement json-evaluator e2e

toddbaert:
* adds tests for JSON custom evaluators
* flagd-rpc and flagd-in-process behave identically :raised_hands:


providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/RunFlagdInProcessCucumberTest.java
```diff
@@ -10,9 +10,10 @@
 
 @Suite
 @IncludeEngines("cucumber")
-@SelectClasspathResource("features")
+@SelectClasspathResource("features/evaluation.feature")
+@SelectClasspathResource("features/flagd-json-evaluator.feature")
 @ConfigurationParameter(key = PLUGIN_PROPERTY_NAME, value = "pretty")
-@ConfigurationParameter(key = GLUE_PROPERTY_NAME, value = "dev.openfeature.contrib.providers.flagd.e2e.process,dev.openfeature.contrib.providers.flagd.e2e")
+@ConfigurationParameter(key = GLUE_PROPERTY_NAME, value = "dev.openfeature.contrib.providers.flagd.e2e.process,dev.openfeature.contrib.providers.flagd.e2e.steps")
```
toddbaert, line 16: @Kavindu-Dodan this was causing our whole issue with both providers being set.

This annotation searches SUB packages too (the in-line doc doesn't mention this) so because I supplied the parent package, it was reaching down into it's children and polluting things and therefore loading both `@BeforeAll()`s.

The suites now live in sibling packages and now only 1 `@BeforeAll()` runs each time.

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/steps/StepDefinitions.java
```diff
@@ -81,57 +82,67 @@ public static void setup() {
     @When("a boolean flag with key {string} is evaluated with default value {string}")
     public void a_boolean_flag_with_key_boolean_flag_is_evaluated_with_default_value_false(String flagKey,
             String defaultValue) {
-        this.booleanFlagValue = client.getBooleanValue(flagKey, Boolean.valueOf(defaultValue));
+        this.booleanFlagKey = flagKey;
+        this.booleanFlagDefaultValue = Boolean.valueOf(defaultValue);
```
toddbaert, line 86: I had to change the way these tests all work to be consistent with the new ones (since identical steps must have matching implementations).

Instead of keeping the resolved value as state in the class, I'm keeping the flagkey/default values in most cases, and doing the assertions in the same step as the evaluation.

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/RunFlagdInProcessCucumberTest.java
Kavindu-Dodan: Good catch :)

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/steps/StepDefinitions.java
Kavindu-Dodan: Good change. Now all tests have a consistent pattern - Setup parameters  -> Evaluate -> Validate

Kavindu-Dodan (approved):

PR #445, opened 2023-09-21
fix: await shutdown in in-process mode

toddbaert:
* fixes an issue where the gRPC channel wasn't shutdown properly
* uses deadline for shutdown timeout
* fixes related issues in testing with tests trampling on each other


providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/grpc/GrpcConnector.java
```diff
@@ -91,12 +91,13 @@ public void shutdown() throws Exception {
...
             if (this.channel != null) {
                 this.channel.shutdown();
-                this.channel.awaitTermination(5, TimeUnit.SECONDS);
+                this.channel.awaitTermination(this.deadline, TimeUnit.MILLISECONDS);
             }
         } finally {
             this.cache.clear();
             if (this.channel != null) {
                 this.channel.shutdownNow();
+                this.channel.awaitTermination(this.deadline, TimeUnit.MILLISECONDS);
```
toddbaert, line 100: I'm using the deadline here instead of the arbitrary 5s. This makes sense, because in theory, the only thing blocking the shutdown should be in-flight calls, which will be limited to the deadline.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/grpc/GrpcStreamConnector.java
```diff
@@ -67,13 +70,24 @@ public BlockingQueue<StreamPayload> getStream() {
...
+            if (this.channel != null) {
+                this.channel.shutdown();
+                this.channel.awaitTermination(this.deadline, TimeUnit.MILLISECONDS);
+            }
+        } finally {
+            if (this.channel != null) {
+                this.channel.shutdownNow();
+                this.channel.awaitTermination(this.deadline, TimeUnit.MILLISECONDS);
+            }
+        }
```
toddbaert, line 90: Same as https://github.com/open-feature/java-sdk-contrib/pull/445/files#r1333106690

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/grpc/GrpcConnector.java
```diff
@@ -89,14 +89,16 @@ public void shutdown() throws Exception {
         }
 
         try {
-            if (this.channel != null) {
+            if (this.channel != null && !this.channel.isShutdown()) {
                 this.channel.shutdown();
-                this.channel.awaitTermination(5, TimeUnit.SECONDS);
+                this.channel.awaitTermination(this.deadline, TimeUnit.MILLISECONDS);
```
toddbaert, line 94: I'm using the deadline here instead of the arbitrary 5s. This makes sense, because in theory, the only thing blocking the shutdown should be in-flight calls, which will be limited to the deadline.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/grpc/GrpcStreamConnector.java
```diff
@@ -67,13 +75,25 @@ public BlockingQueue<StreamPayload> getStream() {
...
+     * @throws InterruptedException if stream can't be closed within deadline.
      */
-    public void shutdown() {
+    public void shutdown() throws InterruptedException {
         if (shutdown.getAndSet(true)) {
             return;
         }
 
-        channel.shutdown();
+        try {
```
toddbaert, line 85: same as https://github.com/open-feature/java-sdk-contrib/pull/445/files#r1333294670

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/grpc/GrpcConnector.java
Kavindu-Dodan: Yes, this is a good improvement :)

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/grpc/GrpcConnector.java
```diff
@@ -89,14 +89,16 @@ public void shutdown() throws Exception {
...
             }
         } finally {
             this.cache.clear();
-            if (this.channel != null) {
+            if (this.channel != null && !this.channel.isShutdown()) {
                 this.channel.shutdownNow();
+                this.channel.awaitTermination(5000, TimeUnit.MILLISECONDS);
```
Kavindu-Dodan: do we need both here 🤔 

shutdown now - Initiates a forceful shutdown in which preexisting and new calls are cancelled 
awaitTermination - Waits for the channel to become terminated, giving up if the timeout is reached 

I think we do not need both here. But rather, Log warning -> call `shutdownNow()`. 🤔

toddbaert: From the javadoc:

> initiates a forceful shutdown in which preexisting and new calls are cancelled. Although forceful, **_the shutdown process is still not instantaneous; [isTerminated()](vscode-file://vscode-app/usr/lib/code/out/vs/code/electron-sandbox/workbench/workbench.html) will likely return false immediately after this method returns._**

So this gives us a pretty reasonable guarantee that the channel will be closed after this second wait, which is what I saw in testing as well.

toddbaert: This should be using `thid.deadline` though, not `5000`. I will fix that.

toddbaert:
```suggestion
                this.channel.awaitTermination(this.deadline, TimeUnit.MILLISECONDS);
```

Kavindu-Dodan (approved):

PR #461, opened 2023-09-27
feat: add Flipt provider

liran2000:
add [Flipt](https://www.flipt.io/) provider.

See [readme](https://github.com/liran2000/java-sdk-contrib/blob/4a0d283d2022e1f04d96e50082f1afceecc79354/providers/flipt/README.md#official-flipt-openfeature-provider-for-java) for details.

@markphelps you are welcome to review from Flipt perspective.

@toddbaert you are welcome to review from OpenFeature perspective.


liran2000:
@markphelps build passed locally but failing here since _flipt-java_ is built with Java 11, and this job is build with Java 8.
Possibly _flipt-java_ can have a Java 8 release, for example _0.2.11-java8_ ?

markphelps:
> @markphelps build passed locally but failing here since _flipt-java_ is built with Java 11, and this job is build with Java 8. Possibly _flipt-java_ can have a Java 8 release, for example _0.2.11-java8_ ?

@liran2000 yeah I will try to setup a Java 8 release. Our Java SDK is actually autogenerated, so hopefully there arent any Java 9+ specific features (been a long time since I've written Java)

markphelps:
@liran2000 I dont think it's going to work with Java 8: https://github.com/flipt-io/flipt-java/actions/runs/6355901633/job/17264700354

Looks like our SDK requires some functionality that isn't in 8 (Map.of)

markphelps:
is it possible to update this job to a more recent version of Java?

liran2000:
> Looks like our SDK requires some functionality that isn't in 8 (Map.of)

I can try to take a look at Flipt SDK to see what is the amount of non compliant code. 

> is it possible to update this job to a more recent version of Java?

As a library, if it compiles with newer Java version it will break Java 8 compatability, and will require newer Java version from all usages.

toddbaert:
> is it possible to update this job to a more recent version of Java?

Lots of enterprises are (unfortunately) still using Java 1.8, which is why both the SDK and contribs target (and use) it.

@markphelps I suspect the biggest challenge will be getting your tooling/plugins/etc working with 8, though that's not strictly necessary. You should be able to target 1.8 and then you'll be only making code changes, which from 11 to 8 shouldn't be too huge :grimacing: ... but obviously it depends on your 11 language feature usage.

liran2000:
@markphelps opened [PR](https://github.com/flipt-io/flipt-java/pull/23) for Flipt SDK.
There seems to be only the one mentioned code place to update, so changing the main branch and release seems simpler than a separate release for Java 8.
You can review, merge and release it if you agree for proceeding. Notice the release job itself has to compile with Java 8 too.
Thanks

markphelps:
Hi @liran2000 ! We just released a new version which uses the [sourceCompatibility / targetCompatibility](https://www.baeldung.com/gradle-sourcecompatiblity-vs-targetcompatibility) features of gradle. Ive been told that this should be enough so that the JAR/class files should target Java 8.

Would it be possible to update the Flipt SDK to 0.2.13? https://central.sonatype.com/artifact/io.flipt/flipt-java ?

cc @dsinghvi

liran2000:
> Hi @liran2000 ! We just released a new version which uses the [sourceCompatibility / targetCompatibility](https://www.baeldung.com/gradle-sourcecompatiblity-vs-targetcompatibility) features of gradle. Ive been told that this should be enough so that the JAR/class files should target Java 8.
> 
> Would it be possible to update the Flipt SDK to 0.2.13? https://central.sonatype.com/artifact/io.flipt/flipt-java ?
> 
> cc @dsinghvi

Thanks for the effort, the output JAR classes are now "Java 8", I tried it, it [failed](https://github.com/open-feature/java-sdk-contrib/actions/runs/6480064286/job/17594826802) since the source code still have Java 8+ code. Failure is in the test, which is like a runtime test:
> dev.openfeature.contrib.providers.flipt.FliptProviderTest
Error:  dev.openfeature.contrib.providers.flipt.FliptProviderTest -- Time elapsed: 1.066 s <<< ERROR!
java.lang.NoSuchMethodError: java.util.Map.of(Ljava/lang/Object;Ljava/lang/Object;Ljava/lang/Object;Ljava/lang/Object;Ljava/lang/Object;Ljava/lang/Object;)Ljava/util/Map;

In the link provided, it is mentioned:
> "Additionally, the source code cannot contain lambda expressions or any feature not available in Java 6."

I don't see how it can compile and run with Java 8 without removing the incompatible _Map.of_ from source code.

dsinghvi:
@liran2000 thanks for testing -- this was oversight on my part. Will rerelease a version of the generator today.

providers/flipt/pom.xml
```diff
@@ -0,0 +1,47 @@
...
+
+    <name>flipt</name>
+    <description>Flipt provider for Java</description>
+    <url>https://www.flipt.io/</url>
+
+    <dependencies>
+        <dependency>
+            <groupId>io.flipt</groupId>
+            <artifactId>flipt-java</artifactId>
+            <version>0.2.13</version>
```
markphelps:
```suggestion
            <version>0.2.14</version>
```

should now be java 8 compat

markphelps (approved):
looks great once the Flipt Java SDK is updated! thank you @liran2000 !!

providers/flipt/src/main/java/dev/openfeature/contrib/providers/flipt/FliptProvider.java
```diff
@@ -0,0 +1,240 @@
...
+            .build();
+    }
+
+    private static Map<String, String> buildContextMap(EvaluationContext ctx) {
+        Map<String, String> contextMap = new HashMap<>();
+        ctx.asObjectMap().forEach((k, v) -> {
+            contextMap.put(k, String.valueOf(v));
+        });
+        return contextMap;
+    }
```
toddbaert: This method seems to be placed in a bit of an odd spot (mixed in with the evaluation methods).

providers/flipt/src/main/java/dev/openfeature/contrib/providers/flipt/FliptProvider.java
```diff
@@ -0,0 +1,240 @@
...
+    public ProviderEvaluation<Value> getObjectEvaluation(String key, Value defaultValue, EvaluationContext ctx) {
+        if (!ProviderState.READY.equals(state)) {
+            if (ProviderState.NOT_READY.equals(state)) {
+                throw new ProviderNotReadyError(PROVIDER_NOT_YET_INITIALIZED);
+            }
+            throw new GeneralError(UNKNOWN_ERROR);
+        }
+        Map<String, String> contextMap = buildContextMap(ctx);
+        EvaluationRequest request = EvaluationRequest.builder().namespaceKey(fliptProviderConfig.getNamespace())
+            .flagKey(key).entityId(ctx.getTargetingKey()).context(contextMap).build();
```
toddbaert, line 188: `entityId(ctx.getTargetingKey())`

Is there any issue if `ctx.getTargetingKey()` is null here?

providers/flipt/src/main/java/dev/openfeature/contrib/providers/flipt/FliptProvider.java
```diff
@@ -0,0 +1,240 @@
...
+
+        VariantEvaluationResponse response;
+        try {
+            response = fliptApiClient.evaluation().variant(request);
+        } catch (Exception e) {
+            log.error("Error evaluating variant", e);
+            return ProviderEvaluation.<Value>builder()
+                .value(defaultValue)
+                .errorMessage(e.getMessage())
+                .build();
```
toddbaert: Throw a [General Error](https://github.com/open-feature/java-sdk/blob/main/src/main/java/dev/openfeature/sdk/exceptions/GeneralError.java) here instead.

providers/flipt/src/main/java/dev/openfeature/contrib/providers/flipt/FliptProvider.java
```diff
@@ -0,0 +1,240 @@
...
+        }
+
+        if (!response.getMatch()) {
+            log.debug("non matching variant for {}", key);
+            return ProviderEvaluation.<Value>builder()
+                .value(defaultValue)
+                .reason(DEFAULT.name())
+                .build();
```
toddbaert: I'm wonder if this should actually be an error? If there's no match for the variant?

providers/flipt/src/main/java/dev/openfeature/contrib/providers/flipt/FliptProvider.java
liran2000: I don't see an issue, it is working even when _ctx.getTargetingKey()_ is null, as entity Id is null. Confirmed via running a unit test via real Flipt instance.

providers/flipt/src/main/java/dev/openfeature/contrib/providers/flipt/FliptProvider.java
liran2000: I don't think it should be an error.
Test case:
- flag configured with variant=x for segment by userId==1.
- flag is evaluated without userId.

In this case, there is no match, API evaluates to no match.
Provider returning the default value.

toddbaert: Ah, I see.

toddbaert:
I'm ready to approve once the above unresolved comments are resolved.

liran2000:
> I'm ready to approve once the above unresolved comments are resolved.

Comments are addressed in recents commits, can you check latest code?

providers/flipt/src/main/java/dev/openfeature/contrib/providers/flipt/FliptProvider.java
liran2000: Done

providers/flipt/src/main/java/dev/openfeature/contrib/providers/flipt/FliptProvider.java
liran2000: Handled

toddbaert:
> > I'm ready to approve once the above unresolved comments are resolved.
> 
> Comments are addressed in recents commits, can you check latest code?

Oh I'm sorry, I kept following sent to old permalinks to previous commits! Yes I see they are resolved. Approving!

toddbaert (approved):

markphelps:
Thank you @liran2000 for all your hard work on this!!! And thank you @toddbaert for the review. Will update our docs as well to mention this new provider

PR #468, opened 2023-10-01
test: zero value and events tests

DBlanchard88:
- Flag change event: in-process-provider does not support flag change list.
- Zero Value - float/double value: BUG - When testing locally defaultValue will not produce 0.0. Instead defaultValue is stuck at 0.1.    Only the in-process evaluator seems to be impacted

### DIFF
--- a/providers/flagd/pom.xml
+++ b/providers/flagd/pom.xml
@@ -321,10 +321,10 @@
                                     <goal>exec</goal>
                                 </goals>
                                 <configuration>
-                                        <!-- copy the feature spec we want to test into resources so them can be easily loaded -->
-                                        <!-- run: cp test-harness/features/evaluation.feature src/test/resources/features/ -->
-                                        <executable>cp</executable>
-                                        <arguments>
+                                    <!-- copy the feature spec we want to test into resources so them can be easily loaded -->
+                                    <!-- run: cp test-harness/features/evaluation.feature src/test/resources/features/ -->
+                                    <executable>cp</executable>
+                                    <arguments>
                                         <argument>test-harness/gherkin/flagd.feature</argument>
                                         <argument>src/test/resources/features/</argument>
                                     </arguments>
--- a/providers/flagd/pom.xml
+++ b/providers/flagd/pom.xml
@@ -322,7 +322,7 @@
                                 </goals>
                                 <configuration>
                                     <!-- copy the feature spec we want to test into resources so them can be easily loaded -->
-                                    <!-- run: cp test-harness/features/evaluation.feature src/test/resources/features/ -->
+                                    <!-- run: cp test-harness/features/flagd.feature src/test/resources/features/ -->
                                     <executable>cp</executable>
                                     <arguments>
                                         <argument>test-harness/gherkin/flagd.feature</argument>
--- a/providers/flagd/pom.xml
+++ b/providers/flagd/pom.xml
@@ -314,6 +314,22 @@
                                     </arguments>
                                 </configuration>
                             </execution>
+                            <execution>
+                                <id>copy-gherkin-flagd.feature</id>
+                                <phase>validate</phase>
+                                <goals>
+                                    <goal>exec</goal>
+                                </goals>
+                                <configuration>
+                                        <!-- copy the feature spec we want to test into resources so them can be easily loaded -->
+                                        <!-- run: cp test-harness/features/evaluation.feature src/test/resources/features/ -->
+                                        <executable>cp</executable>
+                                        <arguments>
+                                        <argument>test-harness/gherkin/flagd.feature</argument>
+                                        <argument>src/test/resources/features/</argument>
+                                    </arguments>
+                                </configuration>
+                            </execution>
                         </executions>
                     </plugin>
                 </plugins>
--- a/providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/steps/StepDefinitions.java
+++ b/providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/steps/StepDefinitions.java
@@ -1,16 +1,21 @@
 package dev.openfeature.contrib.providers.flagd.e2e.steps;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.mockito.Mockito.atMost;
 
 import java.util.HashMap;
 import java.util.Map;
+import java.util.function.Consumer;
 
+import org.awaitility.Awaitility;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Order;
 import org.junit.jupiter.api.parallel.Isolated;
 
 import dev.openfeature.sdk.Client;
 import dev.openfeature.sdk.EvaluationContext;
+import dev.openfeature.sdk.EventDetails;
 import dev.openfeature.sdk.FeatureProvider;
 import dev.openfeature.sdk.FlagEvaluationDetails;
 import dev.openfeature.sdk.ImmutableContext;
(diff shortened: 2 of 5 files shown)


providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/steps/StepDefinitions.java
```diff
@@ -374,4 +388,107 @@ public void the_returned_value_should_be(String expected) {
...
+    @Then("the resolved boolean zero-value should be {string}")
+    public void the_resolved_boolean_zero_value_should_be(String expected) {
+        boolean value = client.getBooleanValue(this.booleanFlagKey, this.booleanFlagDefaultValue);
+         assertEquals(Boolean.valueOf(expected),value);   
+    }
+
+    // float/double value
+    @When("a zero-value float flag with key {string} is evaluated with default value {double}")
+    public void a_zero_value_float_flag_with_key_is_evaluated_with_default_value(String flagKey, Double defaultValue) {
+        // TODO: There is a bug here with 0 value floats
```
DBlanchard88, line 461: When testing locally defaultValue will not produce 0.0. Instead defaultValue is stuck at 0. Only the in-process evaluator seems to be impacted.

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/steps/StepDefinitions.java
```diff
@@ -433,47 +433,53 @@ public void a_provider_ready_handler_is_added() {
...
     @Then("the resolved boolean zero-value should be {string}")
     public void the_resolved_boolean_zero_value_should_be(String expected) {
         boolean value = client.getBooleanValue(this.booleanFlagKey, this.booleanFlagDefaultValue);
-         assertEquals(Boolean.valueOf(expected),value);   
+        assertEquals(Boolean.valueOf(expected), value);
     }
 
     // float/double value
     @When("a zero-value float flag with key {string} is evaluated with default value {double}")
     public void a_zero_value_float_flag_with_key_is_evaluated_with_default_value(String flagKey, Double defaultValue) {
```
DBlanchard88, line 462: When testing locally defaultValue will not produce 0.0. Instead defaultValue is stuck at 0. Only the in-process evaluator seems to be impacted.

toddbaert: @Kavindu-Dodan @beeme1mr  I think there's a bug here. I did a quick review and it seems @DBlanchard88 is right. I think there's a  interesting edge case where when one variants is `0.0` as is the case [here](https://github.com/open-feature/flagd-testbed/blob/main/flags/zero-flags.json#L27-L34), our in-process evaluator parses the variant values as `Integer`. I'm not sure if it matters if the `0.0` value variant comes first or not. See this screenshot from my debugger: 
image

Note how both values are `Integer` type. :thinking: 

I assume we can somehow configure the Json mapping annotations to help with this, but it might be tricky.

toddbaert: @Kavindu-Dodan I'm going to open a bug for this and assign myself since I think I understand what's happening. Feel free to tell me I'm wrong if you think I've missed something.

Kavindu-Dodan: Yeah, this is intersting 🤔 I cannot reproduce this observation locally 

```json
    "float-zero-flag": {
      "state": "ENABLED",
      "variants": {
        "zero": 0.0,
        "non-zero": 1.0
      },
      "defaultVariant": "zero"
    }
```

And ,

image


I have Java version `java 17.0.8 ` if that matters 🤔

toddbaert: @Kavindu-Dodan That's wild. I'm able to reproduce this on openjdk 8, 17, 20. As an experiment, and a sanity check, I changed the variant names (appended `"f"`) and set the non-zero value to `1.1`. When I do this, the non-zero variant is correctly a double, and the `0.0` variant is an int:

image

image

If you remove the commented assertion in @DBlanchard88 's tests, does it pass for you?

I think a deeper issue might be that with the in-process provider, we seem to consider resolving floats as ints and vice-versa, as a `TYPE_ERROR`. **This is different than flagd RPC**. Flagd RPC will happily return you a float flag as an int (rounding) and vice versa. cc @beeme1mr 

Do you guys think I should [update this issue](https://github.com/open-feature/java-sdk-contrib/issues/470) (and maybe the in-process spec) to specify float/int conversion?

providers/flagd/pom.xml
```diff
@@ -314,6 +314,22 @@
...
                             </execution>
+                            <execution>
+                                <id>copy-gherkin-flagd.feature</id>
+                                <phase>validate</phase>
+                                <goals>
+                                    <goal>exec</goal>
+                                </goals>
+                                <configuration>
+                                    <!-- copy the feature spec we want to test into resources so them can be easily loaded -->
+                                    <!-- run: cp test-harness/features/evaluation.feature src/test/resources/features/ -->
```
toddbaert:
```suggestion
                                    <!-- run: cp test-harness/features/flagd.feature src/test/resources/features/ -->
```

toddbaert (approved):
Seeing as how we have a separate issue for [this](https://github.com/open-feature/java-sdk-contrib/pull/468#discussion_r1342179921), and there's useful tests here, I move we merge this as is. I will resolve the broken test/comments _somehow_ with [this](https://github.com/open-feature/java-sdk-contrib/issues/470).

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/steps/StepDefinitions.java
Kavindu-Dodan: Yup, this fails locally and only when I am using grpc sync stream. The example of mine above was a simple parsing from a file (sorry for not stating this)

So the loss of decimals happens either in wire OR at JSON decoding time. I will check in this direction

Kavindu-Dodan: The issue boils down to how to how Go unmarshel json to structs. While it store float64 values, decmials with trailing zeros are stored as integers (`0.0` -> `0`) 

This is why the sync stream sends us a `0` value and the evaluation fails. I think the best solution here is to handle such corner cases at the evaluator's end. 

In Java, we need special handling for Double - check if the variant can be converted to a double (`Double.parseDouble`) and use that value as the resolved value.

Kavindu-Dodan: > The issue boils down to how to how Go unmarshel json to structs. While it store float64 values, decmials with trailing zeros are stored as integers (`0.0` -> `0`)
> 
> This is why the sync stream sends us a `0` value and the evaluation fails. I think the best solution here is to handle such corner cases at the evaluator's end.
> 
> In Java, we need special handling for Double - check if the variant can be converted to a double (`Double.parseDouble`) and use that value as the resolved value.

Kavindu-Dodan (approved):

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/steps/StepDefinitions.java
toddbaert: > The issue boils down to how to how Go unmarshel json to structs. While it store float64 values, decmials with trailing zeros are stored as integers (0.0 -> 0)

I was starting to suspect this was the case.

> In Java, we need special handling for Double - check if the variant can be converted to a double (Double.parseDouble) and use that value as the resolved value.

I was thinking this too... in fact, to be consistent with RPC, I think we have to attempt conversion in both directions, int -> float and float -> int. Additionally, if we do this, we may need to specify how the float -> int conversion is done. I think Java and Go both  round toward zero. I believe .NET does too. This might be worth specifying.

PR #505, opened 2023-10-17
feat: Allow global otel configuration extraction

### LINKED ISSUE #390: [flagd] follow-up refactoring

Kavindu-Dodan:
## This PR

Fixes #390 by completing the missing task - `Use always OTel interceptor`

Flagd java provider now allows extracting tracer from `GlobalTelemetry`

```java
FlagdOptions options =
        FlagdOptions.builder()
            .withGlobalTelemetry(true)
            .build();

FlagdProvider flagdProvider = new FlagdProvider(options);
```


providers/flagd/README.md
```diff
@@ -162,8 +162,17 @@ FlagdOptions options =
 FlagdProvider flagdProvider = new FlagdProvider(options);
 ```
 
-Please refer [OpenTelemetry example](https://opentelemetry.io/docs/instrumentation/java/manual/#example) for best 
-practice guideline.
+Alternatively, if you already have set up `GlobalOpenTelemetry` in your provider runtime, then allow flagd to derive tracer using`withGlobalTelemetry` option.
```
thisthat: nit 
```suggestion
Alternatively, if you already have set up `GlobalOpenTelemetry` in your provider runtime, then allow flagd to derive tracer using `withGlobalTelemetry` option.
```

providers/flagd/README.md
```diff
@@ -146,7 +146,7 @@ By default, the provider is configured to
 use [least recently used (lru)](https://commons.apache.org/proper/commons-collections/apidocs/org/apache/commons/collections4/map/LRUMap.html)
 caching with up to 1000 entries.
 
-### OpenTelemetry support (RPC only)
+### OpenTelemetry tracing (RPC only)
 
 OpenTelemetry support can be enabled either through [automatic instrumentation](https://opentelemetry.io/docs/instrumentation/java/automatic/) 
 or with [manual instrumentation](https://opentelemetry.io/docs/instrumentation/java/manual/).
```
thisthat: reading this is a bit misleading 🤔 
it is true that automatic instrumentation will produce gRPC spans, but it doesn't provide the attribute infos that we expose through the manual instrumentation

Kavindu-Dodan: We still get traces from automatic instrumentation. See this code [1]. As long as we have a tracer, span will be created and exported. 

[1] - https://github.com/open-feature/java-sdk-contrib/blob/main/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/grpc/strategy/TracedResolving.java#L28-L35

toddbaert (approved):
Approving, since this all makes sense to me. I am, however, very confused by [this thread](https://github.com/open-feature/java-sdk-contrib/pull/505#discussion_r1363340069), so I'd like to hear back from @thisthat at some point.

providers/flagd/README.md
thisthat: The problem that I have with how the whole paragraph is written is that it seems I can rely only on auto-instrumentation. This to me sounds like I don't need to provide **any** configuration to the provider about OTel. However, this would lead to the FlagD Provider using the `SimpleResolving` which doesn't augment traces.

Kavindu-Dodan: I updated this section of the documentation. Please check and let me know if this clears the doubts :)  - https://github.com/open-feature/java-sdk-contrib/blob/b0d3bb02ee5809db8f62cf22521e74f2bcdb4830/providers/flagd/README.md#opentelemetry-tracing-rpc-only

thisthat (approved):

providers/flagd/README.md
thisthat: Thanks a lot! ✅

toddbaert (approved):

PR #512, opened 2023-10-21
feat: `$flagd.timestamp` added to in-process evaluator 

DBlanchard88:
- added the ability to use $flagd.timestamp
- added testing to verify implementation of timestamp


providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/targeting/Operator.java
```diff
@@ -3,22 +3,25 @@
...
- * Targeting operator wraps JsonLogic handlers and expose a simple API for external layers.
+ * Targeting operator wraps JsonLogic handlers and expose a simple API for
+ * external layers.
  * This helps to isolate external dependencies to this package.
  */
 public class Operator {
 
-    static final String FLAGD_PROPS_KEY = "$flagd";
+    public static final String FLAGD_PROPS_KEY = "$flagd";
+    public static final String READABLE_TIME_STAMP = "readableTimestamp";
```
toddbaert:
```suggestion
```

This seems unused.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/targeting/Operator.java
```diff
@@ -60,9 +67,9 @@ static class FlagProperties {
                 Map<?, ?> dataMap = (Map<?, ?>) from;
 
                 final Object flagKey = Optional.ofNullable(dataMap.get(FLAGD_PROPS_KEY))
-                    .filter(flagdProps -> flagdProps instanceof Map)
-                    .map(flagdProps -> ((Map<?, ?>)flagdProps).get(FLAG_KEY))
-                    .orElse(null);
+                        .filter(flagdProps -> flagdProps instanceof Map)
+                        .map(flagdProps -> ((Map<?, ?>) flagdProps).get(FLAG_KEY))
+                        .orElse(null);
```
toddbaert: I think you'll need to do something similar here to add the timestamp to this little nested class. The constructor here takes a context, and extracts all the "injected" properties from it. I think it's surprising the timestamp isn't included. The same logic used to extract the `$flagd.flagKey` here on these lines can be used to extract the timestamp too - you might want to make a little private function that accepts an arg for the "sub-property" under `$flag` and returns it.

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/resolver/process/targeting/OperatorTest.java
```diff
@@ -33,6 +35,27 @@ void flagKeyPresent() throws TargetingRuleException {
...
+    void timestampPresent() throws TargetingRuleException {
+        // given
+
+        // rule asserting $flagd.timestamp is a number (i.e., a Unix timestamp)
+        final String targetingRule = "{\"var\":[\"$flagd.timestamp\"]}";
+
+        // when
+        Object evalVariant = OPERATOR.apply("some-key", targetingRule, new ImmutableContext());
+
+        long timestamp = (long)Double.parseDouble(evalVariant.toString());
```
toddbaert:
```suggestion
        // when
        Object timestampString = OPERATOR.apply("some-key", targetingRule, new ImmutableContext());

        long timestamp = (long)Double.parseDouble(timestampString.toString());
```

The result here isn't really a variant, because of how you wrote the rule. I think this name makes more sense.

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/resolver/process/targeting/OperatorTest.java
```diff
@@ -33,6 +35,27 @@ void flagKeyPresent() throws TargetingRuleException {
...
+
+        // generating current unix timestamp & 5 minute threshold
+        long currentTimestamp = Instant.now().getEpochSecond();
+        long thresholdPast = currentTimestamp - (5 * 60);
+        long thresholdFuture = currentTimestamp + (5 * 60);
```
toddbaert: We can probably make this tighter - like a 5 second grace window instead of 5 minutes - the test shouldn't ever take longer than 5s

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/resolver/process/targeting/OperatorTest.java
```diff
@@ -1,9 +1,11 @@
 package dev.openfeature.contrib.providers.flagd.resolver.process.targeting;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.util.HashMap;
 import java.util.Map;
+import java.time.*;
```
toddbaert:
```suggestion
import java.time.Instant;
```

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/targeting/Operator.java
```diff
@@ -3,22 +3,25 @@
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Optional;
+import java.time.*;
```
toddbaert:
```suggestion
import java.time.Instant;
```

toddbaert (changes requested):
This is awesome! Thank you! And thanks for adding the solid test. I left a bunch of small nits and style feedback, but [this](https://github.com/open-feature/java-sdk-contrib/pull/512/files#r1368688840) is my only blocker for approval.

.github/workflows/ci.yml
```diff
@@ -28,7 +28,7 @@ jobs:
         uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4
 
       - name: Set up JDK 8
-        uses: actions/setup-java@cd89f46ac9d01407894225f350157564c9c7cee2 # v3
+        uses: actions/setup-java@0ab4596768b603586c0de567f2430c30f5b0d2b0 # v3
```
toddbaert: please revert this change, perhaps this was an accident when you resolved a conflict.

.github/workflows/release-please.yml
```diff
@@ -23,7 +23,7 @@ jobs:
 
       - name: Set up JDK 8
         if: ${{ steps.release.outputs.releases_created }}
-        uses: actions/setup-java@cd89f46ac9d01407894225f350157564c9c7cee2 # v3
+        uses: actions/setup-java@0ab4596768b603586c0de567f2430c30f5b0d2b0 # v3
```
toddbaert: https://github.com/open-feature/java-sdk-contrib/pull/512/files#r1369009108

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/targeting/Operator.java
```diff
@@ -3,22 +3,24 @@
...
 /**
- * Targeting operator wraps JsonLogic handlers and expose a simple API for external layers.
+ * Targeting operator wraps JsonLogic handlers and expose a simple API for
+ * external layers.
  * This helps to isolate external dependencies to this package.
  */
 public class Operator {
 
-    static final String FLAGD_PROPS_KEY = "$flagd";
+    public static final String FLAGD_PROPS_KEY = "$flagd";
```
toddbaert: Is there a reason to make this public?

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/targeting/Operator.java
```diff
@@ -53,34 +59,38 @@ public Object apply(final String flagKey, final String targetingRule, final Eval
...
         }
+
+        private String extractSubPropertyFromFlagd(Map<?, ?> dataMap, String flagdKey) {
+            return Optional.ofNullable(dataMap.get(FLAGD_PROPS_KEY))
+                    .filter(flagdProps -> flagdProps instanceof Map)
+                    .map(flagdProps -> ((Map<?, ?>) flagdProps).get(flagdKey))
+                    .filter(value -> value instanceof String)
+                    .map(String::valueOf)
+                    .orElse(null);
+        }
```
toddbaert: I think you can make this static.

toddbaert:
```suggestion
        private String extractSubPropertyFromFlagd(Map<?, ?> dataMap, String propertyName) {
            return Optional.ofNullable(dataMap.get(FLAGD_PROPS_KEY))
                    .filter(flagdProps -> flagdProps instanceof Map)
                    .map(flagdProps -> ((Map<?, ?>) flagdProps).get(propertyName))
                    .filter(value -> value instanceof String)
                    .map(String::valueOf)
                    .orElse(null);
        }
```

This is not a flag key.

toddbaert:
```suggestion
        private String extractSubPropertyFromFlagd(Map<?, ?> dataMap, String propertyName) {
            return Optional.ofNullable(dataMap.get(FLAGD_PROPS_KEY))
                    .filter(flagdProps -> flagdProps instanceof Map)
                    .map(flagdProps -> ((Map<?, ?>) flagdProps).get(propertyName))
                    .filter(value -> value instanceof String)
                    .map(String::valueOf)
                    .orElse(null);
        }
```

This is not a flag key.

toddbaert:
@DBlanchard88 I fixed your branch - please just sign it off again.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/targeting/Operator.java
```diff
@@ -53,34 +59,38 @@ public Object apply(final String flagKey, final String targetingRule, final Eval
...
-                    this.flagKey = (String) flagKey;
-                } else {
-                    this.flagKey = null;
-                }
+                this.flagKey = extractSubPropertyFromFlagd(dataMap, FLAG_KEY);
+                this.timestamp = extractSubPropertyFromFlagd(dataMap, TIME_STAMP);
 
                 final Object targetKey = dataMap.get(TARGET_KEY);
-
+
```
toddbaert, line 73:
```suggestion

```

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/resolver/process/targeting/OperatorTest.java
```diff
@@ -33,6 +35,27 @@ void flagKeyPresent() throws TargetingRuleException {
...
+
+        // rule asserting $flagd.timestamp is a number (i.e., a Unix timestamp)
+        final String targetingRule = "{\"var\":[\"$flagd.timestamp\"]}";
+
+        // when
+        Object timestampString = OPERATOR.apply("some-key", targetingRule, new ImmutableContext());
+
+        long timestamp = (long)Double.parseDouble(timestampString.toString());
+
+        // generating current unix timestamp & 5 minute threshold
```
toddbaert, line 50:
```suggestion
        // generating current unix timestamp for testing
```

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/resolver/process/targeting/OperatorTest.java
```diff
@@ -33,6 +35,27 @@ void flagKeyPresent() throws TargetingRuleException {
...
+        long currentTimestamp = Instant.now().getEpochSecond();
+        long thresholdPast = currentTimestamp - (5);
+        long thresholdFuture = currentTimestamp + (5);
+
+        // checks if the timestamp is within 5 minutes of the current time
```
toddbaert, line 55:
```suggestion
        // checks if the timestamp is within 5 seconds of the current time (test tolerance)
```

toddbaert (approved):
Approved with minor nits.

beeme1mr (approved):

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/targeting/Operator.java
```diff
@@ -38,8 +40,12 @@ public Operator() {
      */
     public Object apply(final String flagKey, final String targetingRule, final EvaluationContext ctx)
             throws TargetingRuleException {
-        final Map<String, String> flagdProperties = new HashMap<>();
+        final Map<String, Object> flagdProperties = new HashMap<>();
```
Kavindu-Dodan, line 43: minor remark - you can store string instead of object and this reduced complexity of obtaining this back in `extractSubPropertyFromFlagd`

toddbaert: I'm not sure I follow, - the `unixTimestamp` is a long - are you suggesting we'd convert it to a string and back? I don't think that parsing is worth the small amount of type safety achieved.

Kavindu-Dodan: actually we have a bug if we do not store this as a string 

After storage in map, value extraction happens at `extractSubPropertyFromFlagd` [1]. If we do not store a String, then the value will be always null. 

We also need to add a test case for the fractional evaluator to use the timestamp. This is because that's the only place we extract these values right now. 

[1] - https://github.com/open-feature/java-sdk-contrib/pull/512/files#diff-d6fbc93c9666fdf8ee1e842ca02384dfc839756320162f7978be207739a3f095R91-R92

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/targeting/Operator.java
```diff
@@ -53,34 +59,37 @@ public Object apply(final String flagKey, final String targetingRule, final Eval
...
             }
         }
+
+        private static String extractSubPropertyFromFlagd(Map<?, ?> dataMap, String propertyName) {
+            return Optional.ofNullable(dataMap.get(FLAGD_PROPS_KEY))
+                    .filter(flagdProps -> flagdProps instanceof Map)
+                    .map(flagdProps -> ((Map<?, ?>) flagdProps).get(propertyName))
+                    .filter(value -> value instanceof String)
+                    .map(String::valueOf)
+                    .orElse(null);
```
Kavindu-Dodan: This always return null for timestamp as we are storing it as long. Needs to fix this and also I would like to see a simple fractional test (as this is the only place right now that we use these context values) . See a sample test here - https://github.com/open-feature/java-sdk-contrib/blob/main/providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/resolver/process/targeting/OperatorTest.java#L37

Kavindu-Dodan (changes requested):

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/targeting/Operator.java
toddbaert: We are directly using the timestamp [here](https://github.com/open-feature/java-sdk-contrib/pull/512/files#diff-3c2fc3d40e2a528288cb6e14d60ead19aa8f2ad3ff2d328b0e04fa09cf14a1d9R38-R58) in this new test.

I think you're right that we're missing a test case, but I think it would have to be a test case that manually runs `extractSubPropertyFromFlagd`, not one that uses some kind of fractional rule (I don't know how you'd use the timestamp IN a fractional rule; only the flag key is used by that rule implementation, which we did not change).

Basically, to avoid confusion, I suggested @DBlanchard88  extend the `extractSubPropertyFromFlagd` method to include this new prop, but that extracted prop is not used anywhere by other rules - we probably should just directly test it (which we weren't before).

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/targeting/Operator.java
toddbaert: This seems more like a problem with the validation in `extractSubPropertyFromFlagd`, than a problem with how it's stored. Can we just modify the filter call: `.filter(value -> value instanceof String)` to be instanceof Object?

I still think parsing back and forth from string is a bad idea. :thinking:

Kavindu-Dodan: @toddbaert Sorry I confused myself about the usage. 

Jsonlogic engine can also expand variable references so yes the issue is with `extractSubPropertyFromFlagd `. so the fix should go there.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/targeting/Operator.java
Kavindu-Dodan: @toddbaert agrees with the suggestion.

Kavindu-Dodan: > I don't know how you'd use the timestamp IN a fractional rule; only the flag key is used by that rule implementation, which we did not change

@toddbaert on a second thought, with this background, should we even attempt to expose timestamp through `FlagProperties` ? The confusion for me came from the usage of timestamp at `FlagProperties`. But like you said we don't even need it there 🤔  Json logic can expand the value when this property used in other references.

toddbaert: It's just that the name of the method seems to indicate that it will extract all the properties. I would use it assuming that, which might cause bugs or confusion if I did... maybe at some point we implement some "time-based rollout" rule - I might want to use the timestamp in that rule and then be confused why that prop is missing.

Kavindu-Dodan: that's a fair point. Then let's stick to your proposal - improve this method to extract the newly added timestamp and add a simple validation test

.github/workflows/ci.yml
```diff
@@ -28,7 +28,11 @@ jobs:
         uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4
 
       - name: Set up JDK 8
+<<<<<<< Updated upstream
```
beeme1mr: It looks like there was an issue with a recent merge.

beeme1mr (dismissed):
Please fix the merge issue in the workflow ci.yml.

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/resolver/process/targeting/OperatorTest.java
```diff
@@ -33,6 +35,46 @@ void flagKeyPresent() throws TargetingRuleException {
...
+        Map<String, Object> flagdProperties = new HashMap<>();
+        flagdProperties.put(Operator.FLAG_KEY, "some-key");
+        flagdProperties.put(Operator.TIME_STAMP, 1634000000L); // Changed to long
+
+        Map<String, Object> dataMap = new HashMap<>();
+        dataMap.put(Operator.FLAGD_PROPS_KEY, flagdProperties);
+
+        // When
+        Object flagKey = Operator.FlagProperties.extractSubPropertyFromFlagd(dataMap, Operator.FLAG_KEY);
+        Object timestamp = Operator.FlagProperties.extractSubPropertyFromFlagd(dataMap, Operator.TIME_STAMP);
```
toddbaert: Testing `extractSubPropertyFromFlagd` is less important than testing the constructor ([the constructor is what's actually used in code](https://github.com/open-feature/java-sdk-contrib/blob/d6ba46a540b0d28213998b8ff25418950b652af2/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/targeting/Fractional.java#L27)).

All you have to do is slightly modify this test to instead construct a `FlagProperties`, instance, passing it `dataMap` and make sure it's members (`getTimestamp()` and `getFlagKey`) are what you'd expect. That will indirectly test `extractSubPropertyFromFlagd`, which can just be a private method, used in the constructor.

Kavindu-Dodan (approved):

thisthat (approved):

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/resolver/process/targeting/OperatorTest.java
```diff
@@ -33,6 +35,45 @@ void flagKeyPresent() throws TargetingRuleException {
...
+
+        Map<String, Object> dataMap = new HashMap<>();
+        dataMap.put(Operator.FLAGD_PROPS_KEY, flagdProperties);
+
+        // When
+        Operator.FlagProperties flagProperties = new Operator.FlagProperties(dataMap);
+
+        // Then
+        assertEquals("some-key", flagProperties.getFlagKey());
+        assertEquals(1634000000L, flagProperties.getTimestamp());
```
toddbaert, line 74: great!

toddbaert (approved):

PR #521, opened 2023-10-31
feat: Add ConfigCat provider

liran2000:
@z4kn4fein @laliconfigcat @novalisdenahi can review from ConfigCat point of view.

[Readme](https://github.com/open-feature/java-sdk-contrib/blob/575959a414073f1b031322af4fd98584be2df681/providers/configcat/README.md#unofficial-configcat-openfeature-provider-for-java) describes the provider.


providers/configcat/src/main/java/dev/openfeature/contrib/providers/configcat/ConfigCatProviderConfig.java
```diff
@@ -0,0 +1,24 @@
...
+@Getter
+@Builder
+public class ConfigCatProviderConfig {
+    private Consumer<ConfigCatClient.Options> options;
+
+    // Only holding temporary for initialization
+    private String sdkKey;
+
+    public void postInit() {
+        sdkKey = null; // for security, not holding key in memory for long-term
```
laliconfigcat, line 22: I don't really see the benefit of setting this to null. The ConfigCat SDK holds the SDK key anyway and it needs to know about it during the full run.

providers/configcat/src/main/java/dev/openfeature/contrib/providers/configcat/ConfigCatProvider.java
```diff
@@ -0,0 +1,205 @@
...
+        });
+    }
+
+    @Override
+    public Metadata getMetadata() {
+        return () -> NAME;
+    }
+
+    @Override
+    public ProviderEvaluation<Boolean> getBooleanEvaluation(String key, Boolean defaultValue, EvaluationContext ctx) {
```
laliconfigcat, line 92: As all the evaluations are basically doing the same, I'd introduce a generic evaluate function and call it from the getBooleanEvaluation, getStringEvaluation, getIntegerEvaluation, getDoubleEvaluation.

providers/configcat/src/main/java/dev/openfeature/contrib/providers/configcat/ConfigCatProvider.java
```diff
@@ -0,0 +1,205 @@
...
+            throw new GeneralError("already initialized");
+        }
+        super.initialize(evaluationContext);
+        configCatClient = ConfigCatClient.get(configCatProviderConfig.getSdkKey(),
+            configCatProviderConfig.getOptions());
+        configCatProviderConfig.postInit();
+        state = ProviderState.READY;
+        log.info("finished initializing provider, state: {}", state);
+
+        configCatClient.getHooks().addOnClientReady(() -> {
```
laliconfigcat: I'm not 100% sure that if you add the addOnClientReady hook after the initialization, it will fire in all cases. 
I'd move the hook configuration inside the ConfigCatClient.get call, like here: https://configcat.com/docs/sdk-reference/java/#hooks

After that, I think the state = ProviderState.READY; can be removed after the configCatProviderConfig.postInit(); call. I think, the clientReady should set the state only.

providers/configcat/src/main/java/dev/openfeature/contrib/providers/configcat/ConfigCatProvider.java
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    @Override
+    public Metadata getMetadata() {
+        return () -> NAME;
+    }
+
+    @Override
+    public ProviderEvaluation<Boolean> getBooleanEvaluation(String key, Boolean defaultValue, EvaluationContext ctx) {
+        if (!ProviderState.READY.equals(state)) {
```
laliconfigcat: The ConfigCat SDK has a pretty good way to handle the client initialization delays. 
In auto polling mode, you can set a maxInitWaitTimeSeconds (defaults to 5 seconds).
If you call the getValue or getValueDetails right after the configcatclient initialization, and the client is not ready yet, the client will wait up to the maxInitWaitTimeSeconds to be ready and to provide a valid value. This is a pretty good safety net and I think it would be better to rely on this behaviour instead of throwing errors. 
So I'd remove all these throw errors from the beginning of these methods.

providers/configcat/src/main/java/dev/openfeature/contrib/providers/configcat/ConfigCatProvider.java
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    /**
+     * Initialize the provider.
+     * @param evaluationContext evaluation context
+     * @throws Exception on error
+     */
+    @Override
+    public void initialize(EvaluationContext evaluationContext) throws Exception {
+        boolean initialized = isInitialized.getAndSet(true);
```
laliconfigcat, line 59: The ConfigCatClient.get handles concurrency and multiple initializations. If you call the ConfigCatClient.get multiple times with the same SDK key, it will return a singleton instance of the configcatclient for that SDK Key.  
This is only valid if the hooks will be initialized in the configCatClient.Get call instead of adding them in separate calls (my next comment).
I think this initialized part is unnecessary because of the singleton pattern used in the ConfigCatClient.get, but it is just my opinion.

laliconfigcat (changes requested):
Thank you very much for your efforts! Awesome to see how easy it is to integrate our SDK with OpenFeature.

providers/configcat/src/main/java/dev/openfeature/contrib/providers/configcat/ConfigCatProviderConfig.java
liran2000: See some [references](https://stackoverflow.com/questions/8881291/why-is-char-preferred-over-string-for-passwords/8889285#8889285).

> Some information, such as Social Security numbers (SSNs) and passwords, is highly sensitive. This information should not be kept for longer than necessary nor where it may be seen, even by administrators. For instance, it should not be sent to log files and its presence should not be detectable through searches. Some transient data may be kept in mutable data structures, such as char arrays, and **cleared immediately after use**. Clearing data structures has reduced effectiveness on typical Java runtime systems as objects are moved **in memory** transparently to the programmer.

providers/configcat/src/main/java/dev/openfeature/contrib/providers/configcat/ConfigCatProvider.java
liran2000: generally it's right, here the evaluate is 1-2 actions, with changed class. When trying to call _configCatClient
            .getValueDetails(Class_ with generic type, it does [not work in a simple way](https://stackoverflow.com/questions/3437897/how-do-i-get-a-class-instance-of-generic-type-t), to make it work make it less clean.

providers/configcat/src/main/java/dev/openfeature/contrib/providers/configcat/ConfigCatProviderConfig.java
laliconfigcat: I'm fine with it if you leave it.
But I still think that this is unnecessary as the SDK holds the SDK key anyway (as it needs it for the whole run) so this doesn't provide any layer of extra security but only some extra lines of code to maintain in the long term.

providers/configcat/src/main/java/dev/openfeature/contrib/providers/configcat/ConfigCatProvider.java
laliconfigcat: I'm fine with the current solution too. 
I was thinking on something like this, it seems much cleaner to me:
 
```
    @Override
    public ProviderEvaluation<Boolean> getBooleanEvaluation(String key, Boolean defaultValue, EvaluationContext ctx) {
        return evaluate(Boolean.class, key, defaultValue, ctx);
    }

    @Override
    public ProviderEvaluation<String> getStringEvaluation(String key, String defaultValue, EvaluationContext ctx) {
        return evaluate(String.class, key, defaultValue, ctx);
    }

    @Override
    public ProviderEvaluation<Double> getDoubleEvaluation(String key, Double defaultValue, EvaluationContext ctx) {
        return evaluate(Double.class, key, defaultValue, ctx);
    }

    @Override
    public ProviderEvaluation<Integer> getIntegerEvaluation(String key, Integer defaultValue, EvaluationContext ctx) {
        return evaluate(Integer.class, key, defaultValue, ctx);
    }

    private <T> ProviderEvaluation<T> evaluate(Class<T> classOfT, String key, T defaultValue, EvaluationContext ctx) {
        User user = ctx == null ? null : dev.openfeature.contrib.providers.configcat.ContextTransformer.transform(ctx);
        EvaluationDetails<T> evaluationDetails = configCatClient
            .getValueDetails(classOfT, key, user, defaultValue);
            .value(evaluationDetails.getValue())
            .variant(evaluationDetails.getVariationId())
            .errorMessage(evaluationDetails.getError())
            .build();
    }
```

providers/configcat/src/main/java/dev/openfeature/contrib/providers/configcat/ConfigCatProvider.java
liran2000: It is not expected to get to these, there are still corner cases from invalid usage which gets to the flow of these checks:

- Constructing a provider directly and evaluate before initialization.
- Evaluating after shutdown.

Then prefer to leave them.

These cases are covered at the [test](https://github.com/open-feature/java-sdk-contrib/pull/521/files#diff-9e3dc32d15e260688a9aa99b1f5a361e71d54ababbf2940570dfacc7ec58ba09R128).

providers/configcat/src/main/java/dev/openfeature/contrib/providers/configcat/ConfigCatProvider.java
liran2000: Done, except _getObjectEvaluation_ which evaluates a String, and returns _Value_.

providers/configcat/src/main/java/dev/openfeature/contrib/providers/configcat/ConfigCatProviderConfig.java
liran2000: In my opinion it is better to leave it, as the SDK implementation can be treated as much as possible as a black-box, without need to know its implementation, or adjust to inner implementation changes. Here the key at the provider config itself is not used, then no need to keep it. I'd leave it for now, thanks.

providers/configcat/src/main/java/dev/openfeature/contrib/providers/configcat/ConfigCatProvider.java
liran2000: Ok, removed the _ClientReady_ event, as not seems it will be correct and necessary in all cases.
The non-ready state is at least temporary until initialization for corner cases (other comment).

providers/configcat/src/main/java/dev/openfeature/contrib/providers/configcat/ConfigCatProvider.java
liran2000: _initialize_ has some more actions like adding hooks, not expected to be called more than once. For this, and less dependent on ConfigCatClient internal implementation, prefer to leave it.

providers/configcat/src/test/java/dev/openfeature/contrib/providers/configcat/ConfigCatProviderTest.java
```diff
@@ -0,0 +1,180 @@
...
+        MutableContext evaluationContext = new MutableContext();
+        evaluationContext.setTargetingKey(userId);
+        evaluationContext.add("Country", country);
+        evaluationContext.add("Email", email);
+        evaluationContext.add(customPropertyKey, customPropertyValue);
+
+        HashMap<String, String > customMap = new HashMap<>();
+        customMap.put(customPropertyKey, customPropertyValue);
+        User expectedUser = User.newBuilder().email(email).country(country).custom(customMap).build(
+            evaluationContext.getTargetingKey());
```
novalisdenahi: The userId variable should be used here.

providers/configcat/src/test/java/dev/openfeature/contrib/providers/configcat/ConfigCatProviderTest.java
```diff
@@ -0,0 +1,180 @@
...
+        // ErrorCode.PROVIDER_NOT_READY should be returned when evaluated via the client
+        assertThrows(ProviderNotReadyError.class, ()-> tempConfigCatProvider.getBooleanEvaluation("fail_not_initialized", false, new ImmutableContext()));
+        assertThrows(ProviderNotReadyError.class, ()-> tempConfigCatProvider.getDoubleEvaluation("fail_not_initialized", 0.1, new ImmutableContext()));
+        assertThrows(ProviderNotReadyError.class, ()-> tempConfigCatProvider.getIntegerEvaluation("fail_not_initialized", 3, new ImmutableContext()));
+        assertThrows(ProviderNotReadyError.class, ()-> tempConfigCatProvider.getObjectEvaluation("fail_not_initialized", null, new ImmutableContext()));
+        assertThrows(ProviderNotReadyError.class, ()-> tempConfigCatProvider.getStringEvaluation("fail_not_initialized", "", new ImmutableContext()));
+    }
+
+    @Test
+    void eventsTest() {
```
novalisdenahi, line 204: There is no assert in this Test. I guess you want to be sure no error has been thrown.  Maybe an `assertDoesNotThrow` can be used.

providers/configcat/src/test/java/dev/openfeature/contrib/providers/configcat/ConfigCatProviderTest.java
```diff
@@ -0,0 +1,180 @@
...
+        assertThrows(ProviderNotReadyError.class, ()-> tempConfigCatProvider.getBooleanEvaluation("fail_not_initialized", false, new ImmutableContext()));
+
+        OpenFeatureAPI.getInstance().setProviderAndWait("tempConfigCatProvider", tempConfigCatProvider);
+
+        assertThrows(GeneralError.class, ()-> tempConfigCatProvider.initialize(null));
+
+        tempConfigCatProvider.shutdown();
+
+        // ErrorCode.PROVIDER_NOT_READY should be returned when evaluated via the client
+        assertThrows(ProviderNotReadyError.class, ()-> tempConfigCatProvider.getBooleanEvaluation("fail_not_initialized", false, new ImmutableContext()));
```
novalisdenahi, line 196: The comment says:  ErrorCode.PROVIDER_NOT_READY should be returned when evaluated via the client. Yet the test checks the provider calls. I assume the answer should be the same if you call the provider or the client. You should fix the comment or the calls in the test.

providers/configcat/src/test/java/dev/openfeature/contrib/providers/configcat/ConfigCatProviderTest.java
```diff
@@ -0,0 +1,180 @@
...
+            evaluationContext).getValue());
+        assertEquals(DOUBLE_FLAG_VALUE, client.getDoubleValue(DOUBLE_FLAG_NAME, 1.1));
+        assertEquals(1.1, client.getDoubleValue("non-existing", 1.1));
+
+        // non-number flag value
+        assertEquals(1.1, client.getDoubleValue(VARIANT_FLAG_NAME, 1.1));
+    }
+
+    @Test
+    void getBooleanEvaluationByUser() {
```
novalisdenahi, line 115: I think the test will fully cover all user-related evaluations if all types of evaluations are called with user/evaluationContext parameters.

I would add the user tests to the corresponding evaluation tests or rework this one to cover all evaluations with user parameters.

providers/configcat/src/main/java/dev/openfeature/contrib/providers/configcat/ConfigCatProvider.java
```diff
@@ -0,0 +1,176 @@
...
+                throw new ProviderNotReadyError(PROVIDER_NOT_YET_INITIALIZED);
+            }
+            throw new GeneralError(UNKNOWN_ERROR);
+        }
+        return getEvaluation(Boolean.class, key, defaultValue, ctx);
+    }
+
+    @Override
+    public ProviderEvaluation<String> getStringEvaluation(String key, String defaultValue, EvaluationContext ctx) {
+        if (!ProviderState.READY.equals(state)) {
```
novalisdenahi: These ProviderState checks could be part of the `getEvaluation` method as well to reduce the code duplication.

providers/configcat/src/main/java/dev/openfeature/contrib/providers/configcat/ConfigCatProvider.java
```diff
@@ -0,0 +1,176 @@
...
+                throw new ProviderNotReadyError(PROVIDER_NOT_YET_INITIALIZED);
+            }
+            throw new GeneralError(UNKNOWN_ERROR);
+        }
+        return getEvaluation(Double.class, key, defaultValue, ctx);
+    }
+
+    @SneakyThrows
+    @Override
+    public ProviderEvaluation<Value> getObjectEvaluation(String key, Value defaultValue, EvaluationContext ctx) {
```
novalisdenahi, line 113: @liran2000 I understand your previous comments, you don't want to mix `getObjectEvaluation` with the other methods but please consider it. It will reduce the code duplication and you get a more clean code. All you need is an extra converter method for the `ProviderEvaluation`. It can be done with an expectedClass parameter or a more simple version, the converter method has 3 parameters: value, variant, error and you do the Value.objectToValue call, when it is needed.

providers/configcat/src/main/java/dev/openfeature/contrib/providers/configcat/ConfigCatProvider.java
```diff
@@ -0,0 +1,176 @@
...
+        if (!ProviderState.READY.equals(state)) {
+            if (ProviderState.NOT_READY.equals(state)) {
+                throw new ProviderNotReadyError(PROVIDER_NOT_YET_INITIALIZED);
+            }
+            throw new GeneralError(UNKNOWN_ERROR);
+        }
+        User user = ctx == null ? null : dev.openfeature.contrib.providers.configcat.ContextTransformer.transform(ctx);
+        EvaluationDetails<String> evaluationDetails = configCatClient
+            .getValueDetails(String.class, key, user, defaultValue.asString());
```
novalisdenahi: The `defaultValue.asString()` is the source of a possible NPE.

providers/configcat/src/main/java/dev/openfeature/contrib/providers/configcat/ConfigCatProvider.java
```diff
@@ -0,0 +1,176 @@
...
+        return ProviderEvaluation.<Value>builder()
+            .value(Value.objectToValue(evaluationDetails.getValue()))
```
novalisdenahi: The `Value.objectToValue(evaluationDetails.getValue())` is the source of a possible NPE. If I understand it correctly. I think a test would be nice to see how the `getObjectEvaluaion` works with a  `null` default value,

providers/configcat/src/main/java/dev/openfeature/contrib/providers/configcat/ConfigCatProvider.java
```diff
@@ -0,0 +1,176 @@
...
+    @SneakyThrows
+    @Override
+    public ProviderEvaluation<Value> getObjectEvaluation(String key, Value defaultValue, EvaluationContext ctx) {
+        if (!ProviderState.READY.equals(state)) {
+            if (ProviderState.NOT_READY.equals(state)) {
+                throw new ProviderNotReadyError(PROVIDER_NOT_YET_INITIALIZED);
+            }
+            throw new GeneralError(UNKNOWN_ERROR);
+        }
+        User user = ctx == null ? null : dev.openfeature.contrib.providers.configcat.ContextTransformer.transform(ctx);
```
novalisdenahi: The whole class path is not necessary here.

providers/configcat/src/main/java/dev/openfeature/contrib/providers/configcat/ConfigCatProvider.java
```diff
@@ -0,0 +1,176 @@
...
+        return ProviderEvaluation.<Value>builder()
+            .value(Value.objectToValue(evaluationDetails.getValue()))
+            .variant(evaluationDetails.getVariationId())
+            .errorMessage(evaluationDetails.getError())
+            .build();
+    }
+
+    private <T> ProviderEvaluation<T> getEvaluation(Class<T> classOfT, String key, T defaultValue,
+           EvaluationContext ctx) {
+        User user = ctx == null ? null : ContextTransformer.transform(ctx);
```
novalisdenahi, line 125: I think the null check can be moved to the `ContextTransformer.transform` method.

novalisdenahi (changes requested):

providers/configcat/src/test/java/dev/openfeature/contrib/providers/configcat/ConfigCatProviderTest.java
liran2000: It is used as the targeting key a few lines above: 
```
evaluationContext.setTargetingKey(userId);
```

providers/configcat/src/test/java/dev/openfeature/contrib/providers/configcat/ConfigCatProviderTest.java
liran2000: It is consistent with other providers tests.
- ErrorCode.PROVIDER_NOT_READY should be returned when evaluated via the client
- ProviderNotReadyError is thrown when evaluated via the - like this test line.

providers/configcat/src/test/java/dev/openfeature/contrib/providers/configcat/ConfigCatProviderTest.java
liran2000: right, to be sure no error has been thrown for code coverage. Not sure if _assertDoesNotThrow_ is needed, as test will fail if an exception occurs.

providers/configcat/src/main/java/dev/openfeature/contrib/providers/configcat/ConfigCatProvider.java
liran2000: not always expected to pass null to a single parameter method, I prefer to leave it like this to be consistent with other providers.

providers/configcat/src/test/java/dev/openfeature/contrib/providers/configcat/ConfigCatProviderTest.java
liran2000: enriched the tests accordingly.

providers/configcat/src/test/java/dev/openfeature/contrib/providers/configcat/ConfigCatProviderTest.java
novalisdenahi: I know. It just looks better if you have an assert in the test. But, that's my opinion.

providers/configcat/src/test/java/dev/openfeature/contrib/providers/configcat/ConfigCatProviderTest.java
novalisdenahi: Yes, and we know the evaluationContext do nothing with the Id. But if we handle like a black-box it could mean you got a different value back when you call the getTargetingKey. The nice thing would be to use the userId here.

providers/configcat/src/test/java/dev/openfeature/contrib/providers/configcat/ConfigCatProviderTest.java
novalisdenahi: I think my previous comment was confusing. What I meant was You can have a getBooleanEvaluationByUser, getDoubleEvaluationByUser ... etc. or just one getEvaluationsByUser test method where Boolean, Double, String, Int and Object calls are tested with user.

Sorry again for the confusion.

novalisdenahi (changes requested):

providers/configcat/src/test/java/dev/openfeature/contrib/providers/configcat/ConfigCatProviderTest.java
liran2000: transform() use it to build the user:
```
userBuilder.build(ctx.getTargetingKey());
```

and the assert is checking that userId as expected:
```
assertEquals(expectedUser.toString(), transformedUser.toString());
```

Am I missing something ? can you be more specific ?

providers/configcat/src/test/java/dev/openfeature/contrib/providers/configcat/ConfigCatProviderTest.java
liran2000: ok done

providers/configcat/src/test/java/dev/openfeature/contrib/providers/configcat/ConfigCatProviderTest.java
novalisdenahi: My point is that in the feature, the `evaluationContext.getTargetingKey` will return not "a" (this is the `userId` value in this test case) but "evaluationContextID:a"  because someone modified the code this way. Your test will run successfully. If the `expectedUser` is built with the original variable `userId`, then the test will fail and show that there is a problem.

This test should prove, the `transform` will create the same ConfigCat User from evaluationContext, what you would create from the same input values.

liran2000: fixed now

novalisdenahi (approved):

laliconfigcat (approved):
Thanks for the contribution!

providers/configcat/README.md
```diff
@@ -0,0 +1,52 @@
...
+ConfigCatProviderConfig configCatProviderConfig = ConfigCatProviderConfig.builder().sdkKey(sdkKey).build();
+configCatProvider = new ConfigCatProvider(configCatProviderConfig);
+OpenFeatureAPI.getInstance().setProviderAndWait(configCatProvider);
+boolean featureEnabled = client.getBooleanValue(FLAG_NAME, false);
+
+MutableContext evaluationContext = new MutableContext();
+evaluationContext.setTargetingKey("csp@matching.com");
+evaluationContext.add("Email", "a@b.com");
+evaluationContext.add("Country", "someCountry");
+featureEnabled = client.getBooleanValue(USERS_FLAG_NAME, false, evaluationContext);
```
toddbaert, line 35: I'd prefer an immutable context example, but either is fine so up to you.

providers/configcat/src/main/java/dev/openfeature/contrib/providers/configcat/ConfigCatProvider.java
```diff
@@ -0,0 +1,156 @@
...
+        super.initialize(evaluationContext);
+        configCatClient = ConfigCatClient.get(configCatProviderConfig.getSdkKey(),
+            configCatProviderConfig.getOptions());
+        configCatProviderConfig.postInit();
+        state = ProviderState.READY;
+        log.info("finished initializing provider, state: {}", state);
+
+        configCatClient.getHooks().addOnConfigChanged(map -> {
+            ProviderEventDetails providerEventDetails = ProviderEventDetails.builder()
+                .flagsChanged(new ArrayList<>(map.keySet()))
```
toddbaert, line 72: It's nice this is supported by ConfigCat!

providers/configcat/src/main/java/dev/openfeature/contrib/providers/configcat/ConfigCatProvider.java
toddbaert: This is a nice optimization.

providers/configcat/README.md
liran2000: I think ImmmutableContext only accept Value type objects.
Also observed at some other providers, ImmmutableContext seemed more fit theoretically, but simpler for consumer to use strings than wrapping each parameter with Value object.

providers/configcat/src/test/java/dev/openfeature/contrib/providers/configcat/ConfigCatProviderTest.java
toddbaert: I think @novalisdenahi is just saying the comment isn't _really_ doing what is says. It's true that in this case, the client will return `NOT_READY` based on your implementation, but that's not the responsibility of the your provider, and it's not the responsibility of this test to test this. It's the SDK that does that mapping of thrown error to error code.

providers/configcat/src/test/java/dev/openfeature/contrib/providers/configcat/ConfigCatProviderTest.java
toddbaert: I tend to agree that `assertDoesNotThrow` adds clarity (some linters even fail tests without some assertion), but this isn't a blocker for me.

providers/configcat/src/main/java/dev/openfeature/contrib/providers/configcat/ConfigCatProvider.java
```diff
@@ -0,0 +1,156 @@
...
+                (T) Value.objectToValue(evaluationDetails.getValue());
+        } else {
+            evaluationDetails = configCatClient
+                .getValueDetails(classOfT, key, user, defaultValue);
+            evaluatedValue = evaluationDetails.getValue();
+        }
+        return ProviderEvaluation.<T>builder()
+            .value(evaluatedValue)
+            .variant(evaluationDetails.getVariationId())
+            .errorMessage(evaluationDetails.getError())
```
toddbaert:
```suggestion
```
We have actually been talking about removing this field entirely from the `ProviderEvaluation`, and keeping it only on the `FlagEvaluationDetails`, forcing users to return a successful evaluation OR throw. I don't think it makes sense to ever populate this on a normal return.

If you want a "catch all" for errors, I would recommend wrapping this whole block after your validation in a try/catch, and throwing a `GeneralError`. This will also help if there's some NPE in this block or something like that (you could add a log statement) set the cause (wrap the error).

toddbaert (approved):
Approving, but I left a few comments. I think my only real concern is [this](https://github.com/open-feature/java-sdk-contrib/pull/521/files#r1394739184), which I would recommend changing.

providers/configcat/src/main/java/dev/openfeature/contrib/providers/configcat/ConfigCatProvider.java
liran2000: ok, removed.

liran2000:
Updated according to the comments.

toddbaert (approved):
As always, thanks to @liran2000. Also @laliconfigcat and @novalisdenahi for the review!

PR #567, opened 2023-12-04
fix: edge cases with flagd targeting

toddbaert:
* coerce booleans to stringified versions for matching
* implement new gherkin tests ([not yet released](https://github.com/open-feature/flagd-testbed/pull/85), will block PR for now)


providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolver.java
```diff
@@ -174,7 +174,7 @@ private <T> ProviderEvaluation<T> resolve(Class<T> type, String key,
                     resolvedVariant = flag.getDefaultVariant();
                     reason = Reason.DEFAULT.toString();
                 } else {
-                    resolvedVariant = jsonResolved;
+                    resolvedVariant = jsonResolved.toString(); // convert to string to support shorthand
```
toddbaert, line 177: This is the only functional change.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolver.java
```diff
@@ -198,7 +198,7 @@ private <T> ProviderEvaluation<T> resolve(Class<T> type, String key,
             // if this is a double and we are trying to resolve an integer, convert
             value = ((Double) value).intValue();
         }
-        if (!type.isAssignableFrom(value.getClass()) || !(resolvedVariant instanceof String)) {
+        if (!type.isAssignableFrom(value.getClass())) {
```
Kavindu-Dodan, line 201: What would happen when resolvedVarient is not a string. reason is at line 209, we have a casting to string. This could fail for non string casting exception 🤔 

I think we might need explicit handling to perform this

Kavindu-Dodan (approved):
I have [this](https://github.com/open-feature/java-sdk-contrib/pull/567/files#r1416188411) concern, otherwise looks fine

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolver.java
toddbaert: I think you have a point.

We could expand the caught exception to include this. In that case, it'd still throw a `ParseError`, which I think is OK (you'd have incorrectly configured a flag in that case).

Kavindu-Dodan: [This](https://github.com/open-feature/java-sdk-contrib/blob/a2d319964f3ade5bfd6a05d587e0f403c9c09a87/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolver.java#L177-L178) logic which transforms variant to string solves my concern.

Along with this, I think we can do the following improvement 

 `final Object resolvedVariant` -> ` final String resolvedVariant`
 
 And avoid the cast. at line 209

Kavindu-Dodan: > you'd have incorrectly configured a flag in that case

This will fail at json parsing time. so we don't have to worry here. I think this resolved once we update the variable type

toddbaert: Excellent observation! I have made this change.

Kavindu-Dodan: Great 🤝

PR #367, opened 2023-07-25
feat: [go-feature-flag] Support flag metadata

thomaspoignant:
## This PR
- Support flag metadata in the java provider for GO Feature Flag

PR #572, opened 2023-12-06
fix: [go-feature-flag] Make flag metadata optional

itchyny:
## This PR

- fixes a `NullPointerException` when receive data without flag metadata from go-feature-flag.

### Related Issues

Fixes https://github.com/open-feature/java-sdk-contrib/pull/367#discussion_r1377197199.

### Notes

The metadata in the flag configuration is an optional field.
> metadata
> (optional)	
> This field allows adding a wealth of information about a particular feature flag, such as a configuration URL or the originating Jira issue.
> https://gofeatureflag.org/docs/configure_flag/flag_format

Also see the following `omitempty` tag which means that the field is not included in the response JSON if empty.
> 	Metadata      map[string]interface{} `json:"metadata,omitempty"`
> https://github.com/thomaspoignant/go-feature-flag/blob/ba95173c873f0bc49cabfe2911f04289d154a04b/internal/flagstate/flag_state.go#L16


thomaspoignant (approved):
Good catch @itchyny, this is a bug that I did not catch before.

thomaspoignant:
Version `v0.2.16` has been released and will be available in maven soon.

PR #581, opened 2023-12-08
fix: flagd caching

Kavindu-Dodan:
## This PR

- Fix flagd cache handling : Caching never worked as enum.valueOf(String) always throwed and with no exception loggig we never understood this happening
- Improves logic and adds testing


providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/grpc/cache/Cache.java
```diff
@@ -1,41 +1,42 @@
...
-                this.store = Collections.synchronizedMap(new LRUMap<>(maxCacheSize));
+    public Cache(final String forType, int maxCacheSize) {
+        if (DISABLED.getValue().equals(forType)) {
+            enabled = false;
+        } else if (LRU.getValue().equals(forType)) {
+            enabled = true;
+            this.store = Collections.synchronizedMap(new LRUMap<>(maxCacheSize));
+        } else {
+            enabled = false;
+            log.info(String.format("Unsupported cache type %s, continuing without cache", forType));
```
toddbaert:
```suggestion
            log.warn(String.format("Unsupported cache type %s, continuing without cache", forType));
```

This feels like a warn to me.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/grpc/cache/Cache.java
```diff
@@ -1,41 +1,42 @@
...
-            return;
-        }
-
-        switch (type) {
-            case DISABLED:
-                return;
-            case LRU:
-            default:
-                this.store = Collections.synchronizedMap(new LRUMap<>(maxCacheSize));
+    public Cache(final String forType, int maxCacheSize) {
```
toddbaert, line 30: I'm guessing you went for a string here because we need to support getting this value from an env var (which is always a string) and it makes more sense to convert it to the enum here rather than ahead of time before this constructor is called?

toddbaert (approved):

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/grpc/cache/Cache.java
Kavindu-Dodan: yeah, I agree on this

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/grpc/cache/Cache.java
Kavindu-Dodan: > I'm guessing you went for a string here because we need to support getting this value from an env var (which is always a string) and it makes more sense to convert it to the enum here rather than ahead of time before this constructor is called?

Before this, we had`CacheFactory.java` which performed a String to Enum parsing. Then we had another switch condition to select the caching implementation based on the parsing. I think these two layers added unwanted complexity - for example, the bug was due to incorrect parsing, hidden in the factory layer.

Given we simply have two possible cache types (disabled & lru) I tried to remove the factory layers and have comparison logic in Caching itself. 

And like you said, we always expect a string value, so having a simple comparison here is sufficient.

PR #473, opened 2023-10-06
feat: flagd in process offline mode

Kavindu-Dodan:
## This PR

Introduce offline mode for in-process resolver.

The offline mode works with flags read from a local file,

```java
FlagdProvider flagdProvider = new FlagdProvider(
        FlagdOptions.builder()
                .resolverType(Config.Evaluator.IN_PROCESS)
                .offlineFlagSourcePath("PATH")
                .build());
```

This implementation is kept minimal such that the file is read once at the initialization. No attempt is made to detect file changes. However, the `Connector` contract is flexible for such improvements if we see the need

PR #614, opened 2024-01-03
feat: flagd file polling for offline mode

Kavindu-Dodan:
## This PR

Enhance flagd in-process offline mode introduced through https://github.com/open-feature/java-sdk-contrib/pull/473 by adding polling support and align with the change proposed at flagd js-sdk (https://github.com/open-feature/js-sdk-contrib/pull/708)

Polling happens at intervals of 5 seconds and runs concurrently. If an update is observed, connect offers the newest content to the storage through the connector contract. 

### Note on polling 

The reason to use polling is the known limitation of Java WatchService when it comes to non-local file systems. This blog post [^1] extensively describes the problem. Besides, the offline mode is intended for limited use cases such as testing, local development efforts or POC deployments where the impact of polling is minimal. 

[^1]: https://blog.arkey.fr/2019/09/13/watchservice-and-bind-mount/


providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/file/FileConnector.java
```diff
@@ -9,36 +9,73 @@
...
 
+    private static final int POLL_INTERVAL_MS = 5000;
+
     private final String flagSourcePath;
     private final BlockingQueue<StreamPayload> queue = new LinkedBlockingQueue<>(1);
+    private boolean shutdown = false;
 
     public FileConnector(final String flagSourcePath) {
         this.flagSourcePath = flagSourcePath;
     }
```
thisthat, line 35: we could also add another constructor to customize the polling interval
```
public FileConnector(final String flagSourcePath, final int poll_interval_ms) 
```

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/file/FileConnectorTest.java
```diff
@@ -39,12 +44,82 @@ void readAndExposeFeatureFlagsFromSource() throws IOException {
...
+                "        \"off\": false\n" +
+                "      },\n" +
+                "      \"defaultVariant\": \"off\"\n" +
+                "    }\n" +
+                "  }\n" +
+                "}";
+
+        // given
+        final Path updPath = Paths.get(getResourcePath(UPDATABLE_FILE));
+        Files.write(updPath, initial.getBytes(), StandardOpenOption.WRITE);
```
thisthat: I would use a tmp file via [File. createTempFile()](https://docs.oracle.com/javase/8/docs/api/java/io/File.html#createTempFile-java.lang.String-java.lang.String-) so there is no side-effects on the resources

Kavindu-Dodan: I already attempted this but failed when dealing with github workflow. I will try this again and revert the attempt fails

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/file/FileConnector.java
Kavindu-Dodan: This is useful only if we plan to expose the polling interval as a configuration option. However, I think this is not a priority right now hence I think we can stick with hardcoded value (5000ms) till we see the need.

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/file/FileConnectorTest.java
Kavindu-Dodan: No luck, for some reason temporary file writes within gh action CI doesn't complete and results in timeout of this test. I will keep this test as it is for now. Beside, you can see my previous attempt to use temp files in this commit - https://github.com/open-feature/java-sdk-contrib/pull/614/commits/f548d750f7e836453b579ad85bdbb8ad8e3a2418

Kavindu-Dodan: Update - I disabled the file write test as it was unstable when running on the CI. However, the core functionality works fine in K8s setup and in local deployment.

thisthat: oh, I was aware of this problem in GH Action 😢 
Then, I am fine with this implementation :)

thisthat (approved):

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/file/FileConnectorTest.java
Kavindu-Dodan: @thisthat  I think we don't have much control over the underlying file system of the GH Action (per my search on details). Specially the write speeds and update visibility.

thisthat: Maybe is due to permissions on the job: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

Kavindu-Dodan: I checked on this but this is on repository related content (commits, releases etc.)

toddbaert: > @thisthat I think we don't have much control over the underlying file system of the GH Action (per my search on details). Specially the write speeds and update visibility.

We found this before when Sky was trying to tests file changes. We could never get this working as expected in GH actions.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/file/FileConnector.java
```diff
@@ -9,36 +9,73 @@
...
 
     /**
-     * Initialize file connector. Reads content of the provided source file and offer it through queue.
+     * Initialize file connector. Reads file content, poll for changes and offer content through the queue.
      */
     public void init() throws IOException {
-        final String flagData = new String(Files.readAllBytes(Paths.get(flagSourcePath)), StandardCharsets.UTF_8);
+        Thread watcherT = new Thread(() -> {
```
toddbaert, line 41: I could be missing it, but taking a quick look, it seems like we don't have the necessary synchronization here, or in [FlagStore](https://github.com/open-feature/java-sdk-contrib/blob/main/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/FlagStore.java) to prevent 2 threads from calling this at nearly the same time and starting multiple instances of this thread.

Maybe we can implement something like that in the provider, or in the `FlagStore` so it doesn't need to be implemented in every `Connector`?

toddbaert (approved):
Approving with [this question](https://github.com/open-feature/java-sdk-contrib/pull/614/files#r1445084437). I will let you consider it and decide if you think it's something we should worry about in this PR.

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/file/FileConnectorTest.java
Kavindu-Dodan: @toddbaert  😭 found it the hard way. Thank you for confirming this.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/file/FileConnector.java
Kavindu-Dodan: Good point. I didn't consider a race condition when implementing this. The reason was that I did not expect the initializer to be called concurrently. But a quick walkthrough on potential usage patterns, there could be race conditions if the same provider is registered with multiple names. I think I would like to revisit this topic in a dedicated PR.

Kavindu-Dodan:
> Approving with [this question](https://github.com/open-feature/java-sdk-contrib/pull/614/files#r1445084437). I will let you consider it and decide if you think it's something we should worry about in this PR.

Thanks for the review. Regarding the concurrency concern, I will address it in a dedicated PR. As described [here](https://github.com/open-feature/java-sdk-contrib/pull/614#discussion_r1445318449) only possibility for concurrent issues is if the same provider get registered multiple times (which is allowed in spec). And I think we need to check this edge case in other lang. implementations as well.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/file/FileConnector.java
Kavindu-Dodan: I created follow-up tickets for Go (https://github.com/open-feature/go-sdk-contrib/issues/417) & Java (https://github.com/open-feature/java-sdk-contrib/issues/623)

PR #635, opened 2024-01-16
feat: synchronize initialization and shutdown

### LINKED ISSUE #623: [flagd] validate race conditions of in-process provider
As suggested by [this](https://github.com/open-feature/java-sdk-contrib/pull/614#discussion_r1445084437) comment, our concurrency logic needs to be validated for race conditions. Though we expect initialization to be called once, there could be edge cases such as multiple provider registrations which could invoke unwanted initialization attempts.

**Acceptance Criteria**

In-process provider implementation is validated for potential race conditions and fixed.

Kavindu-Dodan:
## This PR

Fixes #623 by adding synchronization on init & shutdown along with initialize state variable to avoid multiple initializations and shutdowns

### DIFF
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
@@ -29,6 +29,7 @@ public class FlagdProvider extends EventProvider implements FeatureProvider {
     private final ReadWriteLock lock = new ReentrantReadWriteLock();
     private final Resolver flagResolver;
     private ProviderState state = ProviderState.NOT_READY;
+    private boolean initialized = false;
 
     private EvaluationContext evaluationContext;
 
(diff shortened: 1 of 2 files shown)


toddbaert (approved):
Very straightforward.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
```diff
@@ -63,15 +64,25 @@ public FlagdProvider(final FlagdOptions options) {
...
+
         this.evaluationContext = evaluationContext;
         this.flagResolver.init();
+        this.initialized = true;
     }
 
     @Override
-    public void shutdown() {
+    public synchronized void shutdown() {
+        if (!initialized){
```
toddbaert:
```suggestion
        if (!initialized) {
```

I think this is your lint failure.

beeme1mr (approved):

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
Kavindu-Dodan: yeah, fixed with the latest rebase :)

PR #641, opened 2024-01-18
feat: Add Statsig provider

liran2000:
[Readme](https://github.com/open-feature/java-sdk-contrib/blob/2d5cb5103fe6416feef28fe52b7134aade908307/providers/statsig/README.md#unofficial-statsig-openfeature-provider-for-java) describes the provider.

@MovieStoreGuy @jasonwzm @xinlili-statsig @toddbaert you are welcome to review.


providers/statsig/src/main/java/dev/openfeature/contrib/providers/statsig/ContextTransformer.java
```diff
@@ -0,0 +1,63 @@
...
+import dev.openfeature.sdk.Structure;
+import dev.openfeature.sdk.Value;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Transformer from OpenFeature context to statsig User.
+ */
+public class ContextTransformer {
```
Kavindu-Dodan: Can be package private and final

providers/statsig/src/main/java/dev/openfeature/contrib/providers/statsig/ContextTransformer.java
```diff
@@ -0,0 +1,63 @@
...
+
+    public static final String CONTEXT_APP_VERSION = "appVersion";
+    public static final String CONTEXT_COUNTRY = "country";
+    public static final String CONTEXT_EMAIL = "email";
+    public static final String CONTEXT_IP = "ip";
+    public static final String CONTEXT_LOCALE = "locale";
+    public static final String CONTEXT_USER_AGENT = "userAgent";
+    public static final String PRIVATE_ATTRIBUTES = "privateAttributes";
+
+    protected static StatsigUser transform(EvaluationContext ctx) {
```
Kavindu-Dodan: can be package private

providers/statsig/src/main/java/dev/openfeature/contrib/providers/statsig/StatsigProvider.java
```diff
@@ -0,0 +1,262 @@
...
+     * @param evaluationContext evaluation context
+     * @throws Exception on error
+     */
+    @Override
+    public void initialize(EvaluationContext evaluationContext) throws Exception {
+        boolean initialized = isInitialized.getAndSet(true);
+        if (initialized) {
+            throw new GeneralError("already initialized");
+        }
+        super.initialize(evaluationContext);
```
Kavindu-Dodan: Calling super is not needed

providers/statsig/src/main/java/dev/openfeature/contrib/providers/statsig/StatsigProvider.java
```diff
@@ -0,0 +1,262 @@
...
+    /**
+     * Initialize the provider.
+     * @param evaluationContext evaluation context
+     * @throws Exception on error
+     */
+    @Override
+    public void initialize(EvaluationContext evaluationContext) throws Exception {
+        boolean initialized = isInitialized.getAndSet(true);
+        if (initialized) {
+            throw new GeneralError("already initialized");
```
Kavindu-Dodan: Don't have to throw, we can return instead

providers/statsig/src/main/java/dev/openfeature/contrib/providers/statsig/StatsigProvider.java
```diff
@@ -0,0 +1,331 @@
...
+    public Metadata getMetadata() {
+        return () -> NAME;
+    }
+
+    /**
+     * Feature config, as required for evaluation.
+     */
+    @AllArgsConstructor
+    @Getter
+    public static class FeatureConfig {
```
Kavindu-Dodan: I think this fits after well at the end of the file, unmixed from public methods

providers/statsig/src/main/java/dev/openfeature/contrib/providers/statsig/StatsigProvider.java
```diff
@@ -0,0 +1,331 @@
...
+            throw new IllegalArgumentException("feature config name is missing");
+        }
+        String name = nameValue.asString();
+        return new FeatureConfig(type, name);
+    }
+
+    @SneakyThrows
+    @Override
+    public void shutdown() {
+        super.shutdown();
```
Kavindu-Dodan: can avoid super as this is noop

providers/statsig/src/main/java/dev/openfeature/contrib/providers/statsig/StatsigProvider.java
```diff
@@ -0,0 +1,331 @@
...
+        private String name;
+    }
+
+    @Override
+    public ProviderEvaluation<Boolean> getBooleanEvaluation(String key, Boolean defaultValue, EvaluationContext ctx) {
+        if (!ProviderState.READY.equals(state)) {
+            if (ProviderState.NOT_READY.equals(state)) {
+                throw new ProviderNotReadyError(PROVIDER_NOT_YET_INITIALIZED);
+            }
+            throw new GeneralError(UNKNOWN_ERROR);
```
Kavindu-Dodan: Not sure whether these checks are needed. If the underlying provider is shutdown, then we shouldn't get calls to evaluations

liran2000: According to spec Requirement [2.4.5](https://github.com/open-feature/spec/blob/main/specification/sections/02-providers.md#requirement-245):
> The provider SHOULD indicate an error if flag resolution is attempted before the provider is ready.
---            
> If the underlying provider is shutdown, then we shouldn't get calls to evaluations

Right. This is not expected that consumer will call evaluations, the checks are there such that if consumer calls evaluation after provider shutdown, it will get some indication, I think it is better to verify it, similar as methods do checks for invalid parameters.

Kavindu-Dodan: Good point

providers/statsig/src/main/java/dev/openfeature/contrib/providers/statsig/StatsigProvider.java
```diff
@@ -0,0 +1,320 @@
...
+    public ProviderEvaluation<Boolean> getBooleanEvaluation(String key, Boolean defaultValue, EvaluationContext ctx) {
+        verifyEvaluation();
+        StatsigUser user = ContextTransformer.transform(ctx);
+        Future<Boolean> featureOn = Statsig.checkGateAsync(user, key);
+        try {
+            Boolean evaluatedValue = featureOn.get();
+            return ProviderEvaluation.<Boolean>builder()
+                .value(evaluatedValue)
+                .build();
+        } catch (Exception e) {
```
xinlili-statsig: Statsig SDK do not throw if there is an exception instead we log, this is to make sure we don't crash the app. So having a try catch here is not too valuable. 
Instead, I will suggest to set CustomLogger in StatsigOptions, so when exception happen, it can log to the logger https://github.com/statsig-io/java-server-sdk/blob/c3c18a111d464384e85100b302e8561c521baea9/src/main/kotlin/com/statsig/sdk/StatsigOptions.kt#L48

providers/statsig/src/main/java/dev/openfeature/contrib/providers/statsig/StatsigProvider.java
```diff
@@ -0,0 +1,320 @@
...
+    @Getter
+    public static class FeatureConfig {
+
+        /**
+         * Type.
+         *  CONFIG: Dynamic Config
+         *  LAYER: Layer
+         */
+        public enum Type {
+            CONFIG, LAYER
```
xinlili-statsig, line 290: Nit: in statsig , we have concept of experiment, but down the line, it's essential same as dynamicConfig. So consider add experiment

providers/statsig/src/main/java/dev/openfeature/contrib/providers/statsig/StatsigProvider.java
beeme1mr: The OpenFeature SDK will also catch any unhandled expectations and return the default value. I agree with @xinlili-statsig suggestion since the catch block is throwing a general error.

providers/statsig/src/main/java/dev/openfeature/contrib/providers/statsig/StatsigProvider.java
liran2000: According to [Statsig docs](https://docs.statsig.com/server/javaSdk#getting-an-layerexperiment):
> We offer two APIs, but we recommend the use of [layers](https://docs.statsig.com/layers) to enable quicker iterations with parameter reuse.

I thought to support here only the recommended use.

providers/statsig/src/main/java/dev/openfeature/contrib/providers/statsig/StatsigProvider.java
liran2000: updated.

providers/statsig/src/main/java/dev/openfeature/contrib/providers/statsig/StatsigProvider.java
```diff
@@ -0,0 +1,296 @@
...
+ */
+@Slf4j
+public class StatsigProvider extends EventProvider {
+
+    @Getter
+    private static final String NAME = "Statsig";
+
+    public static final String PROVIDER_NOT_YET_INITIALIZED = "provider not yet initialized";
+    public static final String UNKNOWN_ERROR = "unknown error";
+    public static final String FEATURE_CONFIG_KEY = "feature_config";
```
Kavindu-Dodan: should these be public ?

providers/statsig/src/main/java/dev/openfeature/contrib/providers/statsig/StatsigProvider.java
```diff
@@ -0,0 +1,296 @@
...
+
+    @NotNull
+    private static FeatureConfig parseFeatureConfig(EvaluationContext ctx) {
+        Value featureConfigValue = ctx.getValue(FEATURE_CONFIG_KEY);
+        if (featureConfigValue == null) {
+            throw new IllegalArgumentException("feature config not found at evaluation context.");
+        }
+        if (!featureConfigValue.isStructure()) {
+            throw new IllegalArgumentException("feature config is not a structure.");
+        }
```
Kavindu-Dodan, line 239: Probably need to document these exceptions in readme so users add the required entries.

providers/statsig/src/main/java/dev/openfeature/contrib/providers/statsig/StatsigProvider.java
liran2000: updated

providers/statsig/src/main/java/dev/openfeature/contrib/providers/statsig/StatsigProvider.java
liran2000: It is documented in readme:
> String/Integer/Double evaluations evaluation gets [Dynamic config](https://docs.statsig.com/server/javaSdk#reading-a-dynamic-config) or [Layer](https://docs.statsig.com/server/javaSdk#getting-an-layerexperiment) evaluation.
  As the key represents an inner attribute, feature config is required as a parameter with data needed for evaluation.
  For an example of dynamic config of product alias, need to differentiate between dynamic config or layer, and the dynamic config name.

providers/statsig/src/main/java/dev/openfeature/contrib/providers/statsig/ContextTransformer.java
```diff
@@ -0,0 +1,65 @@
...
+    public static final String CONTEXT_APP_VERSION = "appVersion";
+    public static final String CONTEXT_COUNTRY = "country";
+    public static final String CONTEXT_EMAIL = "email";
+    public static final String CONTEXT_IP = "ip";
+    public static final String CONTEXT_LOCALE = "locale";
+    public static final String CONTEXT_USER_AGENT = "userAgent";
+    public static final String CONTEXT_PRIVATE_ATTRIBUTES = "privateAttributes";
+
+    static StatsigUser transform(EvaluationContext ctx) {
+        StatsigUser user = new StatsigUser(ctx.getTargetingKey());
```
toddbaert, line 28: Is it OK if the arg to `new StatsigUser()` here is null? If not you might want to throw invalid context.

toddbaert (dismissed):
LGTM @liran2000 , thanks again.

@xinlili-statsig are you interested in being added to the [component_owners](https://github.com/open-feature/java-sdk-contrib/blob/main/.github/component_owners.yml) for this component? It will mean you are added to PRs that impact it. Note you will have to join the OpenFeature org in that case.

@liran2000 please add yourself to https://github.com/open-feature/java-sdk-contrib/blob/main/.github/component_owners.yml for this at a minimum.

Also, you may be interested in [this issue](https://github.com/open-feature/spec/issues/238) since you've authored quite a few providers.

providers/statsig/src/main/java/dev/openfeature/contrib/providers/statsig/ContextTransformer.java
liran2000: done

providers/statsig/src/main/java/dev/openfeature/contrib/providers/statsig/ContextTransformer.java
```diff
@@ -0,0 +1,69 @@
+package dev.openfeature.contrib.providers.statsig;
+
+import com.statsig.sdk.StatsigUser;
+import dev.openfeature.sdk.EvaluationContext;
+import dev.openfeature.sdk.Structure;
+import dev.openfeature.sdk.Value;
+import dev.openfeature.sdk.exceptions.InvalidContextError;
```
toddbaert:
```suggestion
import dev.openfeature.sdk.exceptions.TargetingKeyMissingError;
```

providers/statsig/src/main/java/dev/openfeature/contrib/providers/statsig/ContextTransformer.java
```diff
@@ -0,0 +1,69 @@
...
+    public static final String CONTEXT_COUNTRY = "country";
+    public static final String CONTEXT_EMAIL = "email";
+    public static final String CONTEXT_IP = "ip";
+    public static final String CONTEXT_LOCALE = "locale";
+    public static final String CONTEXT_USER_AGENT = "userAgent";
+    public static final String CONTEXT_PRIVATE_ATTRIBUTES = "privateAttributes";
+
+    static StatsigUser transform(EvaluationContext ctx) {
+        if (ctx.getTargetingKey() == null) {
+            throw new InvalidContextError("targeting key is missing");
```
toddbaert:
```suggestion
            throw new TargetingKeyMissingError("targeting key is required.");
```

toddbaert (approved):
Hey @liran2000 I led you wrong, I said to use `InvalidContextError` but we have `TargetingKeyMissingError`, so I recommend that... so I made that change. Sorry!

toddbaert:
@Kavindu-Dodan when you have a moment, please review and feel free to merge if it looks OK.

xinlili-statsig (approved):
Overall lgtm!

Kavindu-Dodan (approved):

Kavindu-Dodan:
Merging as this is approved by @xinlili-statsig. Also, looks good in my view and there's good test coverage.

PR #644, opened 2024-01-20
feat: Adopt ConfigCat v9

liran2000:
Adopt [ConfigCat v9](https://github.com/configcat/java-sdk/releases/tag/v9.0.0).

Adjusted test features config file according to [docs](https://configcat.com/docs/sdk-reference/java/#2-complex-full-featured-structure) (not via CLI which required an account).

@novalisdenahi / @laliconfigcat can review.


providers/configcat/pom.xml
```diff
@@ -10,7 +10,7 @@
     </parent>
     <groupId>dev.openfeature.contrib.providers</groupId>
     <artifactId>configcat</artifactId>
-    <version>0.0.2</version> <!--x-release-please-version -->
+    <version>0.1.0</version> <!--x-release-please-version -->
```
novalisdenahi, line 13: I think this version should be updated in the README.md as well.

providers/configcat/src/test/resources/features.json
```diff
@@ -1,88 +1,217 @@
...
+              "u": {
+                "a": "Identifier",
+                "c": 2,
+                "l": [
+                  "@matching.com"
+                ]
+              }
+            }
+          ],
+          "p": [
```
novalisdenahi: You don't need a percentage options array here, but you need a return value for the conditions.  This is true for the other flags in this JSON as well, where a percentage array was added.

The `userIdMatching` should look like this:

```
"userIdMatching": {
      "t": 0,
      "r": [
        {
          "c": [
            {
              "u": {
                "a": "Identifier",
                "c": 2,
                "l": [
                  "@matching.com"
                ]
              }
            }
          ],
          "s": {
              "v": {
                "b": true
              },
              "i": "bcfb84a7"
            }
        }
      ],
      "v": {
        "b": false
      },
      "i": "430bded3"
    }
```

providers/configcat/src/test/java/dev/openfeature/contrib/providers/configcat/ConfigCatProviderTest.java
novalisdenahi, line 1: The `shouldThrowIfNotInitialized` test ran successfully, but if you check the logs, you can see that an unwanted error has been thrown.

This is because the v9.0.0 added some extra SDK key validation to the code. 

The relevant release notes:
> Throw IllegalArgumentException when the SDK key passed to ConfigCatClient.get is in an invalid format (unless the client is set up to use local-only flag override behaviour).

Please use a valid formatted SDK key, for example:
`configcat-sdk-1/TEST_KEY-0123456789012/1234567890123456789012`

novalisdenahi (changes requested):

providers/configcat/pom.xml
liran2000: At the readme the version is under [release-please](https://github.com/open-feature/java-sdk-contrib/blob/main/providers/configcat/README.md?plain=1#L7) block and should be auto update when released.

novalisdenahi (approved):

beeme1mr (approved):

toddbaert (approved):

PR #647, opened 2024-01-23
chore: flagd add offline flag source path support through env variables

### LINKED ISSUE #646: [flagd] offline mode file path from env variable
Similar to go-sdk provider, we should allow to configure java provider's offline flag source path from an environment variable.

For consistency use same variable name - `FLAGD_OFFLINE_FLAG_SOURCE_PATH`

**AC**

Env variable is available and documentation updated accordingly 

See - _Originally posted by @Kavindu-Dodan in https://github.com/open-feature/go-sdk-contrib/pull/421#discussion_r1463811416_

Kavindu-Dodan:
## This PR

Fixes #646 by adding `FLAGD_OFFLINE_FLAG_SOURCE_PATH` env variable to set offline mode's flag source path.

Logic is improved to handle null and documentation was updated to have consistency across languages.

### DIFF
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdOptions.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdOptions.java
@@ -5,8 +5,6 @@ import io.opentelemetry.api.OpenTelemetry;
 import lombok.Builder;
 import lombok.Getter;
 
-import javax.annotation.Nonnull;
-
 import static dev.openfeature.contrib.providers.flagd.Config.BASE_EVENT_STREAM_RETRY_BACKOFF_MS;
 import static dev.openfeature.contrib.providers.flagd.Config.BASE_EVENT_STREAM_RETRY_BACKOFF_MS_ENV_VAR_NAME;
 import static dev.openfeature.contrib.providers.flagd.Config.CACHE_ENV_VAR_NAME;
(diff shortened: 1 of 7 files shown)


beeme1mr (approved):

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdOptions.java
```diff
@@ -135,31 +130,14 @@ public class FlagdOptions {
...
-        }
-
         /**
          * Enable OpenTelemetry instance extraction from GlobalOpenTelemetry. Note that, this is only useful if global
          * configurations are registered.
          */
         public FlagdOptionsBuilder withGlobalTelemetry(final boolean b) {
-            this.openTelemetry = GlobalOpenTelemetry.get();
-
+            if (b) {
```
thisthat, line 138: 🤕 how did we miss seeing this? nice catch :shipit:

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolver.java
```diff
@@ -46,9 +46,10 @@ public class InProcessResolver implements Resolver {
      * Initialize an in-process resolver.
      */
     public InProcessResolver(FlagdOptions options, Consumer<ProviderState> stateConsumer) {
-        final Connector connector = options.isOffline()
-                ? new FileConnector(options.getOfflineFlagSourcePath())
-                : new GrpcStreamConnector(options);
+        final Connector connector =
```
thisthat: maybe we could add some tests in InProcessResolverTest.java to check that the right connector is selected but def not needed as part of this PR :)

thisthat (approved):

toddbaert (approved):

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdOptions.java
Kavindu-Dodan: I saw this when cleaning the builder logic 😿

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolver.java
Kavindu-Dodan: Done, added a test with commit https://github.com/open-feature/java-sdk-contrib/pull/647/commits/f2c331338e96047723f49d3cea9e809301e1d6d4

PR #681, opened 2024-02-14
fix: null exception when no options used

Kavindu-Dodan:
## This PR

Noticed several pipeline failures and seems this is due to the tests fixing in the PR.

@toddbaert  identified this to be an issue with default value of `StatsigOptions`

### DIFF
--- a/providers/statsig/src/main/java/dev/openfeature/contrib/providers/statsig/StatsigProviderConfig.java
+++ b/providers/statsig/src/main/java/dev/openfeature/contrib/providers/statsig/StatsigProviderConfig.java
@@ -12,7 +12,8 @@ import lombok.Getter;
 @Builder
 public class StatsigProviderConfig {
 
-    private StatsigOptions options;
+    @Builder.Default
+    private StatsigOptions options = new StatsigOptions();
 
     // Only holding temporary for initialization
     private String sdkKey;


liran2000 (approved):

providers/statsig/src/test/java/dev/openfeature/contrib/providers/statsig/StatsigProviderTest.java
```diff
@@ -278,19 +278,15 @@ void getBooleanEvaluationByProperties() {
...
         StatsigProviderConfig statsigProviderConfig = StatsigProviderConfig.builder().sdkKey("test").build();
         StatsigProvider tempstatsigProvider = new StatsigProvider(statsigProviderConfig);
 
         assertThrows(ProviderNotReadyError.class, ()-> tempstatsigProvider.getBooleanEvaluation(
         "fail_not_initialized", false, new ImmutableContext()));
-
-        OpenFeatureAPI.getInstance().setProviderAndWait("tempstatsigProvider", tempstatsigProvider);
     }
 
-    @SneakyThrows
```
toddbaert: I think this actually hides the main issue (though you're right, its not needed to do this here). 

The main issue is that options can't be null, and the statsig sdk enforces that.

I pushed a change to the config file which I think will fix the actual issue.

Just to prove it, if you remove your changes in this file, it will still pass with my changes to config.

toddbaert: I think if we're not going to add a brand new test to cover this failure case though, you might want to revert your test changes (because we do indeed need to default these options or the same error is thrown in production use).

Kavindu-Dodan: Thanks, yes you are correct. I will revert the changes keeping configuration relates change

toddbaert (approved):

PR #683, opened 2024-02-14
feat!: use new eval/sync protos (requires flagd v0.7.3+)

toddbaert:
* uses newest protos (eval and sync)
* implements 3 new e2e test suites (targetingKey, time-based, and init-error)
* validation for targeting rules using new schema

I made one functional change here which only WARNS instead of throwing if the flag config is invalid. Happy to go either way on this.


providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/model/FlagParser.java
```diff
@@ -40,37 +39,39 @@ private FlagParser() {
...
 
                 if (!validationMessages.isEmpty()) {
-                    throw new IllegalArgumentException(
-                            String.format("Failed to parse configurations. %d validation error(s) reported.",
-                                    validationMessages.size()));
+                    String message = String.format("Invalid flag configuration: %s", validationMessages.toArray());
+                    log.warn(message);
+                    if (throwIfInvalid) {
+                        throw new IllegalArgumentException(message);
+                    }
```
toddbaert, line 74: I no longer throw if things are invalid... instead just log. This is consistent with flagd itself, but I'm happy to talk about syncing this behavior across providers/flagd. Interested in what others think. I've left all the validation in, and in fact it can be enabled to throw instead of warn with a single bool.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/FlagStore.java
```diff
@@ -32,9 +32,16 @@ public class FlagStore implements Storage {
     private final Map<String, FeatureFlag> flags = new HashMap<>();
 
     private final Connector connector;
+    private final boolean throwIfInvalid;
 
     public FlagStore(final Connector connector) {
         this.connector = connector;
+        this.throwIfInvalid = false;
```
toddbaert: Option to throw if stuff is invalid - hard coded to false.

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/reconnect/rpc/FlagdRpcSetup.java
```diff
@@ -1,31 +1,38 @@
...
+        StepDefinitions.setUnstableProvider(workingProvider);
+
+        FeatureProvider unavailableProvider = new FlagdProvider(FlagdOptions.builder()
+                .resolverType(Config.Evaluator.RPC)
+                .port(8015) // this port isn't serving anything, error expected
+                // set a generous deadline, to prevent timeouts in actions
+                .deadline(100)
+                .cacheType(CacheType.DISABLED.getValue())
+                .build());
+        StepDefinitions.setUnavailableProvider(unavailableProvider);
```
toddbaert, line 36: This provider never connects, just validates the init errors.

providers/flagd/src/test/resources/flagConfigurations/valid-long.json
```diff
@@ -82,16 +83,7 @@
...
-          {
-            "in": [
-              "Chrome",
-              {
-                "var": "userAgent"
-              }
-            ]
-          },
-          "third",
-          null
```
toddbaert, line 94: This was actually invalid.

providers/flagd/src/test/resources/flagConfigurations/valid-long.json
```diff
@@ -1,4 +1,5 @@
 {
+  "$schema": "../../../main/resources/flags.json",
```
toddbaert, line 2: Adding links to local json validation file so we can see these files validation status live.

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/steps/StepDefinitions.java
```diff
@@ -521,4 +519,16 @@ public void the_resolved_string_zero_value_should_be(String expected) {
...
+    public void a_context_containing_a_targeting_key_with_value(String targetingKey) {
+        this.context = new ImmutableContext(targetingKey);
+    }
+
+    @Then("the returned reason should be {string}")
+    public void the_returned_reason_should_be(String reason) {
+        FlagEvaluationDetails<String> details = client.getStringDetails(this.stringFlagKey, this.stringFlagDefaultValue,
+                this.context);
+        assertEquals(reason, details.getReason());
+    }
```
toddbaert, line 533: new e2e assertions

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/reconnect/steps/StepDefinitions.java
```diff
@@ -103,4 +108,21 @@ public void when_the_connection_is_reestablished_the_provider_ready_handler_must
...
+    public void a_flagd_provider_is_set_and_initialization_is_awaited() {
+        // handled below
+    }
+
+    @Then("an error should be indicated within the configured deadline")
+    public void an_error_should_be_indicated_within_the_configured_deadline() {
+        assertThrows(Exception.class, () -> {
+            OpenFeatureAPI.getInstance().setProviderAndWait("unavailable", StepDefinitions.unavailableProvider);
+        });
+    }
```
toddbaert, line 127: new e2e assertions

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/model/FlagParser.java
```diff
@@ -40,37 +39,39 @@ private FlagParser() {
...
+            // we don't want to resolve anything from the network
+            Map<String, String> mappings = new HashMap<>();
+            mappings.put("https://flagd.dev/schema/v0/targeting.json", "classpath:targeting.json");
+            mappings.put("https://flagd.dev/schema/v0/flags.json", "classpath:flags.json");
+
+            SCHEMA_VALIDATOR = JsonSchemaFactory
+                    .getInstance(SpecVersion.VersionFlag.V7,
+                            builder -> builder
+                                    .schemaMappers(schemaMappers -> schemaMappers.mappings(mappings)))
+                    .getSchema(new URI("https://flagd.dev/schema/v0/flags.json"));
```
toddbaert, line 53: I think this is simpler, and it supports loading multiple schemas so they can be referenced from a single root.

Kavindu-Dodan: Nice. At first glance I thought this is a network call, but then saw the classpath loading :) This is really good improvement as we don't maintain custom stream loading logic

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/model/FlagParser.java
```diff
@@ -40,37 +39,39 @@ private FlagParser() {
...
         } catch (Throwable e) {
             // log only, do not throw
-            log.warn(String.format("Error loading resource %s, schema validation will be skipped", SCHEMA_RESOURCE), e);
+            log.warn(String.format("Error loading schema schema resources, schema validation will be skipped"));
```
Kavindu-Dodan:
```suggestion
            log.warn(String.format("Error loading schema resources, schema validation will be skipped"));
```

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/model/FlagParser.java
toddbaert: Ya the https... Is just the schema ID and frequently isn't even network resolvable, though ours is.

> This is really good improvement as we don't maintain custom stream loading logic

I thought so too.

I also thought a classpath uri might solve the issue of loading classpath flag definitions...

cc @thisthat

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/model/FlagParser.java
Kavindu-Dodan: I think this is fine. Parsing will anyway fail if we cannot derive `FeatureFlag` instances. I think it's the targeting rules that can get impacted if something's wrong.

providers/flagd/src/test/resources/flagConfigurations/valid-long.json
Kavindu-Dodan: This is a nice addition 👍

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/common/Util.java
```diff
@@ -21,7 +23,7 @@ public static void busyWaitAndCheck(final Long deadline, final AtomicBoolean che
 
         do {
             if (deadline <= System.currentTimeMillis() - start) {
-                throw new RuntimeException(
+                throw new GeneralError(
```
Kavindu-Dodan, line 26: 👍  a good move

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/model/FlagParser.java
Kavindu-Dodan: >  also thought a classpath uri might solve the issue of loading classpath flag definitions...

The notation is good (prefix `classpath:`)and I like it. Internally it's loading content using the classloader (see `ClasspathSchemaLoader`) :)

providers/flagd/src/test/resources/flagConfigurations/valid-long.json
Kavindu-Dodan: Ohh, we need to update flagd example herer -  https://github.com/open-feature/flagd/blob/main/config/samples/example_flags.flagd.json

I burrowed it 😞

Kavindu-Dodan (approved):
Nice work 🎉

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/FlagStore.java
thisthat: I would prefer
```suggestion
    public FlagStore(final Connector connector) {
        this(connector, false)
```

so we need to maintain a single constructor

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/model/FlagParser.java
thisthat: I think having this behavior configurable is the most sensible decision :)
I wonder now if we should always throw here and swallow the exception in FlagStore 🤔
I would suggest to throw here because otherwise, to know if the parsing was successful, we need to look into the size of the map 🤔

providers/flagd/src/main/resources/.gitignore
```diff
@@ -1 +1 @@
-flagd-definitions.json
\ No newline at end of file
+*.json
```
thisthat: nit
```suggestion
*.json

```

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/model/FlagParser.java
Kavindu-Dodan: We should not swallow the error and instead at least log. This is what we do here as the default behavior. 

Inside flag store, exception will make the provider to emit a stale state [1]. 

[1] - https://github.com/open-feature/java-sdk-contrib/blob/main/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/FlagStore.java#L110

providers/flagd/src/test/resources/flagConfigurations/valid-long.json
Kavindu-Dodan: PR to fix this in flagd - https://github.com/open-feature/flagd/pull/1222

toddbaert:
> I think having this behavior configurable is the most sensible decision :)
I wonder now if we should always throw here and swallow the exception in FlagStore 🤔
I would suggest to throw here because otherwise, to know if the parsing was successful, we need to look into the size of the map 🤔

The targeting doesn't impact the size of the map - it's not used at all until the flag is actually evaluated and it's validity doesn't prevent the flags from being parsed into the map. The behavior in this PR is consistent with flagd itself: it simply warns the targeting is invalid and logs it, and when the expression is evaluated it defaults.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/model/FlagParser.java
toddbaert: @thisthat 

The targeting doesn't impact the size of the map - it's not used at all until the flag is actually evaluated and it's validity doesn't prevent the flags from being parsed into the map. The behavior in this PR is consistent with flagd itself: it simply warns the targeting is invalid and logs it, and when the expression is evaluated it defaults (or if it can't be parsed at all it errors).

I'm happy to discuss this further in a separate issue if you still have concerns though.

Kavindu-Dodan: @toddbaert Yes, this is how I see this. Earlier, we would have failed on targeting rule evaluations and emit a stale state. But now, instead we store this flag and fail on the evaluation. Given this is consistent with flagd,  personally I think this is fine :)

PR #722, opened 2024-03-17
feat: add support for getBooleanEvaluation with default value

liran2000:
add support for getBooleanEvaluation with default value

@weihao-statsig @xinlili-statsig

### DIFF
--- a/providers/statsig/src/main/java/dev/openfeature/contrib/providers/statsig/StatsigProvider.java
+++ b/providers/statsig/src/main/java/dev/openfeature/contrib/providers/statsig/StatsigProvider.java
@@ -94,9 +94,10 @@ public class StatsigProvider extends EventProvider {
         String reason = null;
         if (featureConfigValue == null) {
             APIFeatureGate featureGate = Statsig.getFeatureGate(user, key);
-            if (featureGate.getReason() != null && assumeFailure(featureGate)) {
-                reason = featureGate.getReason().getReason();
-            } else {
+            reason = featureGate.getReason().getReason();
+
+            // in case of evaluation failure, remain with default value.
+            if (!assumeFailure(featureGate)) {
                 evaluatedValue = featureGate.getValue();
             }
         } else {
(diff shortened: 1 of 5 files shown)


weihao-statsig:
Thanks for changing this. Let me know if you find any other issues

providers/statsig/src/main/java/dev/openfeature/contrib/providers/statsig/StatsigProvider.java
```diff
@@ -82,11 +85,21 @@ public Metadata getMetadata() {
...
         StatsigUser user = ContextTransformer.transform(ctx);
         Boolean evaluatedValue = defaultValue;
-        try {
+        Value featureConfigValue = ctx.getValue(FEATURE_CONFIG_KEY);
+        String reason = null;
+        if (featureConfigValue == null) {
+            APIFeatureGate featureGate = Statsig.getFeatureGate(user, key);
+            if (featureGate.getReason() != null && assumeFailure(featureGate)) {
+                reason = featureGate.getReason().getReason();
+            } else {
```
toddbaert: I don't quite understand this. Could you perhaps add a comment?

toddbaert: I think the thing that's confusing me is it seems we are only setting the reason if the reason is a failure?

liran2000: I updated it a bit and added comment here and on the _assumeFailure()_ method.

toddbaert (approved):

providers/statsig/src/main/java/dev/openfeature/contrib/providers/statsig/StatsigProvider.java
toddbaert: Thanks!

PR #724, opened 2024-03-18
feat: Set Unleash changed flags on change event.

RafeArnold:
## This PR
- adds the flags that have been changed to the configuration change event for the Unleash provider.

### Notes
This addition makes checking which flags have updated trivial for the end-user.

### DIFF
--- a/providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashSubscriberWrapper.java
+++ b/providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashSubscriberWrapper.java
@@ -3,6 +3,7 @@ package dev.openfeature.contrib.providers.unleash;
 import dev.openfeature.sdk.EventProvider;
 import dev.openfeature.sdk.ImmutableMetadata;
 import dev.openfeature.sdk.ProviderEventDetails;
+import io.getunleash.FeatureToggle;
 import io.getunleash.UnleashException;
 import io.getunleash.event.ImpressionEvent;
 import io.getunleash.event.ToggleEvaluated;
@@ -18,6 +19,8 @@ import lombok.Generated;
 import lombok.extern.slf4j.Slf4j;
 
 import javax.annotation.Nullable;
+import java.util.ArrayList;
+import java.util.List;
 
 /**
  * UnleashSubscriber wrapper for emitting event provider events.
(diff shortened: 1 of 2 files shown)


liran-p:
hi @RafeArnold thanks for the changes, looks good.
can you please add some background, on how did you come up with this need, where is it used etc' ?

liran2000:
hi @RafeArnold thanks for the changes, looks good.
can you please add some background, on how did you come up with this need, where is it used etc' ?

RafeArnold:
> hi @RafeArnold thanks for the changes, looks good. can you please add some background, on how did you come up with this need, where is it used etc' ?

The background for this change is that being able to see which flags have actually been updated when a `ProviderConfigurationChanged` event is emitted is really useful at preventing unnecessary work for the receiver of the event. If a event handler is only interesting in a single flag, or a subset of flags, they can easily query the received event to see if the flags they care about have changed.

I've realised that my implementation isn't actually doing what I intended, since the ToggleResponse will contain all the feature toggles available, not just the updated ones. I guess the subscriber wrapper would require a cache of the flag keys that it could compare and update with the flags that were retrieved from the Unleash API. I'll have another crack at it and see what I can produce.

RafeArnold:
Having thought about it further, to implement my intention, I believe the subscriber wrapper would need to keep its own cache of all the flags, which feels like a huge waste of resources. If this is the case, having the event receivers perform the checks themselves seems like the best option. Feel free to close this PR, unless a better solution can be thought of.

providers/unleash/src/main/java/dev/openfeature/contrib/providers/unleash/UnleashSubscriberWrapper.java
```diff
@@ -69,7 +72,12 @@ public void toggleEvaluated(ToggleEvaluated toggleEvaluated) {
     public void togglesFetched(FeatureToggleResponse toggleResponse) {
         unleashSubscriber.togglesFetched(toggleResponse);
         if (FeatureToggleResponse.Status.CHANGED.equals(toggleResponse.getStatus())) {
+            List<String> flagsChanged = new ArrayList<>();
+            for (FeatureToggle featureToggle : toggleResponse.getToggleCollection().getFeatures()) {
```
liran2000, line 76: What does _toggleResponse.getToggleCollection().getFeatures()_ contain ?
Only changed features ?
Any documentation reference from Unleash ?

RafeArnold: Yeah, unfortunately, as per my other comment, it contains all the feature toggles available from the Unleash API, not just the changed ones, so this implementation doesn't work as intended.

PR #811, opened 2024-06-07
build: add maven wrapper & update docs

alexandraoberaigner:
## This PR
We are adding the maven wrapper to simplify project setup for contributing.

### How to test
run `./mvnw -version`


aepfli:
Does it make sense to also adopt the build files to use the maven wrapper?

Kavindu-Dodan:
> Does it make sense to also adopt the build files to use the maven wrapper?

hmm, this will require an upgrade to github action and add the same to other providers in this repository. I think having wrapper for flagd for now is sufficient for ease of use.

Kavindu-Dodan (approved):
I think this is a good helper for flagd :)

PR #817, opened 2024-06-11
feat: avoid star imports via checkstyle

### LINKED ISSUE #816: Activate Checkstyle rule to avoid star imports
The usage of Star imports within java is generally seen as a bad practice as it leads to tight coupling between packages and might lead to problems when a new version of a library introduces name clashes. see https://checkstyle.sourceforge.io/version/4.4/5.x/config_imports.html - as we do have a tool in place which can prevent the usage, we should activate the rule 

```
<module name="AvoidStarImport">
</module>
```

see the already raised nit in a pull request https://github.com/open-feature/java-sdk-contrib/pull/810#discussion_r1631573948

wdyt?

aepfli:
Star imports are seen as a bad practice within Java, as they can lead to too tight coupling with packages and the introduction of name clashes. Hence, we should avoid them.

closes: #816

### DIFF
--- a/checkstyle.xml
+++ b/checkstyle.xml
@@ -51,7 +51,7 @@
     <module name="TreeWalker">
         <!-- needed for SuppressWarningsFilter -->
         <module name="SuppressWarningsHolder" />
-        
+
         <module name="SuppressWarnings">
             <property name="id" value="checkstyle:suppresswarnings"/>
         </module>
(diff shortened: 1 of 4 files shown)


checkstyle.xml
```diff
@@ -62,7 +62,11 @@
                         default="checkstyle-xpath-suppressions.xml" />
             <property name="optional" value="true"/>
         </module>
-    
+
+
+        <module name="AvoidStarImport">
+        </module>
```
madhead:
```suggestion
        <module name="AvoidStarImport" />
```

?

madhead (approved):

checkstyle.xml
aepfli: ~~thank you for your feedback, i am just unsure, if this means, the addition to the checkstyle rules is a good idea not.~~

sorry i misread the ui

Kavindu-Dodan (approved):

thiyagu06 (approved):

thomaspoignant (approved):

Issue #410, opened 2023-08-28
[flagd] shutting down provider should reset the state.


thiyagu06:
https://github.com/open-feature/java-sdk-contrib/blob/main/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java#L56
It's not always guaranteed that `FeatureProvider.shutdown()` method is called at the end of application lifecycle. The user can call the shutdown method at any point in time of application lifetime. So it's better to reset the state. may be NOT_READY or new state and publish the event with the new state.

Kavindu-Dodan:
@thiyagu06 I also see a potential improvement here. Though the spec [^1] doesn't enforce a state change requirement after a shutdown, we can set the status to `NOT_READY` after executing the shutdown logic [^2].

Feel free to open up a contribution PR on this. 

[^1]: https://github.com/open-feature/spec/blob/main/specification/sections/02-providers.md#25-shutdown
[^2]: https://github.com/open-feature/java-sdk-contrib/blob/main/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java#L54-L60

thiyagu06:
Sure. I will work on this.

toddbaert:
Hey @thiyagu06 I'm going to unassign this from you for now, since we have an event coming up and KubeCon and it might be a good issue for people to take on. However, if you intend on completing it soon, let me know I will reassign to you.

mehtasankets:
@toddbaert , I would like to contribute to this issue. Is it possible to assign it to me?

toddbaert:
@mehtasankets Assigned.

Kavindu-Dodan:
Fixed with #832 . Thank you @mehtasankets for the contribution 👍

PR #832, opened 2024-06-18
feat: Reset the state on shutting down the flagd resolver (#410)

mehtasankets:
### Notes
I've reset the state in corresponding resolvers instead of adding it in the provider. That way, resolvers can decide when and which state to reset to. Let me know if we should move this logic to provider instead.


providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/FlagdProviderTest.java
```diff
@@ -863,4 +912,29 @@ private FlagdProvider createProvider(GrpcConnector grpc, Cache cache, Supplier<P
...
 
+    // Create an in process provider
+    private FlagdProvider createInProcessProvider() {
+
+        final FlagdOptions flagdOptions = FlagdOptions.builder()
+                .resolverType(Config.Resolver.IN_PROCESS)
+                .deadline(1000)
+                .build();
+        final FlagdProvider provider = new FlagdProvider(flagdOptions);
+        final MockStorage mockStorage = new MockStorage(new HashMap<String, FeatureFlag>(), new LinkedBlockingQueue<StorageState>(List.of(StorageState.OK)));
```
Kavindu-Dodan: `List.of` won't work as we still maintain support for Java 8

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
```diff
@@ -14,6 +14,7 @@
 import dev.openfeature.sdk.Value;
 import lombok.extern.slf4j.Slf4j;
 
+import java.util.List;
```
Kavindu-Dodan: could you please check on unused imports ? Also in other changed classe (ex:- FlagdProviderTest.java )

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
```diff
@@ -183,5 +190,12 @@ private void handleStateTransition(ProviderState oldState, ProviderState newStat
             this.emitProviderReady(details);
             this.emitProviderConfigurationChanged(details);
         }
+        // we shutdown from an error
+        if (ProviderState.ERROR.equals(oldState) && ProviderState.NOT_READY.equals(newState)) {
```
Kavindu-Dodan: This condition will not be reached as there's `if (ProviderState.NOT_READY.equals(newState))` in line 167

Kavindu-Dodan: Besides, this check is not needed. The state can transit from Error to Shutdown [1]

[1] - https://openfeature.dev/specification/sections/flag-evaluation#17-provider-lifecycle-management

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/FlagdProviderTest.java
mehtasankets: Fixed.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
mehtasankets: Fixed.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
mehtasankets: Fixed.

Kavindu-Dodan (approved):
Nice work @mehtasankets

mvnw
Kavindu-Dodan, line 1: Not sure why there's a change here. @mehtasankets could you please check ?

mehtasankets: I just made it executable. i.e. ran `chmod +x mvnw`.

Kavindu-Dodan: Ohh, noted 👍

toddbaert (approved):
Thanks!

PR #833, opened 2024-06-19
feat: Change fractional custom op from percentage-based to relative weighting. #828

### LINKED ISSUE #828: [FEATURE] Change fractional custom op from percentage-based to relative weighting.
### Requirements

Currently, the fractional operator in the in-process provider is percentage-based; meaning you must specify the relative distribution of the variants in terms of a percentage. Consistent with flagd, we should change this instead to be a relative weight, meaning that the integers do not need to sum to 100. For example, this fractional:

```json
"fractional": [
          [
            "red",
            25
          ],
          [
            "blue",
            25
          ],
          [
            "green",
            25
          ],
          [
            "grey",
            25
          ]
        ]
```

Would remain valid, but could also be expressed as:

```json
"fractional": [
          [
            "red",
            1
          ],
          [
            "blue",
            1
          ],
          [
            "green",
            1
          ],
          [
            "grey",
            1
          ]
        ]
```
See [issue](https://github.com/open-feature/flagd/issues/1282) and [PR](https://github.com/open-feature/flagd/pull/1313) in Go/flagd.

aepfli:
### Related Issues

Fixes #828


aepfli:
i am not sure that my change caused the ci to fail. maybe i misread the log, but could some body of the maintainers restart the build

aepfli:
okay i am realizing that the e2e tests are failing, i need to investigate, what might be the issue - do i maybe need to update some dependencies - or do you see an error in my implementation?

toddbaert:
> okay i am realizing that the e2e tests are failing, i need to investigate, what might be the issue - do i maybe need to update some dependencies - or do you see an error in my implementation?

I'll try to debug locally and see. It's possible you are just very unlucky though and this is unrelated, because the assert that's failing doesn't _seem_ to do with fractional at all... but these tests are usually very reliable and not flacky so :shrug: 

I will re-run to double-check

toddbaert:
It's the assertion in `the_provider_configuration_changed_handler_must_run`. It is consistently failing. I'm surprised though, it doesn't really have to do with fractional.

toddbaert:
If you rebase on or merge main the e2e suite should work now.

See: https://github.com/open-feature/java-sdk-contrib/pull/834

aepfli:
relates also to:  https://github.com/open-feature/flagd/issues/1282 - for tracking purpose in the main ticket

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/resolver/process/targeting/FractionalTest.java
```diff
@@ -189,7 +189,49 @@ void invalidRuleSumNot100() throws JsonLogicEvaluationException {
...
+
+        final List<Object> rule = new ArrayList<>();
+
+        final List<Object> bucket1 = new ArrayList<>();
+        bucket1.add("blue");
+        bucket1.add(50);
+
+        final List<Object> bucket2 = new ArrayList<>();
+        bucket2.add("green");
+        bucket2.add(70);
```
Kavindu-Dodan: nit - to match with the rule

```suggestion
        bucket2.add(30);
```

Kavindu-Dodan (approved):
Other than the nit, looks good 🙌

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/resolver/process/targeting/FractionalTest.java
```diff
@@ -189,7 +189,49 @@ void invalidRuleSumNot100() throws JsonLogicEvaluationException {
...
+        final List<Object> bucket2 = new ArrayList<>();
+        bucket2.add("green");
+        bucket2.add(30);
+
+        rule.add(bucket1);
+        rule.add(bucket2);
+
+        Map<String, String> data = new HashMap<>();
+        data.put(FLAG_KEY, "headerColor");
+        data.put(TARGET_KEY, "foo@foo.com");
```
thisthat: Maybe we should have a comment saying that the `target key` maps to the first variant

thisthat (commented):
LGTM but I'd suggest to add the example reported in https://github.com/open-feature/java-sdk-contrib/issues/828 for testing purpose to validate that if we have a 50/50 (or 25/25/25/25) it's equivalent to 1/1 (or 1/1/1/1)

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/resolver/process/targeting/FractionalTest.java
aepfli: I'm not sure what you mean exactly by this. Most likely, this is due to my lack of knowledge; I only copied and pasted the code, adapted input and expectations, and left the rest as is. Can you please elaborate?

aepfli:
> LGTM but I'd suggest to add the example reported in #828 for testing purpose to validate that if we have a 50/50 (or 25/25/25/25) it's equivalent to 1/1 (or 1/1/1/1)

@thisthat i modified the tests, and added some more, each test is now an own json file, with a rule and an expected result, this allows us to easier add test cases in the future - i hope this is fine.

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/resolver/process/targeting/FractionalTest.java
thisthat: sure :) The [targeting key](https://openfeature.dev/docs/reference/concepts/evaluation-context/#targeting-key) is hashed and its value is used to put the user in always the same bucket when we do fractional evaluations. 
Therefore, the value `foo@foo.com` makes the fractional evaluation always return the first variant. Having this as a comment will make the testing easier to follow and understand why we always expect the first variant to be returned :)

thisthat (approved):
Thanks a lot @aepfli :shipit:

toddbaert (approved):

PR #834, opened 2024-06-19
fix: broken netty transport

toddbaert:
The latest netty epoll version seems to be broken and is causing inconsistent test failures impacting `main` and other PRs. This reverts it.

Also makes some improvements to our docker-compose for local e2e.

See: https://github.com/open-feature/java-sdk-contrib/pull/819

### DIFF
--- a/providers/flagd/docker-compose.yaml
+++ b/providers/flagd/docker-compose.yaml
@@ -1,25 +1,17 @@
 services:
   flagd:
-    build:
-      context: test-harness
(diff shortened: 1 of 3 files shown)


providers/flagd/docker-compose.yaml
```diff
@@ -1,25 +1,17 @@
 services:
   flagd:
-    build:
```
toddbaert, line 3: Use the published versions

Kavindu-Dodan (approved):

PR #850, opened 2024-06-28
feat: use namespaced schemas for flagd json schemas (#843)

### LINKED ISSUE #843: [Improvement] flagd json schema files should be namespaced under resources
**Background**

flagd provider utilizes schema files to validate flags when working in in-process mode. Currently, these json schema files are copied when building the binary to `resources` and kept in the root level [1]. And they are named `flags.json` & `targeting.json`.

**Problem**

In an application that uses flagd Java provider for feature flagging, there can be conflicting JSON files with similar names to the above schema files in the classpath. This especially concerns the name `flags.json` as many flagd examples use `flags.json` as a flag source file. 

**Improvement**

Given our underlying library (seems to) support namespaced locations [2], we should improve our schema locations with an appropriate namespace to avoid classpath conflicts. 

**AC**

Provider works & validates flags correctly with the namespaced schemas in the classpath.

[1] - https://github.com/open-feature/java-sdk-contrib/blob/dev.openfeature.contrib.providers.flagd-v0.8.3/providers/flagd/pom.xml#L236-L258 
[2] - https://github.com/networknt/json-schema-validator/blob/master/doc/schema-retrieval.md#configuring-mappings

mehtasankets:
### Related Issues
Fixes #843 

### Notes
* Update build file to copy over schema files to `src/main/resources/flagd/schemas/` instead of `src/main/resources/`
* Refer targeting.json and flags.json from `src/main/resources/flagd/schemas/`
* No new tests are added as existing tests should be sufficient to validate the feature
* Verified the new location of the schema files by building and extracting the jar locally


Kavindu-Dodan (approved):
Change looks good. Validated locally and it works 🙌

toddbaert (approved):

toddbaert:
Thanks @mehtasankets !

PR #915, opened 2024-08-13
feat: add Prefab provider

liran2000:
[Readme](https://github.com/open-feature/java-sdk-contrib/blob/bd22d51beaf46deefc7354fced990932e95ee4f5/providers/prefab/README.md#unofficial-prefab-openfeature-provider-for-java) describes the provider.


providers/prefab/src/test/java/dev/openfeature/contrib/providers/prefab/PrefabProviderTest.java
```diff
@@ -0,0 +1,174 @@
...
+        assertEquals(1.1, client.getDoubleValue("non-existing", 1.1));
+
+        // non-number flag value
+        assertEquals(1.1, client.getDoubleValue(VARIANT_FLAG_NAME, 1.1));
+    }
+
+//    @Test
+    void getBooleanEvaluationByUser() {
+        MutableContext evaluationContext = new MutableContext();
+        evaluationContext.add("domain", "prefab.cloud");
```
liran2000: hi @jkebinger,
can you help on why this evaluation not returning true by the context domain ?
it translates to:
```java
prefabCloudClient.featureFlagClient().featureIsOn(key, context);
with context - notice "\n" which is audo added to the string:
PrefabContext{name=User, properties={domain=string: "prefab.cloud"\n}}
```
[flag config](https://github.com/open-feature/java-sdk-contrib/pull/915/files#diff-f37820eae2aaf5dbf312964f585489ffe90fcb83c2f89441b3e8de6013e2dce6R13):
> just_my_domain: { "feature_flag": "true", value: "new-version", criteria: { operator: PROP_IS_ONE_OF, property: "domain", values: [ "prefab.cloud", "example.com" ] } }

providers/prefab/pom.xml
```diff
@@ -0,0 +1,40 @@
...
+
+    <name>prefab</name>
+    <description>Prefab provider for Java</description>
+    <url>https://www.prefab.cloud</url>
+
+    <dependencies>
+        <dependency>
+            <groupId>cloud.prefab</groupId>
+            <artifactId>client</artifactId>
+            <version>0.3.20</version>
```
liran2000, line 23: hi @jkebinger,
OpenFeature SDK built with Java 8, so it will require Java 8 version.
Is it possible to release Java 8 Prefab client ?

providers/prefab/src/test/java/dev/openfeature/contrib/providers/prefab/PrefabProviderTest.java
jkebinger: I’m traveling this week so I can’t help with any of this yet. A Java 8 client is probably a pretty substantial lift but it’ll require some study

liran2000: ok I managed to solved it via the docs, replaced the deprecated _yaml_ config with _json_ config.
You can review this PR on your convenient time.
For Java 8, you may update Prefab Java client to version 8 if there are not too many specific Java>8 code sections.
thanks

providers/prefab/pom.xml
jkebinger: It appears the primary change to get to java8 compatibility will be removing the use of the java 11 HttpClient, so not too bad. Can't make any promises on when we may be able to prioritize that work; I'll probably aim to implement it such that there's a java8 compatibility add on so java 11+ users don't have to get an additional set of dependencies (probably okhttp). We'll see

PR #925, opened 2024-08-24
feat: emit changed flags in configuration change event

### LINKED ISSUE #901: [flagd-in-process] Support changed flag keys in the provider configuration change event
## Overview

The flagd in-process provider will emit a change event when the flag configuration is updated. However, it does not include a list of flag keys that have changed. This makes it impossible to reevaluate a feature flag only when the configuration changes.

## Requirements

- Include a list of change flags in the provider configuration changed event: https://github.com/open-feature/java-sdk-contrib/blob/main/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java#L174
- A flag is considered changed if it has been added, removed, or updated.
- Associated tests

## Note

This functionality is supported in JavaScript. The PR can be found [here](https://github.com/open-feature/js-sdk-contrib/pull/703).

UtkarshSharma2612:
Adds changed flags to in-process change event.

Fixes: https://github.com/open-feature/java-sdk-contrib/issues/901

### DIFF
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
@@ -4,10 +4,8 @@ import dev.openfeature.contrib.providers.flagd.resolver.Resolver;
 import dev.openfeature.contrib.providers.flagd.resolver.grpc.GrpcResolver;
 import dev.openfeature.contrib.providers.flagd.resolver.grpc.cache.Cache;
 import dev.openfeature.contrib.providers.flagd.resolver.process.InProcessResolver;
-import dev.openfeature.contrib.providers.flagd.resolver.process.model.FeatureFlag;
 import dev.openfeature.sdk.EvaluationContext;
 import dev.openfeature.sdk.EventProvider;
-import dev.openfeature.sdk.FeatureProvider;
 import dev.openfeature.sdk.Metadata;
 import dev.openfeature.sdk.ProviderEvaluation;
 import dev.openfeature.sdk.ProviderEventDetails;
@@ -25,7 +23,7 @@ import java.util.concurrent.locks.ReentrantReadWriteLock;
  */
 @Slf4j
 @SuppressWarnings({"PMD.TooManyStaticImports", "checkstyle:NoFinalizer"})
-public class FlagdProvider extends EventProvider implements FeatureProvider {
+public class FlagdProvider extends EventProvider {
     private static final String FLAGD_PROVIDER = "flagD Provider";
 
     private final ReadWriteLock lock = new ReentrantReadWriteLock();
@@ -173,8 +171,8 @@ public class FlagdProvider extends EventProvider implements FeatureProvider {
         // configuration changed
         if (ProviderState.READY.equals(oldState) && ProviderState.READY.equals(newState)) {
             log.debug("Configuration changed");
-            ProviderEventDetails details = ProviderEventDetails.builder().message("configuration changed").build();
-            details.setFlagsChanged(changedFlagKeys);
+            ProviderEventDetails details = ProviderEventDetails.builder().flagsChanged(changedFlagKeys)
+                    .message("configuration changed").build();
             this.emitProviderConfigurationChanged(details);
             return;
         }
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
@@ -4,6 +4,7 @@ import dev.openfeature.contrib.providers.flagd.resolver.Resolver;
 import dev.openfeature.contrib.providers.flagd.resolver.grpc.GrpcResolver;
 import dev.openfeature.contrib.providers.flagd.resolver.grpc.cache.Cache;
 import dev.openfeature.contrib.providers.flagd.resolver.process.InProcessResolver;
+import dev.openfeature.contrib.providers.flagd.resolver.process.model.FeatureFlag;
 import dev.openfeature.sdk.EvaluationContext;
 import dev.openfeature.sdk.EventProvider;
 import dev.openfeature.sdk.FeatureProvider;
@@ -14,6 +15,7 @@ import dev.openfeature.sdk.ProviderState;
 import dev.openfeature.sdk.Value;
 import lombok.extern.slf4j.Slf4j;
 
+import java.util.List;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReadWriteLock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
@@ -142,7 +144,7 @@ public class FlagdProvider extends EventProvider implements FeatureProvider {
         return clientCallCtx;
     }
 
-    private void setState(ProviderState newState) {
+    private void setState(ProviderState newState, List<String> changedFlagsKeys) {
         ProviderState oldState;
         Lock l = this.lock.writeLock();
         try {
@@ -152,17 +154,17 @@ public class FlagdProvider extends EventProvider implements FeatureProvider {
         } finally {
             l.unlock();
         }
-        this.handleStateTransition(oldState, newState);
+        this.handleStateTransition(oldState, newState, changedFlagsKeys);
     }
 
-    private void handleStateTransition(ProviderState oldState, ProviderState newState) {
+    private void handleStateTransition(ProviderState oldState, ProviderState newState, List<String> changedFlagKeys) {
         // we got initialized
         if (ProviderState.NOT_READY.equals(oldState) && ProviderState.READY.equals(newState)) {
             // nothing to do, the SDK emits the events
             log.debug("Init completed");
             return;
         }
-        // we got shutdown, not checking oldState as behavior remains the same for shutdown 
+        // we got shutdown, not checking oldState as behavior remains the same for shutdown
         if (ProviderState.NOT_READY.equals(newState)) {
             // nothing to do
             log.debug("shutdown completed");
@@ -172,6 +174,7 @@ public class FlagdProvider extends EventProvider implements FeatureProvider {
         if (ProviderState.READY.equals(oldState) && ProviderState.READY.equals(newState)) {
             log.debug("Configuration changed");
             ProviderEventDetails details = ProviderEventDetails.builder().message("configuration changed").build();
+            details.setFlagsChanged(changedFlagKeys);
             this.emitProviderConfigurationChanged(details);
             return;
         }
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/grpc/GrpcConnector.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/grpc/GrpcConnector.java
@@ -53,7 +53,8 @@ public class GrpcConnector {
      * @param cache         cache to use.
      * @param stateConsumer lambda to call for setting the state.
      */
-    public GrpcConnector(final FlagdOptions options, final Cache cache, BiConsumer<ProviderState, List<String>> stateConsumer) {
+    public GrpcConnector(final FlagdOptions options, final Cache cache,
+            BiConsumer<ProviderState, List<String>> stateConsumer) {
         this.channel = ChannelBuilder.nettyChannel(options);
         this.serviceStub = ServiceGrpc.newStub(channel);
         this.serviceBlockingStub = ServiceGrpc.newBlockingStub(channel);
@@ -81,7 +82,8 @@ public class GrpcConnector {
     /**
      * Shuts down all gRPC resources.
      *
-     * @throws Exception is something goes wrong while terminating the communication.
+     * @throws Exception is something goes wrong while terminating the
+     *                   communication.
      */
     public void shutdown() throws Exception {
         // first shutdown the event listener
@@ -115,12 +117,13 @@ public class GrpcConnector {
     }
 
     /**
-     * Event stream observer logic. This contains blocking mechanisms, hence must be run in a dedicated thread.
+     * Event stream observer logic. This contains blocking mechanisms, hence must be
+     * run in a dedicated thread.
      */
     private void observeEventStream() {
         while (this.eventStreamAttempt <= this.maxEventStreamRetries) {
-            final StreamObserver<EventStreamResponse> responseObserver =
-                    new EventStreamObserver(sync, this.cache, this::grpcStateConsumer);
+            final StreamObserver<EventStreamResponse> responseObserver = new EventStreamObserver(sync, this.cache,
+                    this::grpcStateConsumer);
             this.serviceStub.eventStream(EventStreamRequest.getDefaultInstance(), responseObserver);
 
             try {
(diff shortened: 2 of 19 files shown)


providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
```diff
@@ -33,6 +37,7 @@ public class FlagdProvider extends EventProvider implements FeatureProvider {
 
     private EvaluationContext evaluationContext;
 
+    private Map<String, FeatureFlag> changedFlags = new HashMap<>();
```
toddbaert: Keeping this state is tricky, and will almost certainly lead to bugs. It's also not threadsafe.

I think instead of keeping this state in a few places, you should add an additional nullable parameter to the `stateSupplier` function in this class, which specifies the changed flags (when applicable) then you can just use that in the event emission. That way there's no state to manage or synchronize or worry about.

toddbaert (changes requested):
https://github.com/open-feature/java-sdk-contrib/pull/925#discussion_r1731752438

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
UtkarshSharma2612: understood, will do this

UtkarshSharma2612: Have fixed it, can you please help with review 😅

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolver.java
```diff
@@ -100,38 +106,38 @@ public void init() throws Exception {
...
-        stateConsumer.accept(ProviderState.NOT_READY);
+        stateConsumer.accept(ProviderState.NOT_READY,null);
     }
 
     /**
      * Resolve a boolean flag.
      */
     public ProviderEvaluation<Boolean> booleanEvaluation(String key, Boolean defaultValue,
-            EvaluationContext ctx) {
+                                                         EvaluationContext ctx) {
```
aepfli: [nit] this formatting changes look a little bit odd to me

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/FlagStore.java
```diff
@@ -103,6 +107,7 @@ private void streamerListener(final Connector connector) throws InterruptedExcep
                         Map<String, FeatureFlag> flagMap = FlagParser.parseString(take.getData(), throwIfInvalid);
                         writeLock.lock();
                         try {
+                            updateChangedFlags(flagMap);
```
aepfli: I think this produces side effects. Rather than manipulating the `changed flags` within the method, we should return a list and reassign it here. Ultimately, the process creates a fresh list anyway, with clearing up front.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/grpc/GrpcConnector.java
```diff
@@ -37,7 +37,7 @@ public class GrpcConnector {
     private final long deadline;
 
     private final Cache cache;
-    private final Consumer<ProviderState> stateConsumer;
+    private final BiConsumer<ProviderState, List<String>> stateConsumer;
```
aepfli, line 41: For now, this is fine, but we should consider whether it makes sense to just use a Consumer and a POJO to provide more information. Changing the amount of parameters includes a lot of changes. A proper POJO with a good builder could also reduce this heavily in the future.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolver.java
UtkarshSharma2612: sure will update

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/FlagStore.java
UtkarshSharma2612: will implement this

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/FlagStore.java
```diff
@@ -31,6 +32,9 @@ public class FlagStore implements Storage {
     private final BlockingQueue<StorageState> stateBlockingQueue = new LinkedBlockingQueue<>(1);
     private final Map<String, FeatureFlag> flags = new HashMap<>();
 
+    @Getter
```
toddbaert: This is another place where we have some state which presents potential bugs and concurrency issues.

Similar to what you did with the `stateConsumer` function, I think you need to modify the `StorageState` object not to be just an enum, but an object which contains and enum AND a list of changed flags. Then, when, we can offer both to the queue so we can avoid saving this state.

I hope that makes sense! Other than that, everything looks great here. Thanks for all your efforts!

toddbaert:
@UtkarshSharma2612 I squashed your branch to a single commit and left a comment [here](https://github.com/open-feature/java-sdk-contrib/pull/925#discussion_r1745967904).

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/FlagStore.java
toddbaert: I think this is a better solution: https://github.com/open-feature/java-sdk-contrib/pull/925#discussion_r1745967904

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/FlagStore.java
UtkarshSharma2612: understood will implement this 👍

UtkarshSharma2612: @toddbaert have implemented this please help with review

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
```diff
@@ -172,6 +174,9 @@ private void handleStateTransition(ProviderState oldState, ProviderState newStat
         if (ProviderState.READY.equals(oldState) && ProviderState.READY.equals(newState)) {
             log.debug("Configuration changed");
             ProviderEventDetails details = ProviderEventDetails.builder().message("configuration changed").build();
+            if (!changedFlagKeys.isEmpty()) {
```
aepfli: I'm just curious about the benefit of this. Can we add an empty list of changed flags? In my brain, this should work the same way if I pass changed flags or an empty list of changed flags. I have to admit that I don't have the complete picture, nor am I aware of all the implications  - but ideally, we can remove this condition and reduce code complexity

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/StorageStateDTO.java
```diff
@@ -0,0 +1,22 @@
...
+import java.util.Map;
+
+@Getter
+@ToString
+@EqualsAndHashCode
+public class StorageStateDTO {
+    private final StorageState storageState;
+    private final Map<String, FeatureFlag> changedFlags;
+
+    public StorageStateDTO(StorageState storageState, Map<String, FeatureFlag>  changedFlags) {
```
aepfli: a second constructor with just StorageState would be really beneficial, we should try to avoid passing null to a constructor

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/StorageStateDTO.java
```diff
@@ -0,0 +1,22 @@
...
+import lombok.ToString;
+
+import java.util.Map;
+
+@Getter
+@ToString
+@EqualsAndHashCode
+public class StorageStateDTO {
+    private final StorageState storageState;
+    private final Map<String, FeatureFlag> changedFlags;
```
aepfli: I am not sure how we handle this in other SDKs, but do we need the whole "FeatureFlag" object here, or is a list of Keys enough - it seems to me we do a lot of mapping - around this map, for the fact that only the key is relevant to us. We can reduce the complexity of the rest of the code if we only store the key of the flags in a list here (or if I am missing something). The only thing that might be interesting for the changed flags is the actual manipulation, if added, removed, or updated. wdyt?

aepfli (commented):
@UtkarshSharma2612 I think this looks good already; my thoughts for the review are more regarding complexity and maybe unneeded transformations for data, which can increase our footprint. Please have a look and let me know what you think

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/StorageStateDTO.java
UtkarshSharma2612: This is correct will implement this.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/StorageStateDTO.java
UtkarshSharma2612: That's right, we have a method in the storage interface that retrieves the feature flag using its key, allowing us to access it from anywhere whenever needed.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
UtkarshSharma2612: I agree by defining an EMPTY_LIST in the constructor we are eliminating null pointers, and empty list shouldn't cause problems, will implement this

UtkarshSharma2612: Although in GRPC connector, when we were passing values to state consumer, I was passing it as null. Have also made the change to pass an EmptyList instead to avoid null pointer exception and bring null anywhere

UtkarshSharma2612:
@aepfli have implemented changes please help with re-review

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/FlagStore.java
```diff
@@ -132,4 +135,27 @@ private void streamerListener(final Connector connector) throws InterruptedExcep
         log.info("Shutting down store stream listener");
     }
 
+    private List<String> getChangedFlagsKeys(Map<String, FeatureFlag> newFlags) {
+        Map<String, FeatureFlag> changedFlags = new HashMap<>();
```
aepfli, line 139: We can already work with lists here; maybe even one Set is enough to which we add those elements directly.

aepfli (approved):
lgtm, i am just thinking that actually a set should be enough for the flag keys, maybe instead of class `List` for the `changedFlags` we can even use the interface `Collection` so we can easier switch between different collection implementations. wdyt about using a Hashset, rather than a list for the changedflags? do we need that safety of no duplicates within the changedflags?

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/steps/StepDefinitions.java
```diff
@@ -440,7 +446,7 @@ public void the_provider_configuration_changed_handler_must_run() {
 
     @Then("the event details must indicate {string} was altered")
     public void the_event_details_must_indicate_was_altered(String flagKey) {
-        // TODO: In-process-provider doesnt support flag change list.
+        assertEquals(flagKey, this.changedFlag);
```
toddbaert, line 449: We can now assert this in e2e tests!

cc @aepfli @UtkarshSharma2612

toddbaert (approved):
@UtkarshSharma2612 I squashed all your changes, and made a few small changes in [this commit](https://github.com/open-feature/java-sdk-contrib/pull/925/commits/ac5896c39169b318fb0376177eeef15123cd5425): 

- added the changed flags to RPC mode as well (this was just a couple of lines, with the work you already did it was very easy)
- added a new assertion in the e2e test suite which was previously not implemented because we didn't support `flagsChanged`
- some minor checkstyle and spotbugs fixes

Really great work! Thanks for doing all the heavy lifting here!

UtkarshSharma2612:
> @UtkarshSharma2612 I squashed all your changes, and made a few small changes in [this commit](https://github.com/open-feature/java-sdk-contrib/pull/925/commits/ac5896c39169b318fb0376177eeef15123cd5425):
> 
> * added the changed flags to RPC mode as well (this was just a couple of lines, with the work you already did it was very easy)
> * added a new assertion in the e2e test suite which was previously not implemented because we didn't support `flagsChanged`
> * some minor checkstyle and spotbugs fixes
> 
> Really great work! Thanks for doing all the heavy lifting here!

Thank you, @toddbaert! I really enjoyed working on this issue and learned a lot in the process. Are there any other open issues I could contribute to? I'd love to continue helping out. 😄

toddbaert:
> > @UtkarshSharma2612 I squashed all your changes, and made a few small changes in [this commit](https://github.com/open-feature/java-sdk-contrib/pull/925/commits/ac5896c39169b318fb0376177eeef15123cd5425):
> > 
> > * added the changed flags to RPC mode as well (this was just a couple of lines, with the work you already did it was very easy)
> > * added a new assertion in the e2e test suite which was previously not implemented because we didn't support `flagsChanged`
> > * some minor checkstyle and spotbugs fixes
> > 
> > Really great work! Thanks for doing all the heavy lifting here!
> 
> Thank you, @toddbaert! I really enjoyed working on this issue and learned a lot in the process. Are there any other open issues I could contribute to? I'd love to continue helping out. 😄

@UtkarshSharma2612 I'm sure there is. I'll let you know.

I will also invite you to the org, which is the first step to getting more involved.

UtkarshSharma2612:
> > > @UtkarshSharma2612 I squashed all your changes, and made a few small changes in [this commit](https://github.com/open-feature/java-sdk-contrib/pull/925/commits/ac5896c39169b318fb0376177eeef15123cd5425):
> > > 
> > > * added the changed flags to RPC mode as well (this was just a couple of lines, with the work you already did it was very easy)
> > > * added a new assertion in the e2e test suite which was previously not implemented because we didn't support `flagsChanged`
> > > * some minor checkstyle and spotbugs fixes
> > > 
> > > Really great work! Thanks for doing all the heavy lifting here!
> > 
> > 
> > Thank you, @toddbaert! I really enjoyed working on this issue and learned a lot in the process. Are there any other open issues I could contribute to? I'd love to continue helping out. 😄
> 
> @UtkarshSharma2612 I'm sure there is. I'll let you know.
> 
> I will also invite you to the org, which is the first step to getting more involved.

Sure, Thanks!

aepfli:
Hey @UtkarshSharma2612, it was also a pleasure working with you.

 As you seem to be interested, today I opened https://github.com/open-feature/java-sdk-contrib/issues/957, which would be a cool improvement for our testing tooling. ;)

PR #942, opened 2024-09-07
feat: flow instead of exceptions in resolver

aepfli:
…g not found errors

Exceptions are expensive GOTO statements, and in a performance relevant system not the best to be used for control flow . A FlagNotFound-Error can be expected, but the performance draw backs of exception in this case not.

relates to: https://github.com/open-feature/java-sdk/pull/1095

### DIFF
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolver.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolver.java
@@ -19,7 +19,6 @@ import dev.openfeature.sdk.ProviderEvaluation;
 import dev.openfeature.sdk.ProviderState;
 import dev.openfeature.sdk.Reason;
 import dev.openfeature.sdk.Value;
-import dev.openfeature.sdk.exceptions.FlagNotFoundError;
 import dev.openfeature.sdk.exceptions.ParseError;
 import dev.openfeature.sdk.exceptions.TypeMismatchError;
 import lombok.extern.slf4j.Slf4j;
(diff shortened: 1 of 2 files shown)


Kavindu-Dodan:
This should align with https://github.com/open-feature/java-sdk/pull/1095 to avoid the loss of error hook invocations :)

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolver.java
```diff
@@ -166,12 +167,18 @@ private <T> ProviderEvaluation<T> resolve(Class<T> type, String key,
 
         // missing flag
         if (flag == null) {
-            throw new FlagNotFoundError("flag: " + key + " not found");
+           return ProviderEvaluation.<T>builder()
```
toddbaert:
```suggestion
            return ProviderEvaluation.<T>builder()
```

toddbaert:
> This should align with [open-feature/java-sdk#1095](https://github.com/open-feature/java-sdk/pull/1095) to avoid the loss of error hook invocations :)

I think there's no behavioral change here, since even before the PR you mention, if errorCode was null we threw an exception to cause the error hook invokation.

toddbaert (approved):

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolver.java
toddbaert: I think this is causing your build error.

PR #945, opened 2024-09-10
fix: use keepalive for TCP & use unit in env variable name

### LINKED ISSUE #935: Rename FLAGD_KEEP_ALIVE_TIME to FLAGD_KEEP_ALIVE_TIME_MS
This env_var name is unclear, and inconsistent with the DEADLINE. We should rename it to FLAGD_KEEP_ALIVE_TIME_MS but also still support the old env var.

See: https://github.com/open-feature/java-sdk-contrib/blob/c598d9f0a61f2324fb85d72fdfea34811283c575/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/Config.java#L36

alexandraoberaigner:
Fixes:  https://github.com/open-feature/java-sdk-contrib/issues/935

also fixes the issue that keepAlive was only used by unix socket

### DIFF
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/common/ChannelBuilder.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/common/ChannelBuilder.java
@@ -28,6 +28,10 @@ public class ChannelBuilder {
      */
     @SuppressFBWarnings(value = "PATH_TRAVERSAL_IN", justification = "certificate path is a user input")
     public static ManagedChannel nettyChannel(final FlagdOptions options) {
+
+        // keepAliveTime: Long.MAX_VALUE disables keepAlive; very small values are increased automatically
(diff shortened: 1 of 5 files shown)


providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/common/ChannelBuilder.java
```diff
@@ -48,7 +50,9 @@ public static ManagedChannel nettyChannel(final FlagdOptions options) {
 
         // build a TCP socket
         try {
-            final NettyChannelBuilder builder = NettyChannelBuilder.forAddress(options.getHost(), options.getPort());
+            final NettyChannelBuilder builder = NettyChannelBuilder
+                    .forAddress(options.getHost(), options.getPort())
+                    .keepAliveTime(keepAliveMs, TimeUnit.MILLISECONDS);
```
toddbaert, line 55: :fearful: :fearful: :fearful: :fearful: :fearful:

toddbaert (approved):

PR #948, opened 2024-09-10
chore: add build support for newer JDKs

toddbaert:
This PR adds the ability to build the contribs in the newest versions of Java, but still supports Java 8 which we want in CI for maximum compatibility.

To do this I had to:

- use auto-detection of the JDK version to run the maven-surefire plugin differentially
- add the bytebuddy workarounds as we have in the java-sdk

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -54,6 +54,9 @@
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
         <module-name>${groupId}.${artifactId}</module-name>
         <javadoc.failOnWarnings>true</javadoc.failOnWarnings>
+        <!--  This is required for later correct replacement of surefireArgLine -->
+        <!-- see surefire-java8 and surefire-java9+ profiles -->
+        <surefireArgLine/>
     </properties>
 
     <dependencies>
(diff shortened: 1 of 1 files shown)


beeme1mr (approved):

liran2000:
So #915 should be able to pass with this change?
The PR is sort of blocked since it depends on Java 11.

toddbaert:
> So #915 should be able to pass with this change? The PR is sort of blocked since it depends on Java 11.

No, unfortunately. The problem there is that PR includes usage of the Java 11 HTTP client. That will work fine _locally_ if you run the tests with Java 11+, but the CI still runs in Java 8, on purpose, to ensure compatibility.

We could consider some kind of means of doing multiple test runs, some on Java 8 and some on Java 11/X/Y/Z, and then simply say some providers require more recent Java versions - that would require more work though and we'd have to be very careful.

toddbaert:
> > So #915 should be able to pass with this change? The PR is sort of blocked since it depends on Java 11.
> 
> No, unfortunately. The problem there is that PR includes usage of the Java 11 HTTP client. That will work fine _locally_ if you run the tests with Java 11+, but the CI still runs in Java 8, on purpose, to ensure compatibility.
> 
> We could consider some kind of means of doing multiple test runs, some on Java 8 and some on Java 11/X/Y/Z, and then simply say some providers require more recent Java versions - that would require more work though and we'd have to be very careful.

We could open an issue for this though. it's certainly technically possible.

liran2000:
> > > So #915 should be able to pass with this change? The PR is sort of blocked since it depends on Java 11.
> > 
> > 
> > No, unfortunately. The problem there is that PR includes usage of the Java 11 HTTP client. That will work fine _locally_ if you run the tests with Java 11+, but the CI still runs in Java 8, on purpose, to ensure compatibility.
> > We could consider some kind of means of doing multiple test runs, some on Java 8 and some on Java 11/X/Y/Z, and then simply say some providers require more recent Java versions - that would require more work though and we'd have to be very careful.
> 
> We could open an issue for this though. it's certainly technically possible.

Another option is to make another release version which will compile with newer Java from separate branches, and then both branches and release versions can co-exist.

ryanprayogo (approved):

aepfli (approved):

PR #954, opened 2024-09-16
fix: ConcurrentModificationException on flag config change java 9

toddbaert:
We had a static hashmap that could possibly be modified by multiple threads since `computeIfAbsent` mutates the map.

This is specifically more visible in Java 9 due to some [new additional checks](https://stackoverflow.com/questions/56433061/computeifabsent-throwing-concurrentmodificationexception). This is easily solved by making this static member a local variable.

### DIFF
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/model/FlagParser.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/model/FlagParser.java
@@ -29,10 +29,7 @@ public class FlagParser {
     private static final String FLAG_KEY = "flags";
     private static final String EVALUATOR_KEY = "$evaluators";
     private static final String REPLACER_FORMAT = "\"\\$ref\":(\\s)*\"%s\"";
-
     private static final ObjectMapper MAPPER = new ObjectMapper();
-    private static final Map<String, Pattern> PATTERN_MAP = new HashMap<>();
-
     private static JsonSchema SCHEMA_VALIDATOR;
 
     private FlagParser() {
@@ -100,6 +97,7 @@ public class FlagParser {
 
     private static String transposeEvaluators(final String configuration) throws IOException {
         try (JsonParser parser = MAPPER.createParser(configuration)) {
+            final Map<String, Pattern> patternMap = new HashMap<>();
             final TreeNode treeNode = parser.readValueAsTree();
             final TreeNode evaluators = treeNode.get(EVALUATOR_KEY);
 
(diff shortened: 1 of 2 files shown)


providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/FlagStoreTest.java
```diff
@@ -98,9 +99,9 @@ public void changedFlags() throws Exception {
         Map<String, FeatureFlag> expectedChangedFlags =
                 FlagParser.parseString(getFlagsFromResource(VALID_LONG),true);
         expectedChangedFlags.remove("myBoolFlag");
-        // flags changed from initial VALID_SIMPLE flag
-        Assert.assertEquals(expectedChangedFlags.keySet().stream().collect(Collectors.toList()),
-                storageStateDTOS.take().getChangedFlagsKeys());
+        // flags changed from initial VALID_SIMPLE flag, as a set because we don't care about order
+        Assert.assertEquals(expectedChangedFlags.keySet().stream().collect(Collectors.toSet()),
```
toddbaert, line 103: This assert would inconsistently fail because of ordering inconsistencies.

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/FlagStoreTest.java
```diff
@@ -98,9 +99,9 @@ public void changedFlags() throws Exception {
         Map<String, FeatureFlag> expectedChangedFlags =
                 FlagParser.parseString(getFlagsFromResource(VALID_LONG),true);
```
toddbaert, line 100: This call would intermittently fail in Java 9+ due to concurrent modification exceptions.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/model/FlagParser.java
```diff
@@ -29,10 +29,7 @@ public class FlagParser {
     private static final String FLAG_KEY = "flags";
     private static final String EVALUATOR_KEY = "$evaluators";
     private static final String REPLACER_FORMAT = "\"\\$ref\":(\\s)*\"%s\"";
-
     private static final ObjectMapper MAPPER = new ObjectMapper();
-    private static final Map<String, Pattern> PATTERN_MAP = new HashMap<>();
```
toddbaert, line 34: This field presents a concurrency issue, since multiple threads can possible parse at the same time, and `computeIfAbsent` can throw an concurrent-modification-exception in Java 9+

aepfli (approved):

PR #967, opened 2024-09-24
feat: expose sync-metadata, call RPC with (re)connect

toddbaert:
This PR: 

- adds call to [GetMetadata](https://buf.build/open-feature/flagd/docs/main:flagd.sync.v1#flagd.sync.v1.FlagSyncService.GetMetadata) on gRPC connection init with the in-process provider
  - this returns metadata from flagd which can be used to enrich local (in-process) evaluation
  - the metadata is exposed on the provider in the protected `getSyncMetadata` accessor, and updated when the stream is (re)established, from here it can be used in provider hooks, events, etc
- associated tests and refactors

Please note the call logic may look a bit strange, but I think it's optimal. We call the `getMetadata` RPC as soon as we initiate the stream, and ignore exceptions. This is because it's very unlikely that the metadata RPC will fail while the stream succeeds, and every attempt to reconnect the stream will also retry the metadata - so we should be good with this pattern. If for some reason the metadata call fails but we connect without an issue, we log it.

### DIFF
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
@@ -1,10 +1,10 @@
 package dev.openfeature.contrib.providers.flagd;
 
 import java.util.Collections;
-import java.util.List;
 import java.util.Map;
 
 import dev.openfeature.contrib.providers.flagd.resolver.Resolver;
+import dev.openfeature.contrib.providers.flagd.resolver.common.ConnectionEvent;
 import dev.openfeature.contrib.providers.flagd.resolver.grpc.GrpcResolver;
 import dev.openfeature.contrib.providers.flagd.resolver.grpc.cache.Cache;
 import dev.openfeature.contrib.providers.flagd.resolver.process.InProcessResolver;
@@ -145,17 +145,16 @@ public class FlagdProvider extends EventProvider {
         return this.connected;
     }
 
-    private void onConnectionEvent(boolean newConnectedState, List<String> changedFlagKeys,
-            Map<String, Object> syncMetadata) {
+    private void onConnectionEvent(ConnectionEvent connectionEvent) {
         boolean previous = connected;
-        boolean current = newConnectedState;
-        this.connected = newConnectedState;
-        this.syncMetadata = syncMetadata;
+        boolean current = connected = connectionEvent.isConnected();
+        syncMetadata = connectionEvent.getSyncMetadata();
 
         // configuration changed
         if (initialized && previous && current) {
             log.debug("Configuration changed");
-            ProviderEventDetails details = ProviderEventDetails.builder().flagsChanged(changedFlagKeys)
+            ProviderEventDetails details = ProviderEventDetails.builder()
+                    .flagsChanged(connectionEvent.getFlagsChanged())
                     .message("configuration changed").build();
             this.emitProviderConfigurationChanged(details);
             return;
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
@@ -22,7 +22,7 @@ import lombok.extern.slf4j.Slf4j;
 @Slf4j
 @SuppressWarnings({ "PMD.TooManyStaticImports", "checkstyle:NoFinalizer" })
 public class FlagdProvider extends EventProvider {
-    private static final String FLAGD_PROVIDER = "flagd Provider";
+    private static final String FLAGD_PROVIDER = "flagd";
     private final Resolver flagResolver;
     private volatile boolean initialized = false;
     private volatile boolean connected = false;
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
@@ -22,7 +22,7 @@ import lombok.extern.slf4j.Slf4j;
 @Slf4j
 @SuppressWarnings({ "PMD.TooManyStaticImports", "checkstyle:NoFinalizer" })
 public class FlagdProvider extends EventProvider {
-    private static final String FLAGD_PROVIDER = "flagD Provider";
+    private static final String FLAGD_PROVIDER = "flagd Provider";
     private final Resolver flagResolver;
     private volatile boolean initialized = false;
     private volatile boolean connected = false;
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
@@ -1,6 +1,8 @@
 package dev.openfeature.contrib.providers.flagd;
 
+import java.util.Collections;
 import java.util.List;
+import java.util.Map;
 
 import dev.openfeature.contrib.providers.flagd.resolver.Resolver;
 import dev.openfeature.contrib.providers.flagd.resolver.grpc.GrpcResolver;
@@ -24,6 +26,7 @@ public class FlagdProvider extends EventProvider {
     private final Resolver flagResolver;
     private volatile boolean initialized = false;
     private volatile boolean connected = false;
+    private volatile Map<String, Object> syncMetadata = Collections.emptyMap();
 
     private EvaluationContext evaluationContext;
 
@@ -47,13 +50,13 @@ public class FlagdProvider extends EventProvider {
         switch (options.getResolverType().asString()) {
             case Config.RESOLVER_IN_PROCESS:
                 this.flagResolver = new InProcessResolver(options, this::isConnected,
-                        this::onResolverConnectionChanged);
+                        this::onConnectionEvent);
                 break;
             case Config.RESOLVER_RPC:
                 this.flagResolver = new GrpcResolver(options,
                         new Cache(options.getCacheType(), options.getMaxCacheSize()),
                         this::isConnected,
-                        this::onResolverConnectionChanged);
+                        this::onConnectionEvent);
                 break;
             default:
                 throw new IllegalStateException(
@@ -117,6 +120,19 @@ public class FlagdProvider extends EventProvider {
         return this.flagResolver.objectEvaluation(key, defaultValue, mergeContext(ctx));
     }
 
+    /**
+     * An unmodifiable view of an object map representing the latest result of the
+     * SyncMetadata.
+     * Set on initial connection and updated with every reconnection.
+     * see:
+     * https://buf.build/open-feature/flagd/docs/main:flagd.sync.v1#flagd.sync.v1.FlagSyncService.GetMetadata
+     * 
+     * @return Object map representing sync metadata
+     */
+    protected Map<String, Object> getSyncMetadata() {
+        return Collections.unmodifiableMap(syncMetadata);
+    }
+
     private EvaluationContext mergeContext(final EvaluationContext clientCallCtx) {
         if (this.evaluationContext != null) {
             return evaluationContext.merge(clientCallCtx);
@@ -129,10 +145,12 @@ public class FlagdProvider extends EventProvider {
         return this.connected;
     }
 
-    private void onResolverConnectionChanged(boolean newConnectedState, List<String> changedFlagKeys) {
+    private void onConnectionEvent(boolean newConnectedState, List<String> changedFlagKeys,
+            Map<String, Object> syncMetadata) {
         boolean previous = connected;
         boolean current = newConnectedState;
         this.connected = newConnectedState;
+        this.syncMetadata = syncMetadata;
 
         // configuration changed
         if (initialized && previous && current) {
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/common/Convert.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/common/Convert.java
@@ -1,0 +1,184 @@
+package dev.openfeature.contrib.providers.flagd.resolver.common;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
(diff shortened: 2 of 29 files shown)


providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
```diff
@@ -117,6 +120,19 @@ public ProviderEvaluation<Value> getObjectEvaluation(String key, Value defaultVa
...
+    /**
+     * An unmodifiable view of an object map representing the latest result of the
+     * SyncMetadata.
+     * Set on initial connection and updated with every reconnection.
+     * see:
+     * https://buf.build/open-feature/flagd/docs/main:flagd.sync.v1#flagd.sync.v1.FlagSyncService.GetMetadata
+     * 
+     * @return Object map representing sync metadata
+     */
+    protected Map<String, Object> getSyncMetadata() {
```
toddbaert, line 132: We basically just update the field backing this whenever we (re)connect to the sync.service.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/common/Convert.java
```diff
@@ -0,0 +1,184 @@
+package dev.openfeature.contrib.providers.flagd.resolver.common;
```
toddbaert, line 1: Not a single method in this class is new; they were all extracted from [here](https://github.com/open-feature/java-sdk-contrib/pull/967/files#diff-e5ada6a457791691f43b3f36c1cea2708bcf029ea0d13a8e5c6044c133261f8aL203) to avoid duplication.

They are are simple static functions to convert from gRPC data types to SDK data types.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/grpc/GrpcStreamConnector.java
```diff
@@ -123,8 +127,27 @@ static void observeEventStream(final BlockingQueue<StreamPayload> writeTo,
...
+            try {
+                GetMetadataResponse metadataResponse = serviceBlockingStub.getMetadata(metadataRequest.build());
+                metadata = convertProtobufMapToStructure(metadataResponse.getMetadata().getFieldsMap()).asObjectMap();
+            } catch (Exception e) {
+                // the chances this call fails but the syncRequest does not are slim
+                // it could be that the server doesn't implement this RPC
+                // instead of logging here, retain the exception and only log if the
+                // streamReceiver doesn't error
+                metadataException = e;
+            }
```
toddbaert: This is the meat of the new change... we call the getMetadata RPC as soon as we start the stream.

_IF_ the sync-metadata somehow fails but the stream still connects, we log it below; but this shouldn't happen since both use the same underlying channel.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/grpc/GrpcStreamConnector.java
```diff
@@ -50,6 +58,7 @@ public class GrpcStreamConnector implements Connector {
     public GrpcStreamConnector(final FlagdOptions options) {
         channel = ChannelBuilder.nettyChannel(options);
         serviceStub = FlagSyncServiceGrpc.newStub(channel);
+        serviceBlockingStub = FlagSyncServiceGrpc.newBlockingStub(channel);
```
toddbaert, line 56: Add new stub for calling getSyncMetadata, which uses the same underlying netty channel.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/grpc/GrpcStreamConnector.java
```diff
@@ -150,11 +174,18 @@ static void observeEventStream(final BlockingQueue<StreamPayload> writeTo,
...
+                        new QueuePayload(QueuePayloadType.DATA, data, metadata))) {
                     log.error("Stream writing failed");
                 }
 
+                if (metadataException != null) {
+                    // if we somehow are connected but the metadata call failed, something strange
+                    // happened
+                    log.error("Stream connected but getMetadata RPC failed", metadataException);
+                }
+
```
toddbaert: Take note of this.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/grpc/GrpcConnector.java
```diff
@@ -37,7 +38,7 @@ public class GrpcConnector {
     private final long deadline;
 
     private final Cache cache;
-    private final BiConsumer<Boolean, List<String>> stateConsumer;
+    private final TriConsumer<Boolean, List<String>, Map<String, Object>> onConnectionEvent;
```
aepfli: is maybe now the time, to use a pojo for this and a normal consumer?

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/grpc/GrpcConnector.java
```diff
@@ -155,16 +156,16 @@ private void observeEventStream() {
         }
 
         log.error("failed to connect to event stream, exhausted retries");
-        this.grpcStateConsumer(false, null);
+        this.grpconConnectionEvent(false, Collections.emptyList());
     }
 
-    private void grpcStateConsumer(final boolean connected, final List<String> changedFlags) {
+    private void grpconConnectionEvent(final boolean connected, final List<String> changedFlags) {
```
aepfli:
```suggestion
    private void grpcOnConnectionEvent(final boolean connected, final List<String> changedFlags) {
```

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/grpc/EventStreamObserver.java
```diff
@@ -20,25 +20,25 @@
...
 @SuppressFBWarnings(justification = "cache needs to be read and write by multiple objects")
 class EventStreamObserver implements StreamObserver<EventStreamResponse> {
-    private final BiConsumer<Boolean, List<String>> stateConsumer;
+    private final BiConsumer<Boolean, List<String>> onConnectionEvent;
     private final Object sync;
     private final Cache cache;
 
-    private static final String CONFIGURATION_CHANGE = "configuration_change";
-    private static final String PROVIDER_READY = "provider_ready";
+    public static final String CONFIGURATION_CHANGE = "configuration_change";
```
aepfli: [nit] maybe we should put those into an own class/enum

aepfli (approved):
lgtm - but I am unsure if we should switch to a POJO rather than increase the number of consumers parameters again. Strangely, this is the second time within a month, and I think this could be easier with a simple consumer and a POJO

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/grpc/GrpcStreamConnector.java
```diff
@@ -150,11 +174,18 @@ static void observeEventStream(final BlockingQueue<StreamPayload> writeTo,
...
                 if (!writeTo.offer(
-                        new StreamPayload(StreamPayloadType.DATA, data))) {
+                        new QueuePayload(QueuePayloadType.DATA, data, metadata))) {
                     log.error("Stream writing failed");
                 }
 
+                if (metadataException != null) {
+                    // if we somehow are connected but the metadata call failed, something strange
+                    // happened
+                    log.error("Stream connected but getMetadata RPC failed", metadataException);
```
guidobrei: Should we reset `metadataException` to not log in every iteration?

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/grpc/GrpcConnector.java
guidobrei: you mean a pojo for the metadata map, or a proper named functional interface for the TriConsumer?

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/grpc/GrpcStreamConnector.java
toddbaert: Ah, yes, we certainly should.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/grpc/GrpcConnector.java
aepfli: a pojo for boolean, list and map - one to bind them all ;)

toddbaert: We may actually already have such a POJO :sweat_smile: 

I think I'll make this change.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/grpc/GrpcStreamConnector.java
toddbaert: Oh actually we already so this, since this exception is declared inside the outer while loop.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/grpc/GrpcConnector.java
aepfli: see also https://github.com/open-feature/java-sdk-contrib/pull/925#discussion_r1743094379 - from not so long ago

toddbaert: Done: https://github.com/open-feature/java-sdk-contrib/pull/967/commits/667946fba576537e134f87ba85e10c2c77f11f66

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/grpc/GrpcStreamConnector.java
```diff
@@ -123,8 +125,29 @@ static void observeEventStream(final BlockingQueue<StreamPayload> writeTo,
...
-            serviceStub.syncFlags(request, new GrpcStreamHandler(streamReceiver));
+            final SyncFlagsRequest.Builder syncRequest = SyncFlagsRequest.newBuilder();
+            final GetMetadataRequest.Builder metadataRequest = GetMetadataRequest.newBuilder();
+            Map<String, Object> metadata = Collections.emptyMap();
+
+            if (selector != null) {
+                syncRequest.setSelector(selector);
+            }
+
+            serviceStub.withDeadlineAfter(deadline, TimeUnit.MILLISECONDS).syncFlags(syncRequest.build(),
```
toddbaert: @guidobrei @aepfli I found we weren't using the passed deadline on either of the gRPC calls in this class, so I passed that along as well and added related assertions in existing tests.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/grpc/GrpcConnector.java
```diff
@@ -104,7 +105,7 @@ public void shutdown() throws Exception {
                 this.channel.awaitTermination(this.deadline, TimeUnit.MILLISECONDS);
                 log.warn(String.format("Unable to shut down channel by %d deadline", this.deadline));
             }
-            this.stateConsumer.accept(false, Collections.emptyList());
+            this.onConnectionEvent.accept(new ConnectionEvent(false, Collections.emptyList(), Collections.emptyMap()));
```
aepfli: Do we need the empty map and list here? Can we maybe add this also as defaults to the connectionEvent?

aepfli (approved):

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/grpc/EventStreamObserver.java
toddbaert: Done: https://github.com/open-feature/java-sdk-contrib/pull/967/commits/31860104cda6f54a3f1d8bc94f77c5884a9028b1

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/grpc/GrpcConnector.java
toddbaert: Oh we do not. I actually have other constructors, I'm just not using them here. Will update.

toddbaert: Done, I think this was the only case.

guidobrei (approved):

aepfli:
We should also document this feature within the providers readme - i think this is important to be aware of.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
```diff
@@ -20,10 +22,11 @@
 @Slf4j
 @SuppressWarnings({ "PMD.TooManyStaticImports", "checkstyle:NoFinalizer" })
 public class FlagdProvider extends EventProvider {
-    private static final String FLAGD_PROVIDER = "flagD Provider";
+    private static final String FLAGD_PROVIDER = "flagd Provider";
```
beeme1mr: Should we drop the provider suffix?

toddbaert: ya, probably. I'll update.

toddbaert:
```suggestion
    private static final String FLAGD_PROVIDER = "flagd";
```

PR #991, opened 2024-09-30
feat!: context enrichment via contextEnricher, not from init

toddbaert:
- :warning: provider no longer merges context passed at init into each evaluation (this was non-standard behavior
- `contextEnricher` option added which does the above, and more
  - runs with every (re)connection
  - updates per-evaluation context, optionally with data from sync-metadata response.

Continuation of https://github.com/open-feature/java-sdk-contrib/pull/967.


providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
```diff
@@ -97,48 +110,44 @@ public Metadata getMetadata() {
...
      * SyncMetadata.
      * Set on initial connection and updated with every reconnection.
      * see:
      * https://buf.build/open-feature/flagd/docs/main:flagd.sync.v1#flagd.sync.v1.FlagSyncService.GetMetadata
      * 
      * @return Object map representing sync metadata
      */
-    protected Map<String, Object> getSyncMetadata() {
-        return Collections.unmodifiableMap(syncMetadata);
+    protected Structure getSyncMetadata() {
```
toddbaert, line 145: This is a much safer type which we already use for this sort of thing.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
```diff
@@ -149,6 +158,7 @@ private void onConnectionEvent(ConnectionEvent connectionEvent) {
         boolean previous = connected;
         boolean current = connected = connectionEvent.isConnected();
         syncMetadata = connectionEvent.getSyncMetadata();
+        enrichedContext = contextEnricher.apply(connectionEvent.getSyncMetadata());
```
toddbaert, line 165: Every time the resolver connects, we update the cached context by re-running the passed enricher.

providers/flagd/README.md
```diff
@@ -177,6 +177,12 @@ By default, the provider is configured to
 use [least recently used (lru)](https://commons.apache.org/proper/commons-collections/apidocs/org/apache/commons/collections4/map/LRUMap.html)
 caching with up to 1000 entries.
 
+##### Context enrichment
```
toddbaert, line 180: This can be used to select a subset of fields from the `getMetadata` response. By default, all of it is injected into the context.

providers/flagd/README.md
```diff
@@ -177,6 +177,12 @@ By default, the provider is configured to
...
+
+The `contextEnricher` option is a function which provides a context to be added to each evaluation.
+This function runs on the initial provider connection and ever reconnection, and is passed the [sync-metadata](#sync-metadata).
```
```suggestion
This function runs on the initial provider connection and every reconnection, and is passed the [sync-metadata](#sync-metadata).
```

providers/flagd/README.md
```diff
@@ -177,6 +177,12 @@ By default, the provider is configured to
 use [least recently used (lru)](https://commons.apache.org/proper/commons-collections/apidocs/org/apache/commons/collections4/map/LRUMap.html)
 caching with up to 1000 entries.
 
+##### Context enrichment
+
+The `contextEnricher` option is a function which provides a context to be added to each evaluation.
+This function runs on the initial provider connection and every reconnection, and is passed the [sync-metadata](#sync-metadata).
+By default, a simple implementation which uses the sync-metadata payload its entirety is used.
```
```suggestion
By default, a simple implementation which uses the sync-metadata payload in its entirety is used.
```

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/SyncMetadataHook.java
```diff
@@ -0,0 +1,26 @@
...
+class SyncMetadataHook implements Hook<Object> {
+
+    private Supplier<EvaluationContext> contextSupplier;
+
+    SyncMetadataHook(Supplier<EvaluationContext> contextSupplier) {
+        this.contextSupplier = contextSupplier;
+    }
+
+    /**
+     * Return the enriched context, including the additional attributes from the sync-metadata.
```
aepfli: I am not sure this doc is correct; we're not enriching context; we're only returning the result of the supplier - should we merge this with the provided ctx?

toddbaert: No, we shouldn't merge anything, because the SDK will merge this result into things before the provider evaluates; that was the weird thing about the previous implementation; the SDK should do all the merging.

The comment is wrong though, as you point out. I will update it.

toddbaert:
```suggestion
     * Return the context adapted from the sync-metadata provided by the supplier.
```

aepfli (approved):

beeme1mr (approved):

PR #998, opened 2024-10-03
build(flagd): OS agnostic flagd provider build

guidobrei:
## This PR
removes pom.xml dependency on `cp` command from `flagd` provider maven build.

Proposed changes:
- Compile GRPC proto files directly from schemas repository
- Use copy resources maven plugin to copy JSON schema files
- Improve test logic to load resource file contents

### Notes
Flagd Provider uses plain `cp` commands in maven build. This causes problems on Windows systems.

### How to test

`./mvnw verify` will build the project (also on windows)


providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/resolver/process/TestUtils.java
```diff
@@ -15,21 +18,23 @@ public class TestUtils {
     public static final String INVALID_CFG = "flagConfigurations/invalid-configuration.json";
     public static final String UPDATABLE_FILE = "flagConfigurations/updatableFlags.json";
 
-
     public static String getFlagsFromResource(final String file) throws IOException {
-        final URL url = FlagParser.class.getClassLoader().getResource(file);
-        if (url == null) {
-            throw new IllegalStateException(String.format("Resource %s not found", file));
-        } else {
-            return new String(Files.readAllBytes(Paths.get(url.getPath())));
```
guidobrei, line 24: `url.getPath` creates a path string representation, that is not accepted by `Paths.get(String)` on Windows (e.g. `/C:/sample.csv`).

See [this issue](https://bugs.openjdk.org/browse/JDK-8197918).

aepfli (approved):

providers/flagd/src/main/resources/flags.json
```diff
@@ -1,184 +0,0 @@
-{
-  "$id": "https://flagd.dev/schema/v0/flags.json",
```
toddbaert, line 2: I pushed a small change to some test files that use this schema; they now use the pulled/gitignored one.

The test files using this schema only use it for editor support (so editing the schemas locally has editor support) not in actual unit tests.

toddbaert (approved):

toddbaert:
This is a helpful additional and makes the POM simpler too. Thanks!

providers/flagd/src/main/resources/flags.json
guidobrei: Ahh, there they have been used. I was wondering why there have been JSON schemas checked in in resources when they're copied to a different resource directory on build.

PR #999, opened 2024-10-03
feat: gRPC stream connection deadline

guidobrei:
## This PR
adds deadlines to GRPC stream connections in the flagd provider.

In GRPC setups, where L7 proxies route traffic between the provider and flagd, broken connections in the line can cause a provider never to get updates of chaged flag configurations. The provider never recognizes broken e2e connections because it only listens on the RPC stream and never sends data. 
This feature forces a client to open a new GRPC stream after the deadline is exceeded. 

### Notes

Default stream deadline is 10 Minutes. Deadlines can be disabled by setting it to `0`.


providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/grpc/GrpcStreamConnector.java
```diff
@@ -55,6 +56,7 @@ public GrpcStreamConnector(final FlagdOptions options) {
         serviceStub = FlagSyncServiceGrpc.newStub(channel);
         serviceBlockingStub = FlagSyncServiceGrpc.newBlockingStub(channel);
         deadline = options.getDeadline();
+        streamDeadlineMs = options.getStreamDeadlineMs();
```
aepfli, line 61: [suggestion] Cant we just use the options within this class, and pass it further down the call chain? - adding a new field and parameters might be tedious over time

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/grpc/GrpcConnector.java
```diff
@@ -64,6 +65,7 @@ public GrpcConnector(final FlagdOptions options, final Cache cache, final Suppli
         this.startEventStreamRetryBackoff = options.getRetryBackoffMs();
         this.eventStreamRetryBackoff = options.getRetryBackoffMs();
         this.deadline = options.getDeadline();
+        this.streamDeadlineMs = options.getStreamDeadlineMs();
```
aepfli, line 68: [suggestion] using the options object instead of separate fields would reduce this over head of adding a new field all the time

aepfli (approved):

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdOptions.java
```diff
@@ -92,6 +92,14 @@ public class FlagdOptions {
     @Builder.Default
     private int deadline = fallBackToEnvOrDefault(Config.DEADLINE_MS_ENV_VAR_NAME, Config.DEFAULT_DEADLINE);
 
+    /**
+     * Streaming connection deadline in milliseconds.
+     * Set to 0 to disable the deadline.
```
toddbaert, line 97:
```suggestion
     * Streaming connection deadline in milliseconds.
     * Set to 0 to disable the deadline.
     * Defaults to 600000 (10 minutes); recommended to prevent infrastructure from killing idle connections.
```

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/grpc/GrpcConnector.java
toddbaert: I can go either way on this one. In fact it might be a good thing to do in a pure refactor/cleanup PR. There's also some naming that we can probably improve with the provider.

toddbaert (approved):

PR #1019, opened 2024-10-10
feat: tolerate immediately recoverable stream faults, improve logging

### LINKED ISSUE #1010: [flagd] improve error logging for reconnect scenarios
Our FlagD in-process provider is built with resilience. We will try to reconnect when there is a loss of connection, which is excellent and the right thing to do in a distributed world. We want to be able to continue to serve flags properly. The Problem we're facing is that each reconnect will log an error, but is it an error if we recover from this state? IMHO, it is an error if we reach our maximum reconnect delay.

I propose to change our log level [here](https://github.com/open-feature/java-sdk-contrib/blob/main/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/grpc/GrpcStreamConnector.java#L179) to info, and additionally log an error, if we reach the max delay [here](https://github.com/open-feature/java-sdk-contrib/blob/main/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/grpc/GrpcStreamConnector.java#L217-L219).

https://github.com/open-feature/java-sdk-contrib/blob/2a8ea6c8e7f0eba248ca003cf1afa6d0410a68d2/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/grpc/GrpcStreamConnector.java#L178-L186

This way, we handle reconnects gracefully but will still retrieve the information about connection issues in a timely manner in an error case.

Additionally/Optionally, we can add an immediate error log for the first connection attempts. In this case, we might not want to wait for the maximum delay.

## Goals

- reducing normal reconnection logs to info or warn
- separate logs with dedicated message for metadata or stream
- log the error if we reach max delay
- Optional: log the error immediately if we're in the first connection.

guidobrei:
## This PR
- Improves error logging for in process resolver with remote mode.
- Harmonizes backoff implementations across different gRPC handlers.
- Uses `FlagdOptions.getRetryBackoffMs()` to initialize the backoff in all Backoff scenarios. `GrpcStreamConnector` previously used a hardcoded value of 2 seconds.
- Immediately reconnect on first stream error in `GrpcStreamConnector`. This removes a backoff when a planned deadline exceeds and the connector reconnects.
- Unified standard max jitter of 250ms for all backoff use-cases

Fixes #1010 

### Notes

Different to #1010, error logs are not written when the max retry delay is reached, but already at the second error in a row.
Waiting for max retry delay (120 seconds) with exponential backoff starting with 2 seconds would require 126 seconds until the first error gets visible.

Instead, error logs are generated whenever an error queue payload is emitted. Only on the first error we try to reconnect immediately without any backoff (only with default jitter 250ms max) and without emitting an error payload. Starting with the second error in a row we log an error and emit the error payload.

The initial Backoff is now `FlagdOptions.getRetryBackoffMs()` in `GrpcStreamConnector` (new) and `GrpcConnector` (no change). 
For the `GrpcStreamConnector` this means an initial Backoff of 1 sec (default option) instead of 2 secs.

I've also removed the special handling of `DEADLINE_EXCEEDED' errors, as the connector now tries to reconnect silently on any first error. This also solves `DEADLINE_EXCEEDED` issues related to Envoy, where a wrong gRPC status code is reported. See [here](https://github.com/envoyproxy/envoy/issues/3619)

With the first immediate retry the new Backoff times for `GrpcStreamConnector` are now:
- 0s
- 1s
- 2s
- 4s
- 8s
- ...
- 120s


toddbaert:
Hey @guidobrei ... so the reason the e2e test is failing here is that we restart the "unstable" container too fast - it's not down long enough for your new implementation to see 2 failures. I've fixed that with container version `v0.5.10` - so your changes to the in-process mode here here work fine.

Unfortunately, that exposes some other bug. It seems like adding the deadline to to the RPC stream a few commits ago is causing some kind of unexpected failure with reconnection for the RPC resolver if flagd is down for a few seconds. When I comment out the stream deadline in THAT resolver, both e2e test suites work as expected. :confounded: 

I don't know if this is a gRPC bug, flagd bug, or a bug with our provider. Please check out my PR to your branch [here](https://github.com/guidobrei/open-feature-java-sdk-contrib/pull/2).

If we can't get to the bottom of it soon, I suggest we disable the deadline for the RPC version for now, since it's a new feature anyway, and tackle it later; it may be bugged in the wild.

I've done this in m PR.

toddbaert:
> Hey @guidobrei ... so the reason the e2e test is failing here is that we restart the "unstable" container too fast - it's not down long enough for your new implementation to see 2 failures. I've fixed that with container version `v0.5.10` - so your changes to the in-process mode here here work fine.
> 
> Unfortunately, that exposes some other bug. It seems like adding the deadline to to the RPC stream a few commits ago is causing some kind of unexpected failure with reconnection for the RPC resolver if flagd is down for a few seconds. When I comment out the stream deadline in THAT resolver, both e2e test suites work as expected. 😖
> 
> I don't know if this is a gRPC bug, flagd bug, or a bug with our provider. Please check out my PR to your branch [here](https://github.com/guidobrei/open-feature-java-sdk-contrib/pull/2).
> 
> If we can't get to the bottom of it soon, I suggest we disable the deadline for the RPC version for now, since it's a new feature anyway, and tackle it later; it may be bugged in the wild.

It almost seems like adding a deadline delays the exceptions... which is odd because no deadline (infinite) throws exceptions around connections immediately.

cc @aepfli

toddbaert:
Wow... so [this](https://grpc.github.io/grpc/core/md_doc_connection-backoff.html) is illuminating. Apparently, deadlines can impact the reconnect logic of the underlying CONNECTION (not stream). gRPC has an underlying connection retry mechanism independent of the stream reconnect. I seems at least in grpc-java, our deadline, as well as the fact our testbed is now down for 5 seconds (instead of just for an instant), has had an impact on that.

I'm convinced at this point that the changes in our deadlines and to bring flagd down for 5s (see [resolved comment](https://github.com/open-feature/java-sdk-contrib/pull/1019#issuecomment-2406225069) above) instead of just for a second are what broke our e2e tests. The simple solution was to change the cadence of the "unstable" containers... instead of being down for 5s and up for 5s, we go down for 5s and stay up for 20s... considering the connection backoff algorithm linked above this sensibly improves the reliability of the test; before this change I couldn't make the test pass; now it passes consistently; the downside is the tests can take longer if we are unlucky - for that reason I've updated the reconnect tests to wait a max of 240s (though they generally complete much before that). See [my changes](https://github.com/open-feature/java-sdk-contrib/pull/1019/commits/6d9b7f8e2a7042a692074744ec2a2db3a2002245).

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/common/backoff/BackoffService.java
```diff
@@ -0,0 +1,90 @@
...
+
+import lombok.Getter;
+
+import java.util.concurrent.ThreadLocalRandom;
+
+/**
+ * A service that provides backoff functionality.
+ */
+public class BackoffService {
+    public static final int DEFAULT_MAX_JITTER = 0x1 << 8; // 256; Random likes boundaries that are a power of 2
```
toddbaert, line 11: :sunglasses:

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/common/backoff/BackoffService.java
```diff
@@ -0,0 +1,90 @@
+package dev.openfeature.contrib.providers.flagd.resolver.common.backoff;
+
+import lombok.Getter;
+
+import java.util.concurrent.ThreadLocalRandom;
+
+/**
+ * A service that provides backoff functionality.
+ */
+public class BackoffService {
```
toddbaert, line 10: This is a nice refactor, thanks.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/common/backoff/CombinedBackoff.java
```diff
@@ -0,0 +1,73 @@
+package dev.openfeature.contrib.providers.flagd.resolver.common.backoff;
+
+import lombok.Getter;
+
+/**
```
toddbaert, line 5: Also a nice abstraction.

toddbaert:
Could you inject your backoff stuff into the `EventStreamObserver.java` (for the RPC resolver) and take advantage of the same logging/eventing rules there too, for consistency (meaning immediately and silently retry 1 time)? And could you remove the `DEADLINE_EXCEEDING` exception logic there as well? https://github.com/guidobrei/open-feature-java-sdk-contrib/blob/801e5d0e4ee3dc8adaf0cbb79878279c996899f6/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/grpc/EventStreamObserver.java#L58-L68

toddbaert (approved):
Approved, but consider: https://github.com/open-feature/java-sdk-contrib/pull/1019#issuecomment-2406470222

toddbaert:
@guidobrei I updated this by rebasing on main, sorry! 

`git reset --hard HEAD~10 && git pull` should sync your local branch back up with this.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/common/backoff/GrpcStreamConnectorBackoffService.java
```diff
@@ -0,0 +1,40 @@
+package dev.openfeature.contrib.providers.flagd.resolver.common.backoff;
+
+/**
+ * Backoff service that supports "silent" backoff.
+ */
+public class GrpcStreamConnectorBackoffService extends BackoffService {
```
guidobrei, line 6: Moved this class from the `remote` to `common.backoff` backage.

aepfli (approved):

toddbaert (approved):
Great work! Nice to see this improvement. I will soon use this as the basis for some new specifications in the [flagd provider specs](https://flagd.dev/reference/specifications/in-process-providers/).

PR #1028, opened 2024-10-21
feat: multi-provider implementation

liran2000:
[Readme ](https://github.com/open-feature/java-sdk-contrib/blob/dbf0c5cd63c4d005a481d83ecbc112ec1a554c02/providers/multiprovider/README.md#openfeature-multi-provider-for-java) describes the provider.

### References

- [Spec](https://github.com/open-feature/spec/blob/main/specification/appendix-a-included-utilities.md#multi-provider)
- [blog post](https://openfeature.dev/blog/openfeature-multi-provider-release/) by @emmawillis


providers/multiprovider/src/main/java/dev/openfeature/contrib/providers/multiprovider/FirstSuccessfulStrategy.java
```diff
@@ -0,0 +1,38 @@
...
+    public FirstSuccessfulStrategy(Map<String, FeatureProvider> providers) {
+        super(providers);
+    }
+
+    @Override
+    public <T> ProviderEvaluation<T> evaluate(String key, T defaultValue, EvaluationContext ctx,
+              Function<FeatureProvider, ProviderEvaluation<T>> providerFunction) {
+        for (FeatureProvider provider: getProviders().values()) {
+            try {
+                return providerFunction.apply(provider);
```
aepfli: Not all providers throw exceptions sometimes, they return the information in the provider evaluation instead of throwing as exceptions for flows is expensive see https://github.com/open-feature/java-sdk/pull/1095 - applies to all strategies

providers/multiprovider/src/main/java/dev/openfeature/contrib/providers/multiprovider/FirstMatchStrategy.java
```diff
@@ -0,0 +1,50 @@
...
+ * In all other cases, use the value returned by the provider.
+ * If any provider returns an error result other than FLAG_NOT_FOUND, the whole evaluation should error and
+ * “bubble up” the individual provider’s error in the result.
+ * As soon as a value is returned by a provider, the rest of the operation should short-circuit and not call
+ * the rest of the providers.
+ */
+@Slf4j
+public class FirstMatchStrategy extends BaseStrategy {
+
+    public FirstMatchStrategy(Map<String, FeatureProvider> providers) {
```
aepfli: This might be a problem, as the map is generated in the multi-provider; we need these values to be generated before we pass the strategy to the provider. Maybe we should also put those as parameters to the `evaluate` method else it will be hard to use any other strategy than the default

aepfli (commented):
Thank you, this looks really cool. But especially the Strategy initialization and the handling of 'exception-less Errors' might be problematic

providers/multiprovider/src/main/java/dev/openfeature/contrib/providers/multiprovider/FirstMatchStrategy.java
liran2000: I updated it a bit. The providers can be reached via _BaseStrategy_, as done at [MultiProviderTest](https://github.com/liran2000/java-sdk-contrib/blob/daeefe93d86910065bbdcd38f641d0d353c4c506/providers/multiprovider/src/test/java/dev/openfeature/contrib/providers/multiprovider/MultiProviderTest.java#L172). What do you think ?

providers/multiprovider/src/main/java/dev/openfeature/contrib/providers/multiprovider/FirstSuccessfulStrategy.java
liran2000: updated accordingly.

providers/multiprovider/src/main/java/dev/openfeature/contrib/providers/multiprovider/FirstMatchStrategy.java
aepfli: i am not sure, now we do have to separate maps where the state needs to be in sync. this might cause problems. we should centralize this on the provider, what if we change the signature of the `evaluate` method like

```java
public interface Strategy {
    <T> ProviderEvaluation<T> evaluate(Map<...> providers, String key, T defaultValue, EvaluationContext ctx,
           Function<FeatureProvider, ProviderEvaluation<T>> providerFunction);
}

Comment (liran2000) [reply providers/multiprovider/src/main/java/dev/openfeature/contrib/providers/multiprovider/FirstMatchStrategy.java]: it is commonly the same providers instance at the multi-provider and the strategy, and it is built as unmodifiable when initialized. Why can this be a problem ? I am fine with your approach above too, just want to understand better first.

another thing, reg. _ProviderEvaluation<T>> providerFunction_ as parameter:
it is pretty straight-forward to implement with it, but I have some concerns that if a custom strategy implementation want to do change the function call and/or arguments, it will limit it. Other approach can be to pass the type. e.g. Boolean.class instead of the function, and convert it to the specific evaluation method. What do you think ?

File: providers/multiprovider/CHANGELOG.md
Comment (beeme1mr) [inline providers/multiprovider/CHANGELOG.md:1]: Please remove this changelog. A new one should be created by Release Please.

File: release-please-config.json [hunk 3]
```diff
@@ -100,6 +100,17 @@
         "README.md"
       ]
     },
+    "providers/multi-provider": {
```
beeme1mr: The file path is `providers/multiprovider`. Please either update the path name or this configuration.

providers/multiprovider/src/main/java/dev/openfeature/contrib/providers/multiprovider/FirstMatchStrategy.java
aepfli: if i eg. switch to the FirstSuccessful implementation my code would look like 
```java
var providers = List.of(...);
var strategy = new FirstSuccessful(providers);
var provider = new MultiProvider(providers, strategy);
```

but that means that theoretically the providers list could be a different one, or am i missing here something?

Regarding your providerFunction, i do think that this is the clearest way also maintainence wise - trying to hassle with type information adds unnecessary overhead in my opinion

providers/multiprovider/CHANGELOG.md
liran2000: done

release-please-config.json
liran2000: done

providers/multiprovider/src/main/java/dev/openfeature/contrib/providers/multiprovider/FirstMatchStrategy.java
liran2000: ok, updated accordingly.

release-please-config.json
liran2000: BTW, originally I had "multi-provider" like other providers, but it failed validation on [Invalid Automatic-Module-Name](https://sormuras.github.io/blog/2018-11-16-invalid-automatic-module-names.html):
> ❌ no - allowed

beeme1mr: Why would Maven Central allow invalid names to be uploaded?

toddbaert: Maybe it's because they aren't invalid in other JVM langauges, like Scala or Kotlin, and maven supports all of these.

providers/multiprovider/src/test/java/dev/openfeature/contrib/providers/multiprovider/MultiProviderTest.java
```diff
@@ -0,0 +1,186 @@
...
+
+class MultiProviderTest {
+
+    @BeforeEach
+    void setUp() {
+    }
+
+    @SneakyThrows
+    @Test
+    public void testInit() {
```
toddbaert, line 36: I think we need 2 more tests for init:

- one where both providers implement a successful init() function
- one where one provider's init functions throws (and init should throw).

providers/multiprovider/src/main/java/dev/openfeature/contrib/providers/multiprovider/MultiProvider.java
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    /**
+     * Initialize the provider.
+     * @param evaluationContext evaluation context
+     * @throws Exception on error
+     */
+    @Override
+    public void initialize(EvaluationContext evaluationContext) throws Exception {
+        JSONObject json = new JSONObject();
```
toddbaert, line 80: Instead of JSON object, could we use a `HashMap`, or our SDK's own build-in object representation ([Structure](https://github.com/open-feature/java-sdk/blob/main/src/main/java/dev/openfeature/sdk/Structure.java) and it's implementations)?

providers/multiprovider/src/main/java/dev/openfeature/contrib/providers/multiprovider/MultiProvider.java
```diff
@@ -0,0 +1,133 @@
...
+        json.put("name", NAME);
+        JSONObject providersMetadata = new JSONObject();
+        json.put("originalMetadata", providersMetadata);
+        for (FeatureProvider provider: providers.values()) {
+            provider.initialize(evaluationContext);
```
toddbaert: We could probably initialize in parallel with some threads; not a blocker for me but if you initialize 3 providers that needed to setup remote connections, you could be waiting a while before you get success/failure.

providers/multiprovider/README.md
```diff
@@ -0,0 +1,97 @@
+# OpenFeature Multi-Provider for Java
```
toddbaert, line 1: Nice work on the README.

Small nit: 1 sentence per line makes reviewing and change-deltas much nicer.
Markdown will add it's own line-breaks when it renders.

toddbaert (commented):
I left a few comments but overall looks very close.

providers/multiprovider/src/main/java/dev/openfeature/contrib/providers/multiprovider/MultiProvider.java
liran2000: According to [Multi-provider metadata spec](https://github.com/open-feature/spec/blob/main/specification/appendix-a-included-utilities.md#metadata), it is expected to be printed as JSON, which makes sense.  
Both HashMap and Structure not printed as Json.

providers/multiprovider/README.md
liran2000: Thanks, updated.

providers/multiprovider/src/main/java/dev/openfeature/contrib/providers/multiprovider/MultiProvider.java
liran2000: done

providers/multiprovider/src/test/java/dev/openfeature/contrib/providers/multiprovider/MultiProviderTest.java
liran2000: added

toddbaert (approved):

beeme1mr (approved):

aepfli (approved):

PR #1064, opened 2024-11-05
feat!: changing cache provider to caffeine over guava

### LINKED ISSUE #1985: (change) Change cache in openfeature java provider
### Motivation

As mentioned by @liran2000 in this PR comment https://github.com/open-feature/java-sdk-contrib/pull/821#discussion_r1636452408 the cache system used in the java provider is from `guava` and they discouraged using it and recommend using [`caffeine`](https://javadoc.io/doc/com.google.guava/guava/latest/com/google/common/cache/package-summary.html) instead.

Since it is a breaking change we didn't want to introduce it to the provider.

When going for version `1.0.0` of the provider we should consider changing the cache mechanism in favour of `caffeine`.

### Requirements

We should refactor the provider to use [`caffeine`](https://github.com/ben-manes/caffeine) instead of the `guava` cache.

To do so we will have to introduce a breaking change in the `GoFeatureFlagProviderOptions` file, to change the type of the `cacheBuilder` option.

When doing that we should ensure a proper documentation on how to migrate from `guava` to `caffeine`.

matheusverissimo:
This PR changes the implementation of caching from [Guava](https://github.com/google/guava) to [Caffeine](https://github.com/ben-manes/caffeine).

### Related Issues

Resolves [#1985](https://github.com/thomaspoignant/go-feature-flag/issues/1985) from go-feature-flag.

### Notes

This change introduces a breaking change on the API, because the property `cacheConfig` of `GoFeatureFlagProviderOptions` now expects a Caffeine configuration object instead of a Guava CacheBuilder object.

#### Testing

This implementation caused test failure on `should_resolve_from_cache_max_size` in `GoFeatureFlagProviderTest`. 

Unlike Guava, Caffeine does not guarantee that the cache size never exceeds the specified limit at any time due to important differences in the eviction algorithm compared to Guava.

Because of this, the last assertion from this test case was removed.

More about these differences can be found here:
[Caffeine Wiki - Guava](https://github.com/ben-manes/caffeine/wiki/Guava)
https://github.com/ben-manes/caffeine/issues/420#issuecomment-630377203

### Follow-up Tasks

We need to document the breaking changes on the API in:
1. Go Feature Flag Website
2. READ.ME

@thomaspoignant


providers/go-feature-flag/pom.xml
```diff
@@ -62,6 +62,12 @@
             <version>33.3.1-jre</version>
         </dependency>
```
thomaspoignant: **suggestion:** Since `guava` is not used anymore we should probably remove the dependency.

providers/go-feature-flag/src/main/java/dev/openfeature/contrib/providers/gofeatureflag/GoFeatureFlagProviderOptions.java
```diff
@@ -49,14 +50,26 @@ public class GoFeatureFlagProviderOptions {
...
+     * <code>GoFeatureFlagProviderOptions.builder()
+     *   .caffeineConfig(
+     *      Caffeine.newBuilder()
+     *          .initialCapacity(100)
+     *          .maximumSize(100000)
+     *          .expireAfterWrite(Duration.ofMillis(5L * 60L * 1000L))
+     *          .build()
+     *    )
+     *   .build();
+     * </code>
```
thomaspoignant, line 65: **praise:** This is great to have this example in the javadoc 🙌

providers/go-feature-flag/src/main/java/dev/openfeature/contrib/providers/gofeatureflag/controller/CacheController.java
```diff
@@ -23,12 +24,11 @@ public class CacheController {
 
     @Builder
     public CacheController(GoFeatureFlagProviderOptions options) {
-        this.cache = options.getCacheBuilder() != null ? options.getCacheBuilder().build() : buildDefaultCache();
+        this.cache = options.getCacheConfig() != null ? options.getCacheConfig().build() : buildDefaultCache();
     }
 
     private Cache<String, ProviderEvaluation<?>> buildDefaultCache() {
-        return CacheBuilder.newBuilder()
-                .concurrencyLevel(DEFAULT_CACHE_CONCURRENCY_LEVEL)
```
thomaspoignant, line 31: **suggestion:** Since we are not using the constant `DEFAULT_CACHE_CONCURRENCY_LEVEL` it would be great to delete it from the `CacheController`.

thomaspoignant (changes requested):
@matheusverissimo thanks a lot for your pull request this is great and I think we are almost ready to merge it.

I have made some suggestion to keep the code clean and avoid unused dependencies.

Also can you add the breaking change in the README?

providers/go-feature-flag/pom.xml
matheusverissimo: Actually the reason I did not remove the dependency is because `HttpHeaders` from the package `com.google.common.net` is being used on `GoFeatureFlagController`, which comes from Guava.

providers/go-feature-flag/src/main/java/dev/openfeature/contrib/providers/gofeatureflag/controller/CacheController.java
matheusverissimo: Done!

providers/go-feature-flag/pom.xml
thomaspoignant: > Actually the reason I did not remove the dependency is because `HttpHeaders` from the package `com.google.common.net` is being used on `GoFeatureFlagController`, which comes from Guava.

In that case we should probably set as dependency `com.google.common.net` instead of guava or just replace the constants by our own constants.

matheusverissimo: Ok, I'm working on that

liran2000 (approved):

PR #1065, opened 2024-11-06
feat!: changing cache provider to caffeine over guava

### LINKED ISSUE #1985: (change) Change cache in openfeature java provider

matheusverissimo:
This PR changes the implementation of caching from [Guava](https://github.com/google/guava) to [Caffeine](https://github.com/ben-manes/caffeine).

## Related Issues
Resolves https://github.com/thomaspoignant/go-feature-flag/issues/1985 from go-feature-flag.

## Notes
This change introduces a breaking change on the API, because the property cacheConfig of GoFeatureFlagProviderOptions now expects a Caffeine configuration object instead of a Guava CacheBuilder object.

## Testing
This implementation caused test failure on should_resolve_from_cache_max_size in GoFeatureFlagProviderTest.

Unlike Guava, Caffeine does not guarantee that the cache size never exceeds the specified limit at any time due to important differences in the eviction algorithm compared to Guava.

Because of this, the last assertion from this test case was removed.

More about these differences can be found here:
[Caffeine Wiki - Guava](https://github.com/ben-manes/caffeine/wiki/Guava)
https://github.com/ben-manes/caffeine/issues/420#issuecomment-630377203

## Follow-up Tasks
We need to document the breaking changes on the API in:
1. Go Feature Flag Website

@thomaspoignant I opened a new PR because I messed up with my GPG signature and had to rewrite a lot of commits


thomaspoignant (approved):
@matheusverissimo thanks a lot for this contribution.
I will merge it and it will be part of the next release of the provider.

PR #1092, opened 2024-12-02
feat(flagd): Improve e2e coverage

aepfli:
Adding the newest gherkin files to the release, one for config, and one to ensure caching is working properly


providers/flagd/pom.xml
```diff
@@ -335,6 +335,38 @@
...
                             </execution>
+                            <execution>
+                                <id>copy-gherkin-flagd-rpc-caching.feature</id>
+                                <phase>validate</phase>
+                                <goals>
+                                    <goal>exec</goal>
+                                </goals>
+                                <configuration>
+                                    <!-- copy the feature spec we want to test into resources so them can be easily loaded -->
+                                    <!-- run: cp test-harness/features/flagd.feature src/test/resources/features/ -->
```
toddbaert: Please update these comments so they are accurate, or just remove them.

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/RunFlagdRpcCucumberTest.java
```diff
@@ -16,9 +16,10 @@
 @Order(value = Integer.MAX_VALUE)
 @Suite
 @IncludeEngines("cucumber")
-@SelectClasspathResource("features/evaluation.feature")
-@SelectClasspathResource("features/flagd-json-evaluator.feature")
-@SelectClasspathResource("features/flagd.feature")
+//@SelectClasspathResource("features/evaluation.feature")
+//@SelectClasspathResource("features/flagd-json-evaluator.feature")
+//@SelectClasspathResource("features/flagd.feature")
```
toddbaert: Are these meant to be commented? We're losing a bunch of coverage here right? Maybe you just didn't mean to commit this and just did it to speed up your local testing.

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/steps/config/ConfigSteps.java
```diff
@@ -0,0 +1,103 @@
...
+import java.util.Objects;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class ConfigSteps {
+
+    FlagdOptions.FlagdOptionsBuilder builder = FlagdOptions.builder();
+    FlagdOptions options;
+
+    @When("we initialize a config")
```
toddbaert, line 23: This is a bit of a nit, but IMO we should use passive voice for these ... so instead of:

`we initialize a config`

It should be:

`a config is initialized`

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/steps/config/ConfigSteps.java
```diff
@@ -0,0 +1,103 @@
...
+                .findFirst()
+                .orElseThrow(RuntimeException::new);
+        method.invoke(builder, converted);
+    }
+
+
+    Map<String, String> envVarsSet = new HashMap<>();
+
+    @When("we have an environment variable {string} with value {string}")
+    public void we_have_an_environment_variable_with_value(String string, String string2) throws IllegalAccessException, NoSuchFieldException {
```
toddbaert:
```suggestion
    public void we_have_an_environment_variable_with_value(String string, String string2) throws IllegalAccessException, NoSuchFieldException {
```

I'd prefer if we update the param names here to be more descriptive instead of the auto-generated `string1/2`.

toddbaert (commented):
I see this as pretty useful, though the mileage may very depending on whether or not the implementation exposes a config object that can be independently tested.

I would also prefer to somehow mark this gherkin feature file as a "unit" style test somehow (perhaps make the extension `.unit.feature`?)

I also would prefer the passive voice is used: https://github.com/open-feature/java-sdk-contrib/pull/1092/files#r1870241239

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/steps/config/ConfigSteps.java
aepfli: I will adapt the gherkin file, to be more precise

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/RunFlagdRpcCucumberTest.java
aepfli: Yes you are totally right - for local debugging I commented them out.

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/steps/config/ConfigSteps.java
aepfli: ... but as this is also already in a python pr, I would love to tackle our whole Gherkin files as one  - I think we can improve them and how we're using them. Currently it is quiet hard in all our SDKs repos, with no separations between steps etc

providers/flagd/pom.xml
aepfli: do we need those comments? should we myabe remove them?

toddbaert: That's OK with me.

toddbaert (approved):
Good on my end. Feel free to merge if you want @aepfli

providers/flagd/pom.xml
aepfli:
```suggestion
```

PR #1093, opened 2024-12-03
feat: added interception of parameterized tests to Junit OpenFeature Extension

### LINKED ISSUE #1091: [Junit5 extension] Add Interception of @ParameterizedTest to JUnit Open Feature Extension
If the JUnit Open Feature Extension is used, a developer needs to use the `@Flag` annotation on `@Test `and `@ParameterizedTest` methods.

By intercepting the test method, the `OpenFeatureExtension` sets the current namespace for the `TestProvider` which is used later on in the evaluation. This is done by implementing the interface `InvocationInterceptor` and overriding the method `interceptTestMethod`.
Actually, the method `interceptTestMethod` does not intercept methods annotated with `@ParameterizedTest`. This leads to the issue that during the feature flag evaluation, the current namespace is not set int the `TestProvider` and therefore, the flag values cannot be stubbed with the annotation `@Flag`.
To fix the issue not only the method `interceptTestMethod` must be overridden, but `InvocationInterceptor.interceptTestTemplateMethod()` must be overridden as well.

 

### A/C

@Flag annotation mocks feature flags on test methods annotated with @Test and @ParameterizedTest

sebastian-zahrhuber:
## This PR
If the JUnit Open Feature Extension is used, a developer needs to use the `@Flag` annotation on test methods annotated with  `@Test` and `@ParameterizedTest`. This PR adds support for parameterized tests.

### Related Issues
Fixes #1091

### DIFF
--- a/tools/junit-openfeature/src/main/java/dev/openfeature/contrib/tools/junitopenfeature/OpenFeatureExtension.java
+++ b/tools/junit-openfeature/src/main/java/dev/openfeature/contrib/tools/junitopenfeature/OpenFeatureExtension.java
@@ -3,7 +3,11 @@ package dev.openfeature.contrib.tools.junitopenfeature;
 import dev.openfeature.sdk.OpenFeatureAPI;
 import dev.openfeature.sdk.providers.memory.Flag;
 import org.apache.commons.lang3.BooleanUtils;
-import org.junit.jupiter.api.extension.*;
+import org.junit.jupiter.api.extension.AfterEachCallback;
+import org.junit.jupiter.api.extension.BeforeEachCallback;
+import org.junit.jupiter.api.extension.ExtensionContext;
+import org.junit.jupiter.api.extension.InvocationInterceptor;
+import org.junit.jupiter.api.extension.ReflectiveInvocationContext;
 import org.junitpioneer.internal.PioneerAnnotationUtils;
 
 import java.lang.reflect.Method;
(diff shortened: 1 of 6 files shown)


tools/junit-openfeature/src/main/java/dev/openfeature/contrib/tools/junitopenfeature/OpenFeatureExtension.java
```diff
@@ -3,11 +3,7 @@
 import dev.openfeature.sdk.OpenFeatureAPI;
 import dev.openfeature.sdk.providers.memory.Flag;
 import org.apache.commons.lang3.BooleanUtils;
-import org.junit.jupiter.api.extension.AfterEachCallback;
-import org.junit.jupiter.api.extension.BeforeEachCallback;
-import org.junit.jupiter.api.extension.ExtensionContext;
-import org.junit.jupiter.api.extension.InvocationInterceptor;
-import org.junit.jupiter.api.extension.ReflectiveInvocationContext;
+import org.junit.jupiter.api.extension.*;
```
aepfli: We need to improve our project's defaults here, which is more of a configuration fault on our side. But can you please revert the star import?

aepfli (approved):
Thank you! This looks amazing. Just to make it more clear to the users, a small example in the readme would be highly appreciated.

PR #1111, opened 2024-12-18
feat(flagd): ssl e2e tests

aepfli:
interestingly, it seems like ssl is not working with this approach for inprocess.

- opened bug for flagd - https://github.com/open-feature/flagd/issues/1479

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,6 +4,7 @@
 [submodule "providers/flagd/test-harness"]
 	path = providers/flagd/test-harness
 	url = https://github.com/open-feature/test-harness.git
+	branch = v0.5.19
 [submodule "providers/flagd/spec"]
 	path = providers/flagd/spec
 	url = https://github.com/open-feature/spec.git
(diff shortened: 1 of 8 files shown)


.gitmodules
```diff
@@ -4,6 +4,7 @@
 [submodule "providers/flagd/test-harness"]
 	path = providers/flagd/test-harness
 	url = https://github.com/open-feature/test-harness.git
+	branch = v0.5.19
```
aepfli, line 7: this change should allow renovate to create prs to update it

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/RunFlagdInProcessSSLCucumberTest.java
```diff
@@ -0,0 +1,27 @@
...
+import static io.cucumber.junit.platform.engine.Constants.GLUE_PROPERTY_NAME;
+import static io.cucumber.junit.platform.engine.Constants.PLUGIN_PROPERTY_NAME;
+
+/**
+ * Class for running the reconnection tests for the RPC provider
+ */
+@Order(value = Integer.MAX_VALUE)
+@Suite(failIfNoTests = false)
+@IncludeEngines("cucumber")
+//@SelectClasspathResource("features/evaluation.feature")
```
aepfli, line 20: i do not get this working, maybe i am wrong here, but if somebody has time to take a look this would be highly appreciated - or we do have a bug

aepfli: seems like, flagd currently does not support SSL for the sync server

toddbaert (approved):

PR #1112, opened 2024-12-18
feat!: Use grpc intern reconnections for  rpc event stream

warber:
## This PR

* Updates the reconnection behavior of the flagd provider in RPC mode.
* **Breaking Change**: If the provider loses connection, it will emit a STALE event. If the connection is successfully reestablished within the `FLAGD_RETRY_GRACE_PERIOD`, the provider will emit a READY event and resume full operation. If reconnection is not achieved within this time frame, an ERROR event will be emitted. The default value for `FLAGD_RETRY_GRACE_PERIOD` is set to `5`.

This PR marks the initial step in phasing out our custom gRPC backoff and reconnect logic. Instead, we are transitioning to fully rely on the default behaviour provided by the gRPC library.
This change is currently limited to the RPC mode of the provider (specifically the `dev.openfeature.contrib.providers.flagd.resolver.grpc` package). In this first phase, the `GrpcConnector` within this package has been refactored to operate independently of custom backoff logic, leveraging only gRPC internals to maintain a reliable connection.

In the next phase/PR, the refactored GrpcConnector should be reused within the`dev.openfeature.contrib.providers.flagd.resolver.process` package, enabling the removal of custom reconnect logic from that implementation as well.

### DIFF
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/Config.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/Config.java
@@ -15,6 +15,7 @@ public final class Config {
 
     static final int DEFAULT_DEADLINE = 500;
     static final int DEFAULT_STREAM_DEADLINE_MS = 10 * 60 * 1000;
+    static final int DEFAULT_STREAM_RETRY_GRACE_PERIOD = 5;
     static final int DEFAULT_MAX_CACHE_SIZE = 1000;
     static final long DEFAULT_KEEP_ALIVE = 0;
 
@@ -35,6 +36,7 @@ public final class Config {
     static final String KEEP_ALIVE_MS_ENV_VAR_NAME_OLD = "FLAGD_KEEP_ALIVE_TIME";
     static final String KEEP_ALIVE_MS_ENV_VAR_NAME = "FLAGD_KEEP_ALIVE_TIME_MS";
     static final String TARGET_URI_ENV_VAR_NAME = "FLAGD_TARGET_URI";
+    static final String STREAM_RETRY_GRACE_PERIOD = "FLAGD_RETRY_GRACE_PERIOD";
 
     static final String RESOLVER_RPC = "rpc";
     static final String RESOLVER_IN_PROCESS = "in-process";
@@ -52,7 +54,6 @@ public final class Config {
     public static final String LRU_CACHE = CacheType.LRU.getValue();
     static final String DEFAULT_CACHE = LRU_CACHE;
 
-    static final int DEFAULT_MAX_EVENT_STREAM_RETRIES = 5;
     static final int BASE_EVENT_STREAM_RETRY_BACKOFF_MS = 1000;
 
     static String fallBackToEnvOrDefault(String key, String defaultValue) {
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdOptions.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdOptions.java
@@ -13,82 +13,111 @@ import java.util.function.Function;
 import lombok.Builder;
 import lombok.Getter;
 
-/** FlagdOptions is a builder to build flagd provider options. */
+/**
+ * FlagdOptions is a builder to build flagd provider options.
+ */
 @Builder
 @Getter
 @SuppressWarnings("PMD.TooManyStaticImports")
 public class FlagdOptions {
 
-    /** flagd resolving type. */
+    /**
+     * flagd resolving type.
+     */
     private Config.EvaluatorType resolverType;
 
-    /** flagd connection host. */
+    /**
+     * flagd connection host.
+     */
     @Builder.Default
     private String host = fallBackToEnvOrDefault(Config.HOST_ENV_VAR_NAME, Config.DEFAULT_HOST);
 
-    /** flagd connection port. */
+    /**
+     * flagd connection port.
+     */
     private int port;
 
-    /** Use TLS connectivity. */
+    /**
+     * Use TLS connectivity.
+     */
     @Builder.Default
     private boolean tls = Boolean.parseBoolean(fallBackToEnvOrDefault(Config.TLS_ENV_VAR_NAME, Config.DEFAULT_TLS));
 
-    /** TLS certificate overriding if TLS connectivity is used. */
+    /**
+     * TLS certificate overriding if TLS connectivity is used.
+     */
     @Builder.Default
     private String certPath = fallBackToEnvOrDefault(Config.SERVER_CERT_PATH_ENV_VAR_NAME, null);
 
-    /** Unix socket path to flagd. */
+    /**
+     * Unix socket path to flagd.
+     */
     @Builder.Default
     private String socketPath = fallBackToEnvOrDefault(Config.SOCKET_PATH_ENV_VAR_NAME, null);
 
-    /** Cache type to use. Supports - lru, disabled. */
+    /**
+     * Cache type to use. Supports - lru, disabled.
+     */
     @Builder.Default
     private String cacheType = fallBackToEnvOrDefault(Config.CACHE_ENV_VAR_NAME, Config.DEFAULT_CACHE);
 
-    /** Max cache size. */
+    /**
+     * Max cache size.
+     */
     @Builder.Default
     private int maxCacheSize =
             fallBackToEnvOrDefault(Config.MAX_CACHE_SIZE_ENV_VAR_NAME, Config.DEFAULT_MAX_CACHE_SIZE);
 
-    /** Max event stream connection retries. */
-    @Builder.Default
-    private int maxEventStreamRetries = fallBackToEnvOrDefault(
-            Config.MAX_EVENT_STREAM_RETRIES_ENV_VAR_NAME, Config.DEFAULT_MAX_EVENT_STREAM_RETRIES);
-
-    /** Backoff interval in milliseconds. */
+    /**
+     * Backoff interval in milliseconds.
+     */
     @Builder.Default
     private int retryBackoffMs = fallBackToEnvOrDefault(
             Config.BASE_EVENT_STREAM_RETRY_BACKOFF_MS_ENV_VAR_NAME, Config.BASE_EVENT_STREAM_RETRY_BACKOFF_MS);
 
     /**
-     * Connection deadline in milliseconds. For RPC resolving, this is the deadline to connect to
-     * flagd for flag evaluation. For in-process resolving, this is the deadline for sync stream
-     * termination.
+     * Connection deadline in milliseconds.
+     * For RPC resolving, this is the deadline to connect to flagd for flag
+     * evaluation.
+     * For in-process resolving, this is the deadline for sync stream termination.
      */
     @Builder.Default
     private int deadline = fallBackToEnvOrDefault(Config.DEADLINE_MS_ENV_VAR_NAME, Config.DEFAULT_DEADLINE);
 
     /**
-     * Streaming connection deadline in milliseconds. Set to 0 to disable the deadline. Defaults to
-     * 600000 (10 minutes); recommended to prevent infrastructure from killing idle connections.
+     * Streaming connection deadline in milliseconds.
+     * Set to 0 to disable the deadline.
+     * Defaults to 600000 (10 minutes); recommended to prevent infrastructure from killing idle connections.
      */
     @Builder.Default
     private int streamDeadlineMs =
             fallBackToEnvOrDefault(Config.STREAM_DEADLINE_MS_ENV_VAR_NAME, Config.DEFAULT_STREAM_DEADLINE_MS);
 
-    /** Selector to be used with flag sync gRPC contract. */
+    /**
+     * Grace time period in seconds before provider moves from STALE to ERROR.
+     * Defaults to 5
+     */
+    @Builder.Default
+    private int streamRetryGracePeriod =
+            fallBackToEnvOrDefault(Config.STREAM_RETRY_GRACE_PERIOD, Config.DEFAULT_STREAM_RETRY_GRACE_PERIOD);
+    /**
+     * Selector to be used with flag sync gRPC contract.
+     **/
     @Builder.Default
     private String selector = fallBackToEnvOrDefault(Config.SOURCE_SELECTOR_ENV_VAR_NAME, null);
 
-    /** gRPC client KeepAlive in milliseconds. Disabled with 0. Defaults to 0 (disabled). */
+    /**
+     * gRPC client KeepAlive in milliseconds. Disabled with 0.
+     * Defaults to 0 (disabled).
+     **/
     @Builder.Default
     private long keepAlive = fallBackToEnvOrDefault(
             Config.KEEP_ALIVE_MS_ENV_VAR_NAME,
             fallBackToEnvOrDefault(Config.KEEP_ALIVE_MS_ENV_VAR_NAME_OLD, Config.DEFAULT_KEEP_ALIVE));
 
     /**
-     * File source of flags to be used by offline mode. Setting this enables the offline mode of the
-     * in-process provider.
+     * File source of flags to be used by offline mode.
+     * Setting this enables the offline mode of the in-process provider.
      */
     @Builder.Default
     private String offlineFlagSourcePath = fallBackToEnvOrDefault(Config.OFFLINE_SOURCE_PATH, null);
@@ -96,31 +125,35 @@ public class FlagdOptions {
     /**
      * gRPC custom target string.
      *
-     * <p>Setting this will allow user to use custom gRPC name resolver at present we are supporting
-     * all core resolver along with a custom resolver for envoy proxy resolution. For more visit
-     * (https://grpc.io/docs/guides/custom-name-resolution/)
+     * <p>Setting this will allow user to use custom gRPC name resolver at present
+     * we are supporting all core resolver along with a custom resolver for envoy proxy
+     * resolution. For more visit (https://grpc.io/docs/guides/custom-name-resolution/)
      */
     @Builder.Default
     private String targetUri = fallBackToEnvOrDefault(Config.TARGET_URI_ENV_VAR_NAME, null);
 
     /**
-     * Function providing an EvaluationContext to mix into every evaluations. The sync-metadata
-     * response
+     * Function providing an EvaluationContext to mix into every evaluations.
+     * The sync-metadata response
      * (https://buf.build/open-feature/flagd/docs/main:flagd.sync.v1#flagd.sync.v1.GetMetadataResponse),
-     * represented as a {@link dev.openfeature.sdk.Structure}, is passed as an argument. This function
-     * runs every time the provider (re)connects, and its result is cached and used in every
-     * evaluation. By default, the entire sync response (converted to a Structure) is used.
+     * represented as a {@link dev.openfeature.sdk.Structure}, is passed as an
+     * argument.
+     * This function runs every time the provider (re)connects, and its result is cached and used in every evaluation.
+     * By default, the entire sync response (converted to a Structure) is used.
      */
     @Builder.Default
     private Function<Structure, EvaluationContext> contextEnricher =
             (syncMetadata) -> new ImmutableContext(syncMetadata.asMap());
 
-    /** Inject a Custom Connector for fetching flags. */
+    /**
+     * Inject a Custom Connector for fetching flags.
+     */
     private Connector customConnector;
 
     /**
-     * Inject OpenTelemetry for the library runtime. Providing sdk will initiate distributed tracing
-     * for flagd grpc connectivity.
+     * Inject OpenTelemetry for the library runtime. Providing sdk will initiate
+     * distributed tracing for flagd grpc
+     * connectivity.
      */
     private OpenTelemetry openTelemetry;
 
@@ -139,11 +172,14 @@ public class FlagdOptions {
         };
     }
 
-    /** Overload default lombok builder. */
+    /**
+     * Overload default lombok builder.
+     */
     public static class FlagdOptionsBuilder {
         /**
-         * Enable OpenTelemetry instance extraction from GlobalOpenTelemetry. Note that, this is only
-         * useful if global configurations are registered.
+         * Enable OpenTelemetry instance extraction from GlobalOpenTelemetry. Note that,
+         * this is only useful if global
+         * configurations are registered.
          */
         public FlagdOptionsBuilder withGlobalTelemetry(final boolean b) {
             if (b) {
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdOptions.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdOptions.java
@@ -98,7 +98,7 @@ public class FlagdOptions {
      * Defaults to 5
      */
     @Builder.Default
-    private int streamRetryGracePeriod =
+    private int retryGracePeriod =
             fallBackToEnvOrDefault(Config.STREAM_RETRY_GRACE_PERIOD, Config.DEFAULT_STREAM_RETRY_GRACE_PERIOD);
     /**
      * Selector to be used with flag sync gRPC contract.
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
@@ -21,7 +21,9 @@ import java.util.List;
 import java.util.function.Function;
 import lombok.extern.slf4j.Slf4j;
 
-/** OpenFeature provider for flagd. */
+/**
+ * OpenFeature provider for flagd.
+ */
 @Slf4j
 @SuppressWarnings({"PMD.TooManyStaticImports", "checkstyle:NoFinalizer"})
 public class FlagdProvider extends EventProvider {
@@ -38,7 +40,9 @@ public class FlagdProvider extends EventProvider {
         // DO NOT REMOVE, spotbugs: CT_CONSTRUCTOR_THROW
     }
 
-    /** Create a new FlagdProvider instance with default options. */
+    /**
+     * Create a new FlagdProvider instance with default options.
+     */
     public FlagdProvider() {
         this(FlagdOptions.builder().build());
     }
@@ -55,10 +59,7 @@ public class FlagdProvider extends EventProvider {
                 break;
             case Config.RESOLVER_RPC:
                 this.flagResolver = new GrpcResolver(
-                        options,
-                        new Cache(options.getCacheType(), options.getMaxCacheSize()),
-                        this::isConnected,
-                        this::onConnectionEvent);
+                        options, new Cache(options.getCacheType(), options.getMaxCacheSize()), this::onConnectionEvent);
                 break;
             default:
                 throw new IllegalStateException(
@@ -80,7 +81,7 @@ public class FlagdProvider extends EventProvider {
         }
 
         this.flagResolver.init();
-        this.initialized = true;
+        this.initialized = this.connected = true;
     }
 
     @Override
@@ -129,8 +130,10 @@ public class FlagdProvider extends EventProvider {
     }
 
     /**
-     * An unmodifiable view of a Structure representing the latest result of the SyncMetadata. Set on
-     * initial connection and updated with every reconnection. see:
+     * An unmodifiable view of a Structure representing the latest result of the
+     * SyncMetadata.
+     * Set on initial connection and updated with every reconnection.
+     * see:
      * https://buf.build/open-feature/flagd/docs/main:flagd.sync.v1#flagd.sync.v1.FlagSyncService.GetMetadata
      *
      * @return Object map representing sync metadata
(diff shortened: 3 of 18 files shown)


providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/Config.java
```diff
@@ -55,7 +57,7 @@ public final class Config {
     public static final String LRU_CACHE = CacheType.LRU.getValue();
     static final String DEFAULT_CACHE = LRU_CACHE;
 
-    static final int DEFAULT_MAX_EVENT_STREAM_RETRIES = 5;
+    static final int DEFAULT_MAX_EVENT_STREAM_RETRIES = 7;
```
aepfli: As this should not be used anymore, i think we can remove it
```suggestion
```

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/common/ConnectionState.java
```diff
@@ -0,0 +1,58 @@
...
+/**
+ * Represents the state of a connection, indicating whether it is connected,
+ * disconnected, or stale.
+ * <p>
+ * This class is immutable and uses the {@link lombok.AllArgsConstructor} annotation
+ * to generate a constructor with parameters for all fields.
+ * It also uses {@link lombok.Getter} to provide getter methods for the fields.
+ */
+@AllArgsConstructor
+public class ConnectionState {
```
aepfli: Currently, this class can handle two dedicated state properties. However, they could also be represented via enums, which may be easier to read/maintain. We currently have four states (Disconnected, Connected, Stale, and Error). Would it be worth changing this to enums? We could skip handling different state variables and just utilize the enums.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/common/Util.java
```diff
@@ -1,36 +1,123 @@
...
+    }
+
+    /**
+     * Monitors the state of a gRPC {@link ManagedChannel} and triggers callbacks for specific state changes.
+     *
+     * @param ch      the gRPC {@link ManagedChannel} to monitor
+     * @param onReady a {@link Runnable} to execute when the channel becomes READY
+     * @param onLost  a {@link Runnable} to execute when the channel enters a TRANSIENT_FAILURE state
+     */
+    public static void monitorChannelState(ManagedChannel ch, Runnable onReady, Runnable onLost) {
```
aepfli: `onLost` can be interpreted as a lot of things, maybe we should rename it to `onConnectionLost` or `onFailure` to signal the connection with the ConnectivityState.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/grpc/EventStreamObserver.java
```diff
@@ -107,16 +112,13 @@ private void handleConfigurationChangeEvent(EventStreamResponse value) {
         this.onConnectionEvent.accept(true, changedFlags);
     }
 
+    /**
+     * Handles provider readiness events by clearing the cache (if enabled) and notifying listeners of readiness.
+     */
     private void handleProviderReadyEvent() {
         this.onConnectionEvent.accept(true, Collections.emptyList());
-        if (this.cache.getEnabled()) {
+        if (this.cache.getEnabled().equals(Boolean.TRUE)) {
```
aepfli, line 118: [question] Do we always want to purge the cache on a reconnect, that means even from Stale to ready? - (sidenote: need to check what is in the gherkin file for this)

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/grpc/GrpcConnector.java
```diff
@@ -1,174 +1,232 @@
...
+    }
 
-        log.error("failed to connect to event stream, exhausted retries");
-        this.onConnectionEvent(false, Collections.emptyList());
+    /**
+     * Handles the event when the GRPC channel loses its connection, marking the connection as lost.
+     * Schedules a reconnection task after a grace period and emits a stale connection event.
+     */
+    private synchronized void onConnectionLost() {
+        log.warn("Connection lost. Emit STALE event...");
```
aepfli: [suggestion] Stale is not something actionable by the sdk user, i do think a info log, or even debug is enough. We should only expect an action by our users, if there is an error

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/grpc/GrpcResolver.java
```diff
@@ -86,53 +94,56 @@ public void shutdown() throws Exception {
      * Boolean evaluation from grpc resolver.
      */
     public ProviderEvaluation<Boolean> booleanEvaluation(String key, Boolean defaultValue,
-            EvaluationContext ctx) {
+                                                         EvaluationContext ctx) {
         ResolveBooleanRequest request = ResolveBooleanRequest.newBuilder().buildPartial();
 
-        return resolve(key, ctx, request, this.connector.getResolver()::resolveBoolean, null);
+
+        return resolve(key, ctx, request, ((ServiceGrpc.ServiceBlockingStub) connector.getResolver())::resolveBoolean,
```
aepfli: [question] Can we prevent the casting here? 

We could eliminate the casts by providing the generic types to the field declaration.

aepfli (commented):
Thank you, this looks really good, and sets the foundation for a lot of simplifications. 

I added some comments, more questions out of curiosity. 

Thank you

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/common/ConnectionState.java
warber: absolutely, thx.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/grpc/GrpcConnector.java
warber: I'd argue that loosing the connection (even temporarily) has at least the user facing side effect that evaluations fall back to their default value. Maybe the user want to have a hint why this happened? I interpreted it as a precurser to the error case which is permanently loosing connection. wdyt?

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/grpc/EventStreamObserver.java
warber: i don't know, this change was only suggested by sonar.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/grpc/GrpcResolver.java
warber: i don't know. Note, that currently the the only reason why the blocking stub stuff is part of the `GrpcConnector` is to make the tests in `FlagdProviderTest` happy. They rely on a mocked stub given back by the `GrpcConnector`. I have an idea how to make the tests green without that. Then we can delete this code as well and let the caller create the stubs. I could do that as a follow up.

providers/flagd/pom.xml
```diff
@@ -150,7 +150,12 @@
             <version>1.20.4</version>
             <scope>test</scope>
         </dependency>
-
+        <dependency>
```
warber: I've added this package to be able to use an "in-memory" grpc server during tests...- quite nice

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/Config.java
```diff
@@ -18,6 +18,7 @@ public final class Config {
 
     static final int DEFAULT_DEADLINE = 500;
     static final int DEFAULT_STREAM_DEADLINE_MS = 10 * 60 * 1000;
+    static final int DEFAULT_STREAM_RETRY_GRACE_PERIOD = 50_000;
```
warber: i don't know if this is a good default, pls check

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
```diff
@@ -159,33 +159,41 @@ private boolean isConnected() {
     }
 
     private void onConnectionEvent(ConnectionEvent connectionEvent) {
-        boolean previous = connected;
-        boolean current = connected = connectionEvent.isConnected();
+        final boolean wasConnected = connected;
```
aepfli, line 159: [suggestion] We are storing the application's state in a' connected' boolean and mapping our ConnectionState to it. Would it be worth changing the type of connection and working with the state directly? This whole method looks like a state machine, and I think handling this with a state-like pattern could be beneficial, like 

```java
 private ConnectionState connected;

 private void onConnectionEvent(ConnectionEvent connectionEvent) {

        syncMetadata = connectionEvent.getSyncMetadata();
        enrichedContext = contextEnricher.apply(connectionEvent.getSyncMetadata());

        if (!isInitialized) {
            return;
        }

        switch (connectionEvent.getConnectionState()) {
            case CONNECTED:
                ProviderEventDetails.ProviderEventDetailsBuilder<?, ?>  builder = ProviderEventDetails.builder()
                        .flagsChanged(connectionEvent.getFlagsChanged());
                if (connected != ConnectionState.CONNECTED) {
                    this.emitProviderReady(builder.message("connected to flagd").build());
                } else if (connected == ConnectionState.CONNECTED) {
                    this.emitProviderConfigurationChanged( builder.message("configuration changed").build());
                }
                break;
            case STALE:
                this.emitProviderStale(ProviderEventDetails.builder().message("there has been an error").build());
                break;
            case ERROR:
                this.emitProviderError(ProviderEventDetails.builder().message("there has been an error").build());
                break;
        }
        this.connected = connectionEvent.getConnectionState();

    }
```
just an idea/suggestion - no need to act

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/grpc/GrpcConnector.java
aepfli: I agree that the information is interesting, but more often, this is just information (info) that it happened, and there is no action needed. eg. I know certain companies where error logs will trigger internal processes, and for an expected reconnect within the grace period, a warning would be a high priority.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/grpc/GrpcResolver.java
aepfli:
```java
private final GrpcConnector<ServiceGrpc.ServiceStub, ServiceGrpc.ServiceBlockingStub> connector;
```
removes the need to cast ;)

aepfli (approved):

aepfli:
important, before merging https://github.com/open-feature/flagd/pull/1478 should be merged to fix on the property and env var naming

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/grpc/GrpcConnector.java
warber: convinced. thx.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/Config.java
toddbaert: I think 50s is an odd choice, mostly just because it's not a fraction or multiple of a minute, but that's a nit.

I wonder if there's any reason is should or should not be the same as the `DEFAULT_STREAM_DEADLINE_MS` ... :thinking: 

I think I'd prefer something more like 30s but I don't have a great reason for that.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
toddbaert: Agreed this seems "nice" but not required for me.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
```diff
@@ -159,33 +159,41 @@ private boolean isConnected() {
...
+        final boolean isConnected = connected = connectionEvent.isConnected();
```
toddbaert, line 160: :sunglasses:

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/common/Util.java
```diff
@@ -1,36 +1,124 @@
...
+     *
+     * @param ch           the gRPC {@link ManagedChannel} to monitor
+     * @param desiredState the desired {@link ConnectivityState} to wait for
+     * @param callback     a {@link Runnable} to execute when the desired state is reached
+     * @param timeout      the maximum time to wait
+     * @param unit         the {@link TimeUnit} for the timeout parameter
+     * @throws InterruptedException if the waiting thread is interrupted
+     * @throws GeneralError         if the deadline is exceeded before reaching the desired state
+     */
+    public static void waitForDesiredState(ManagedChannel ch, ConnectivityState desiredState, Runnable callback,
```
toddbaert:
```suggestion
    public static void waitForDesiredState(ManagedChannel channel, ConnectivityState desiredState, Runnable callback,
```

nit: var names like `ch` read more like Go than Java to me :sweat_smile:

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/grpc/EventStreamObserver.java
```diff
@@ -52,37 +57,38 @@ public void onNext(EventStreamResponse value) {
...
     public void onError(Throwable throwable) {
-        if (Boolean.TRUE.equals(shouldRetrySilently.get())) {
-            log.debug("Event stream error, trying to recover", throwable);
-        } else {
-            log.error("Event stream error", throwable);
-            if (this.cache.getEnabled()) {
-                this.cache.clear();
-            }
-            this.onConnectionEvent.accept(false, Collections.emptyList());
+        if (this.cache.getEnabled().equals(Boolean.TRUE)) {
```
toddbaert, line 70: Another nit, but generally I prefer such expressions in the reverse order in Java, since `Boolean.TRUE` will obviously never ever be `null`, so it's less likely to cause NPEs whereas if `this.cache.getEnabled()` becomes nullable at some point (maybe it's a boxed type) you can see NPEs here calling `.equals`.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/grpc/EventStreamObserver.java
```diff
@@ -52,37 +57,38 @@ public void onNext(EventStreamResponse value) {
...
-        handleEndOfStream();
     }
 
+    /**
+     * Called when the stream is completed.
+     */
     @Override
     public void onCompleted() {
-        if (this.cache.getEnabled()) {
+        if (this.cache.getEnabled().equals(Boolean.TRUE)) {
```
toddbaert, line 80: [Same](https://github.com/open-feature/java-sdk-contrib/pull/1112/files#r1890838708)

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/grpc/EventStreamObserver.java
toddbaert: I think we do need to purge here ; we could have missed change events while we were stale, so now that we are connected we should build up a new cache.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/grpc/GrpcConnector.java
toddbaert: Another thing to note is that it's possible to attach SDK event handlers to any provider state transition, such as STALE. I'm leaning towards making these all debug for that reason. WDYT?

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/FlagdProviderTest.java
```diff
@@ -470,143 +459,145 @@ void reason_mapped_correctly_if_unknown() {
...
-        objectDetails = api.getClient().getObjectDetails(FLAG_KEY_OBJECT, new Value());
-        assertEquals(INNER_STRUCT_VALUE, objectDetails.getValue().asStructure()
-                .asMap().get(INNER_STRUCT_KEY).asString());
-        assertEquals(OBJECT_VARIANT, objectDetails.getVariant());
-        assertEquals(STATIC_REASON, objectDetails.getReason());
-    }
+//    @Test
+//    @Disabled
+//    void invalidate_cache() {
+//        ResolveBooleanResponse booleanResponse = ResolveBooleanResponse.newBuilder()
```
toddbaert: I'm assuming these aren't meant to be commented?

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/FlagdProviderTest.java
```diff
@@ -707,146 +700,149 @@ private void do_resolvers_cache_responses(String reason, Boolean eventStreamAliv
...
-        assertEquals(DOUBLE_VARIANT, floatDetails.getVariant());
-        assertEquals(STATIC_REASON, floatDetails.getReason());
-
-        objectDetails = api.getClient().getObjectDetails(FLAG_KEY_OBJECT, new Value());
-        assertEquals(INNER_STRUCT_VALUE, objectDetails.getValue().asStructure()
-                .asMap().get(INNER_STRUCT_KEY).asString());
-        assertEquals(OBJECT_VARIANT, objectDetails.getVariant());
-        assertEquals(STATIC_REASON, objectDetails.getReason());
-    }
+//    @Test
```
toddbaert: Same as https://github.com/open-feature/java-sdk-contrib/pull/1112/files#r1890845896

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/resolver/grpc/GrpcConnectorTest.java
```diff
@@ -1,496 +1,123 @@
...
+        tearDownGrpcServer();
     }
 
-    /**
-     * OS Specific test - This test is valid only on Linux system as it rely on
-     * epoll availability
-     */
-    @Test
-    @EnabledOnOs(OS.LINUX)
-    void path_arg_should_build_domain_socket_with_correct_path() {
```
toddbaert, line 372: I'm OK with deleting some of these tests if there's no longer any need for them, but I want to be sure that we actually have these covered elsewhere or in e2e. This one in particular is testing unix socket support I'm not sure we have elsewhere.

If there's some reason that this test is super hard to update or maintain, I'm also OK with deleting it here, but maybe we need to have some kind of e2e testing support for unix sockets (as well as SSL). I'd be interesting in helping to support that in our test bed which would probably be much easier than unit testing. :thinking: 

WDYT @aepfli ?

toddbaert (commented):
I'm nearly ready to approve this but I have a few questions/comments, mostly nits/testing.

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/resolver/grpc/GrpcConnectorTest.java
aepfli: I think we're mixing integration and unit tests here —in the end, we should test that we're correctly setting everything within the channels, etc. Therefore, I would only need to test our channel builder, which is more like an integration test. To be fair, I am not sure I would cover this with an integration test. 

As you suggest, this is better suited for an e2e test with the docker image - as sockets are just file mounts, this should be reasonably easy. But I am not sure we need an integration test like this. If we do a proper builder test, we should be fine.

IMHO, our config gherkin test tests if the configuration is set correctly - part one of those tests (with env vars and the other way around)  - part two would be checking if the options are correctly used in the channel builder. The rest here is just bloated, in my opinion. Writing a proper unit test for the channel builder would be sufficient

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/common/Util.java
```diff
@@ -1,36 +1,124 @@
...
+            ConnectivityState state = ch.getState(false);
+            log.debug("Channel state changed to: {}", state);
+            if (state == ConnectivityState.READY) {
+                onConnectionReady.run();
+            }
+            if (state == ConnectivityState.TRANSIENT_FAILURE) {
+                onConnectionLost.run();
+            }
+            // Re-register the state monitor to watch for the next state transition.
+            monitorChannelState(ch, onConnectionReady, onConnectionLost);
```
chrfwow: Apparently the SDK is unreliable when it comes to state changes
>"Transitions" to the same state are possible, because intermediate states may not have been observed. The API is only reliable in tracking the current state.

can we live with that?

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
```diff
@@ -159,33 +159,41 @@ private boolean isConnected() {
     }
 
     private void onConnectionEvent(ConnectionEvent connectionEvent) {
```
chrfwow, line 158: Are we ok with this method not being thread safe?

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/grpc/EventStreamObserver.java
warber: Yeah, i actually just followed the sonar linter suggestion here

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/FlagdProviderTest.java
warber: these tests do not compile anymore. discussed with @aepfli. The functionality should be covered by e2e tests. Given the complexity of this test, its not worth to make them work or maintain them. I will delete this. ok?

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/grpc/GrpcConnector.java
warber: convinced again :)

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/Config.java
warber: hehe yeah, this is a bit subjective. i think 50s is nice because it is half of 100s :D but you could also argue that 30s is nice because it's half of 1m :D 

Maybe lets use the same as @aepfli used in the python provider, or at least we should maybe align these default anyway?

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/common/Util.java
warber: that's interesting. @toddbaert @aepfli

warber: As I understand it there is a small chance that we miss state transitions happening during calling `getState()` and registering the callback via `notifyWhenStateChanged()`. I don't see any way out here. This could only be mitigated by checking the current state right after calling `notifyWhenStateChanged()` again and react accordingly, but this does only minimize the time frame where we might miss a transition.

The other option would be to switch the logic to a "poll-based" approach and forget about using `notifyWhenStageChanged()` at all. A nice side effect is that the code becomes quite trivial, and does not involve any recursive calls for re-register the callbacks. Something like this:

```java
public static void monitorChannelState(ManagedChannel channel, Runnable onConnectionReady,
                                                  Runnable onConnectionLost, long pollIntervalMs) {
    ScheduledExecutorService scheduler = Executors.newSingleThreadScheduledExecutor();

    AtomicReference<ConnectivityState> lastState = new AtomicReference<>(ConnectivityState.IDLE);

    Runnable pollTask = () -> {
        ConnectivityState currentState = channel.getState(true);
        if (currentState != lastState.get()) {
            if (currentState == ConnectivityState.READY) {
                onConnectionReady.run();
            } else if (currentState == ConnectivityState.TRANSIENT_FAILURE) {
                onConnectionLost.run();
            }
            lastState.set(currentState);
        }
    };

    scheduler.scheduleAtFixedRate(pollTask, 0, pollIntervalMs, TimeUnit.MILLISECONDS);
}
```
wdyt?

aepfli: I am not sure that polling will solve our issue here; we may have interpreted the API wrong (see https://github.com/grpc/grpc-java/issues/11615#issuecomment-2414105358). We should provide the state we think we're currently in into the method and not fetch the current one, and use it as the parameter.

toddbaert: > As I understand it there is a small chance that we miss state transitions happening during calling getState() and registering the callback via notifyWhenStateChanged(). I don't see any way out here. 

I don't either.

@chrfwow also pointed out to me that `monitorChannelState` is recursive and, at least as this is written currently, it will cause stack overflows _eventually_. We could use a runnable or something to call the callback instead, but that seems like it could be combersome.

I think the polling approach is better, TBH. It gets my vote for both the above reasons.

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/FlagdProviderTest.java
toddbaert: Full support for a delete since they are covered e2e

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/resolver/grpc/GrpcConnectorTest.java
toddbaert: I'm convinced, I'm happy to delete. cc @warber

toddbaert (commented):
Besides deleting some commented tests, a resolution for [this](https://github.com/open-feature/java-sdk-contrib/pull/1112#discussion_r1891245154) is my only blocker. I've left my opinion on that one, but as long as we have a reliable re-connection and no stack overflow potential, I don't mind our approach.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/Config.java
aepfli: it is now defined within https://flagd.dev/reference/specifications/providers/#configuration-options  - also the property names etc.

aepfli:
I am sorry that I am causing so many conflicts. If it is okay for you, I will resolve them for you

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdOptions.java
```diff
@@ -13,114 +13,147 @@
...
+/**
+ * FlagdOptions is a builder to build flagd provider options.
+ */
 @Builder
 @Getter
 @SuppressWarnings("PMD.TooManyStaticImports")
 public class FlagdOptions {
 
-    /** flagd resolving type. */
+    /**
```
aepfli, line 24: hmm my spotless removed those extra lines, i need to check the configuration, i hope this is not a common theme

aepfli: nvm this was just an error on my side ;) sorry for all the  unnecessary changes, with making the javadoc multiline again

aepfli (approved):

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/common/Util.java
warber: @chrfwow did some additional tests on that. Turned out that we are not dealing with a recursion that will end up in a SO. So this PR contains the implementation that is using `monitorChannelState`. I will also create a separate branch/PR and immediately close it that will contain the alternative polling logic, just in case.. :) Because I don't think making this configurable and maintain both approaches is worth it.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
```diff
@@ -80,7 +81,7 @@ public synchronized void initialize(EvaluationContext evaluationContext) throws
         }
 
         this.flagResolver.init();
-        this.initialized = true;
+        this.initialized = this.connected = true;
```
warber, line 84: i think here we have a potential race condition between event emiting stuff happening in `init()`and setting these fields and the logic which depends on them ([onConnectionEvent(...)](https://github.com/open-feature/java-sdk-contrib/pull/1112/files#diff-7ac82836e0450dd1f6581536670fba629b6fa21a8ccb4a5ffcdce54a53dd57e3R158))

aepfli: i do think, that the state of initialization is not bound to the connection state. i can be initialized but not connected, or there could have been a connection issue in the middle. Maybe it would be better if the `.init()` method returns a boolean. The provider it self should/does not know about the status connected in theory. because that is something, which happens in this pr purely because we only have grpc connection. But on the other hand for inprocess, we do have a filewatcher, what does connected mean here? is the method onConnectionEvent correct, or is this more an abstraction of the underlying provider events, which we could directlu y use? All this discussion brings me more to the conclusio, that this connection logic etc should be part of the connector, and not of this outside class. 

As this pr is already quiet big, and this is a potential race condition but with limited impact, i would suggest to move forward with this pr as is. Start the migration of the in-process to this really cool and neat reusable implementation of our connector. Afterwards we can tackle this in a better and controlled manner, as soon as both implementations are normalized

toddbaert: Agreed

toddbaert (approved):

PR #1115, opened 2024-12-28
fix: rpc caching not behaving as expected (cleared too often)

aepfli:
With this pull request, I am fixing a regression within the RPC provider. We cleared the cache when there was an error, but we had to wait until we hit the error state and not stay in the stale state for the clean to happen. I also used this to normalize the eventing a little bit and reduce the complexity.

Additionally, I updated the e2e tests to the newest Gherkin version:
- missing tests: targetUri (will be own task)
- error and mismatch as the spec file does not match this configuration anymore (will be added to the gherkin files)


providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/resolver/common/GrpcConnectorTest.java
```diff
@@ -71,11 +72,12 @@ private void tearDownGrpcServer() throws InterruptedException {
     }
 
     @Test
+    @Disabled("not sure this test makes sense in this kind of way")
```
aepfli: if it is okay, i will delete it

aepfli:
blocked by https://github.com/open-feature/flagd-testbed/pull/184

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/common/ChannelMonitor.java
```diff
@@ -32,7 +32,7 @@ public static void monitorChannelState(
             Runnable onConnectionLost) {
         channel.notifyWhenStateChanged(expectedState, () -> {
             ConnectivityState currentState = channel.getState(true);
-            log.info("Channel state changed to: {}", currentState);
+            // log.info("Channel state changed to: {}", currentState);
```
warber: should we change this to debug?

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
```diff
@@ -31,10 +37,29 @@ public class FlagdProvider extends EventProvider {
...
-    private volatile boolean connected = false;
     private volatile Structure syncMetadata = new ImmutableStructure();
     private volatile EvaluationContext enrichedContext = new ImmutableContext();
     private final List<Hook> hooks = new ArrayList<>();
+    private volatile ProviderEvent previousEvent = null;
+
+    /**
+     * An executor service responsible for scheduling reconnection attempts.
+     */
+    private final ScheduledExecutorService reconnectExecutor;
```
warber: I think we should rename this to e.g. `scheduledErrorNotifier` or something and adjust the doc.

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/resolver/common/GrpcConnectorTest.java
warber: i also think this does not make sense anymore

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/State.java
```diff
@@ -0,0 +1,26 @@
...
+import dev.openfeature.contrib.providers.flagd.e2e.steps.FlagSteps;
+import dev.openfeature.contrib.providers.flagd.e2e.steps.ProviderType;
+import dev.openfeature.sdk.Client;
+import dev.openfeature.sdk.FlagEvaluationDetails;
+import dev.openfeature.sdk.MutableContext;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Optional;
+
+public class State {
```
toddbaert, line 15: Oh this is an interesting approach, I like it.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
```diff
@@ -151,47 +187,75 @@ EvaluationContext getEnrichedContext() {
...
-        syncMetadata = connectionEvent.getSyncMetadata();
-        enrichedContext = contextEnricher.apply(connectionEvent.getSyncMetadata());
+        switch (flagdProviderEvent.getEvent()) {
+            case PROVIDER_CONFIGURATION_CHANGED:
+                if (previousEvent == ProviderEvent.PROVIDER_READY) {
+                    this.emitProviderConfigurationChanged(ProviderEventDetails.builder()
+                            .flagsChanged(flagdProviderEvent.getFlagsChanged())
+                            .message("configuration changed")
+                            .build());
+                    break;
```
toddbaert: Maybe for consistency we should also extract this into a method but it's a nit.

toddbaert:
Every run I get a timeout locally on 1 assertion:

```
@rpc @in-process @grace
Scenario: Provider events chain ready -> stale -> error -> ready # test-harness/gherkin/events.feature:21
  Given a stable flagd provider                                  # dev.openfeature.contrib.providers.flagd.e2e.steps.ProviderSteps.setupProvider(java.lang.String)
  Given a ready event handler                                    # dev.openfeature.contrib.providers.flagd.e2e.steps.EventSteps.a_stale_event_handler(java.lang.String)
  And a error event handler                                      # dev.openfeature.contrib.providers.flagd.e2e.steps.EventSteps.a_stale_event_handler(java.lang.String)
  And a stale event handler                                      # dev.openfeature.contrib.providers.flagd.e2e.steps.EventSteps.a_stale_event_handler(java.lang.String)
  Then the ready event handler should have been executed         # dev.openfeature.contrib.providers.flagd.e2e.steps.EventSteps.eventHandlerShouldBeExecuted(java.lang.String)
  When the connection is lost for 6s                             # dev.openfeature.contrib.providers.flagd.e2e.steps.ProviderSteps.the_connection_is_lost_for(int)
  Then the stale event handler should have been executed         # dev.openfeature.contrib.providers.flagd.e2e.steps.EventSteps.eventHandlerShouldBeExecuted(java.lang.String)
  
  // HANGS HERE FOR A WHILE AND THEN:
  
  Then the error event handler should have been executed         # dev.openfeature.contrib.providers.flagd.e2e.steps.EventSteps.eventHandlerShouldBeExecuted(java.lang.String)
      org.awaitility.core.ConditionTimeoutException: Condition with lambda expression in dev.openfeature.contrib.providers.flagd.e2e.steps.EventSteps was not fulfilled within 30 seconds.
```

I've ran the suite about 10 times and I cannot get it to pass.

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/steps/ProviderSteps.java
```diff
@@ -206,7 +206,7 @@ public void the_connection_is_lost_for(int seconds) throws InterruptedException,
         String randomizer = RandomStringUtils.randomAlphanumeric(5);
         String timoutName = "restart" + randomizer;
         Proxy proxy = toxiproxyClient.getProxy(generateProxyName(State.resolverType, state.providerType));
-        proxy.toxics().timeout(timoutName, ToxicDirection.UPSTREAM, seconds);
+        proxy.toxics().timeout(timoutName, ToxicDirection.DOWNSTREAM, seconds);
```
toddbaert: I had to change this to downstream to fix [my issue](https://github.com/open-feature/java-sdk-contrib/pull/1115#issuecomment-2593446267)... hopefully it doesn't break the e2e tests in CI, but TBH this setting makes more sense to me. :thinking:

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/steps/ProviderSteps.java
```diff
@@ -204,16 +204,19 @@ public void setupProvider(String providerType) throws IOException {
     public void the_connection_is_lost_for(int seconds) throws InterruptedException, IOException {
         log.info("Timeout and wait for {} seconds", seconds);
         String randomizer = RandomStringUtils.randomAlphanumeric(5);
-        String timoutName = "restart" + randomizer;
+        String timeoutUpName = "restart-up-" + randomizer;
+        String timeoutDownName = "restart-down-" + randomizer;
         Proxy proxy = toxiproxyClient.getProxy(generateProxyName(State.resolverType, state.providerType));
-        proxy.toxics().timeout(timoutName, ToxicDirection.UPSTREAM, seconds);
+        proxy.toxics().timeout(timeoutDownName, ToxicDirection.DOWNSTREAM, seconds);
+        proxy.toxics().timeout(timeoutUpName, ToxicDirection.UPSTREAM, seconds);
```
toddbaert, line 211: OK I had to change it to BOTH directions.

toddbaert (approved):
Amazing work. Minor nit and I made one small change [here](https://github.com/open-feature/java-sdk-contrib/pull/1115#discussion_r1917066303) which I hope you don't mind. Otherwise on my system the timeout was not detected.

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/resolver/common/GrpcConnectorTest.java
aepfli: deleted

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
aepfli: Extracted

aepfli:
All dependencies are resolved. Ready to merge ;)

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/grpc/GrpcStreamConnector.java
```diff
@@ -126,6 +121,7 @@ void observeEventStream(
                     metadataException = e;
                 }
 
+                log.info("stream");
```
warber, line 121:
```suggestion
```

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
```diff
@@ -81,17 +111,23 @@ public synchronized void initialize(EvaluationContext evaluationContext) throws
         }
 
         this.flagResolver.init();
-        this.initialized = this.connected = true;
+        // block till ready - this works with deadline fine for rpc, but with in_process we also need to take parsing
+        // into the equation
+        // TODO: evaluate where we are losing time, so we can remove this magic number - follow up
+        Util.busyWaitAndCheck(this.deadline + 500, () -> initialized);
```
chrfwow: Do we really want to wait busily?

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
```diff
@@ -151,47 +187,86 @@ EvaluationContext getEnrichedContext() {
...
         }
-
-        if (!wasConnected && isConnected) {
-            ProviderEventDetails details = ProviderEventDetails.builder()
-                    .flagsChanged(connectionEvent.getFlagsChanged())
-                    .message("connected to flagd")
-                    .build();
-            this.emitProviderReady(details);
-            return;
+        if (reconnectTask != null && !reconnectTask.isCancelled()) {
```
chrfwow: Don't we also have to cancel the `errorExecutor`?

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
```diff
@@ -151,47 +187,86 @@ EvaluationContext getEnrichedContext() {
...
-                    .message("there has been an error")
-                    .build());
-        } else {
-            this.emitProviderError(ProviderEventDetails.builder()
-                    .message("there has been an error")
-                    .build());
+        if (!errorExecutor.isShutdown()) {
+            reconnectTask = errorExecutor.schedule(
+                    () -> {
+                        log.debug(
```
chrfwow: I think we should also check here in a thread safe manner if we are still in a stale/error state

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
```diff
@@ -151,47 +187,86 @@ EvaluationContext getEnrichedContext() {
         return enrichedContext;
     }
 
-    private boolean isConnected() {
-        return this.connected;
-    }
+    @SuppressWarnings("checkstyle:fallthrough")
+    private void onProviderEvent(FlagdProviderEvent flagdProviderEvent) {
```
chrfwow, line 209: Is there a mechanism somewhere to ensure that this method is only ever executed by a single thread at a time?

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/steps/EventSteps.java
```diff
@@ -0,0 +1,70 @@
...
+
+    @Given("a {} event handler")
+    public void a_stale_event_handler(String eventType) {
+        state.client.on(mapEventType(eventType), eventDetails -> {
+            log.info("event tracked for {} ", eventType);
+            state.events.add(new Event(eventType, eventDetails));
+        });
+    }
+
+    private static @NotNull ProviderEvent mapEventType(String eventType) {
```
chrfwow, line 33: Could this be a candidate for the Utils class?

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/RunRpcTest.java
```diff
@@ -0,0 +1,38 @@
...
+import org.testcontainers.junit.jupiter.Testcontainers;
+
+/**
+ * Class for running the reconnection tests for the RPC provider
+ */
+@Order(value = Integer.MAX_VALUE)
+@Suite
+@IncludeEngines("cucumber")
+@SelectDirectories("test-harness/gherkin")
+// @SelectFile("test-harness/gherkin/rpc-caching.feature")
```
chrfwow, line 26: remove?

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/RunInProcessTest.java
```diff
@@ -0,0 +1,38 @@
...
+import org.testcontainers.junit.jupiter.Testcontainers;
+
+/**
+ * Class for running the reconnection tests for the RPC provider
+ */
+@Order(value = Integer.MAX_VALUE)
+@Suite
+@IncludeEngines("cucumber")
+@SelectDirectories("test-harness/gherkin")
+// @SelectFile("test-harness/gherkin/connection.feature")
```
chrfwow, line 26: remove?

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
```diff
@@ -81,17 +111,23 @@ public synchronized void initialize(EvaluationContext evaluationContext) throws
...
     }
 
     @Override
     public synchronized void shutdown() {
         if (!this.initialized) {
             return;
         }
-
         try {
             this.flagResolver.shutdown();
```
chrfwow:
```java
this.flagResolver.shutdown();
this.reconnectTask.cancel(true);
```
do we need to cancel the `reconnectTask` as well?

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
aepfli: I have to admit I am not good with concurrency. But I am not sure we must ensure this is only executed once. (naive thought). The `previousEvent` is volatile, so we should see modification throughout all threads (naive knowledge). Or is my understanding wrong here, or do i miss some crucial possibilities?

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
aepfli: Please provide a little bit more reasoning. I am unsure what you're aiming at; why should I cancel the errorExecutor, too? The executor is used for the whole instance and is only canceled when shut down. So I can feed new tasks into it on demand. Ultimately, I want to ensure that the reconnect task is not executed. But I still want to continue using my executor.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
toddbaert: I think @chrfwow is right about this one, we don't want any method calls here to interleave or we can end up in weird states.

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/steps/EventSteps.java
aepfli: It is specific to the events, hence I do not think it is worth to migrate to a different file.

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/RunRpcTest.java
aepfli: I left this on purpose, so if somebody wants to test a dedicated feature file, they do have a pointer ok how to do it.

aepfli: But adding a comment might be a good idea

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
aepfli: I will rename the task, but it is executed via the error executor - so it will be cancelled in the next line anyways

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
aepfli: For now, yes, this has just been moved further down to up here.

We should generate a task to tackle this specifically on its own and maybe use a more elegant solution. Suggestions and issues are highly welcome.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
aepfli: added synchronized to the method

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/RunRpcTest.java
aepfli: added comment

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/RunInProcessTest.java
aepfli: added comment to signal the purpose of this commented out code

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
aepfli: added thank you nice catch

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
aepfli: created https://github.com/open-feature/java-sdk-contrib/issues/1160 to tackle this in a nicer way

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
```diff
@@ -151,47 +189,91 @@ EvaluationContext getEnrichedContext() {
...
 
-    private boolean isConnected() {
-        return this.connected;
-    }
+    @SuppressWarnings("checkstyle:fallthrough")
+    private void onProviderEvent(FlagdProviderEvent flagdProviderEvent) {
+        log.info("FlagdProviderEvent: {}", flagdProviderEvent);
+        syncMetadata = flagdProviderEvent.getSyncMetadata();
+        if (flagdProviderEvent.getSyncMetadata() != null) {
+            enrichedContext = contextEnricher.apply(flagdProviderEvent.getSyncMetadata());
```
chrfwow: Writing to `syncMetadata` and `enrichedContext` should also happen in the synchronized block to be consistent with `previousEvent`. I think locking on the `eventLock` is fine (as opposed to locking on `this`, which I believe could make problems with the `initialize` method, but I am not 100% sure)

chrfwow (changes requested):

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
aepfli: Done in the latest push

chrfwow (approved):

toddbaert:
I pushed [one more small change](https://github.com/open-feature/java-sdk-contrib/pull/1115/commits/3d2b826deb70f5e7d8a232e61657340545b5fa61) to move all the volatile fields on the provider into a dedicated object to be used as the intrinsic lock in our synchronized blocks. I think this has the benefit of being very clear but also it unifies all our locking onto a single object.

PR #1121, opened 2025-01-01
build: change checkstyle to google code format, plus adding spotless

aepfli:
We regularly fight with code styles in pull requests, which is annoying to both the contributors and the maintainers.

There are tools out there, like Spotless, that can automate those processes. We could define our code style, but there is already one widely used code style from Google (Google Java format).

Existing plugins help us follow this code style within IntelliJ and Visual Studio Code. Switching to this definition can improve the overall experience.

This supersedes the pull request with `.editorconfig`

## notes

- first commit contains the changes to the tooling
- second commit contains the applied formatting changes

## Disclaimer

This pull request should be used to discuss this; most people do not have significant opinions regarding code styles as long as there are tools to enforce them. Now we have a maven target, `spotless:apply`, which will inform you about violations and fix them.


pom.xml
```diff
@@ -440,7 +444,48 @@
...
+                            <endWithNewline/>
+                            <indent>
+                                <spaces>true</spaces>
+                                <spacesPerTab>4</spacesPerTab>
+                            </indent>
+                        </format>
+                    </formats>
+                    <!-- define a language-specific format -->
+                    <java>
+                        <palantirJavaFormat />
```
justinabrahms, line 473: From the PR description, I expected us to be using Google's style?

justinabrahms (commented):
Seems.. fine? No question from me that style should be enforced by robots.

From the slack thread, I expected the rule changes to be more significant. Is the real problem here "The checkstyle that we had before is under-specified and we need to use more modules; Also... add spotless plugin"?

pom.xml
aepfli: Palantir code style si a fork of Google code style with better lambda formatting - lesser indentations and improved readability for lambdas. The rest is basic Google java Format.

https://github.com/palantir/palantir-java-format

aepfli:
> Seems.. fine? No question from me that style should be enforced by robots.
> 
> From the slack thread, I expected the rule changes to be more significant. Is the real problem here "The checkstyle that we had before is under-specified and we need to use more modules; Also... add spotless plugin"?

The check style changes are more related to be fact that i  wanted to use spotless to ease the pain. But our check style configuration was not compatible with the default settings. 

As I favor simplicity over custom specification, I thought it is easier to migrate to google check style and a adapt the full java code Format configuration, rather than making spotless work with our current check style config.

justinabrahms (approved):
"Make config work well w/ the auto enforcer" works for me as a rationale. Thanks so much!

guidobrei:
Thanks so much for adding this!
Should we also document this in the Contribution guide?

aepfli:
https://github.com/open-feature/java-sdk/pull/1264 - sibling pr for java-sdk

aepfli:
> Thanks so much for adding this! Should we also document this in the Contribution guide?

added a section, with a detailed explanation

toddbaert (approved):
As a vscode user, thanks so much! :heart:

chrfwow (approved):

PR #1122, opened 2025-01-02
feat: Update in-process resolver to support flag metadata #1102

### LINKED ISSUE #1102: [flagd] Update in-process resolver to support flag metadata
With [this schema change](https://github.com/open-feature/flagd-schemas/pull/173), flag metadata can now be stored in flag configuration files. Specifically, these can contain things like `version` and  flag set `id` information so that consumers can use this data in flag evaluations, caching, telemetry etc.

We need to add this metadata into flags which are evaluated by in-process resolvers. Essentially, we need to take the new properties from the flag definition and include them in metadata returned in the [flag resolution](https://github.com/open-feature/java-sdk-contrib/blob/2fc6af0994e6643b704e0f1cd7aba8fa5af688f6/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolver.java#L154).

Acceptance criteria:

- metadata defined in flag configuration (specifically `id` and `version`) are returned from any flag resolved from this set
- associated tests

chrfwow:
## This PR
Adds support for flag metadata in the in-process resolver

### Related Issues

Fixes #1102


providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolver.java
```diff
@@ -226,12 +238,51 @@ private <T> ProviderEvaluation<T> resolve(Class<T> type, String key, EvaluationC
...
+                .flagMetadata(getFlagMetadata(flag))
+                .build();
+    }
+
+    private ImmutableMetadata getFlagMetadata(FeatureFlag flag) {
+        if (flag == null) {
+            return fallBackMetadata;
+        }
+
+        ImmutableMetadata.ImmutableMetadataBuilder metadataBuilder = ImmutableMetadata.builder();
```
chrfwow: I would prefer to have a `mergeMetdata` or `addMetadata` method in the `ImmutableMetadataBuilder`

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolverTest.java
```diff
@@ -79,7 +81,7 @@ public void eventHandling() throws Throwable {
                 final MutableStructure syncMetadata = new MutableStructure();
                 syncMetadata.add(key, val);
 
-                InProcessResolver inProcessResolver = getInProcessResolverWth(new MockStorage(new HashMap<>(), sender),
+                InProcessResolver inProcessResolver = getInProcessResolverWith(new MockStorage(new HashMap<>(), sender),
```
chrfwow: I assume this was a typo

toddbaert:
@chrfwow this is definitely a step in the right direction - I think there's probably some null check missing related to the metadata causing the e2e failures: `Run 1: Cannot invoke "dev.openfeature.sdk.Structure.asMap()" because "structure" is null`

Working with @beeme1mr and a few others in in slack, we've decided another thing we'd want here is a very basic mechanism for reducing duplication: the ability to add metadata for the entire flag set at the "top level" which all flags inherit:

```json
{
  "metadata": {
    "foo": "bar"
  },
  "flags": {
    "flag1": {
      // ...
      "metadata": {
        // "foo":  "bar" is automatically included, but can be overridden if defined in `flag1`
      }
    }
  }
}
```

The same thing is implemented here: https://github.com/open-feature/js-sdk-contrib/pull/1120

See: https://github.com/open-feature/flagd-schemas/blob/37baa2cdea48a5ac614ba3e718b7d02ad4120611/json/flags.json#L53

cc @beeme1mr

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/model/FeatureFlag.java
```diff
@@ -23,18 +23,21 @@ public class FeatureFlag {
...
-    public FeatureFlag(
-            @JsonProperty("state") String state,
-            @JsonProperty("defaultVariant") String defaultVariant,
-            @JsonProperty("variants") Map<String, Object> variants,
-            @JsonProperty("targeting") @JsonDeserialize(using = StringSerializer.class) String targeting) {
+    public FeatureFlag(@JsonProperty("state") String state,
+                       @JsonProperty("defaultVariant") String defaultVariant,
+                       @JsonProperty("variants") Map<String, Object> variants,
+                       @JsonProperty("targeting") @JsonDeserialize(using = StringSerializer.class) String targeting,
+                       @JsonProperty("metadata") Map<String, Object> metadata
```
aepfli: [question] Currently, we do need to provide an empty map everywhere to be backward compatible. Could we also omit this parameter and make it optional? more a question out of curiosity ;)

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/resolver/process/model/FlagParserTest.java
```diff
@@ -66,16 +87,58 @@ public void validJsonConfigurationWithTargetingRulesParsing() throws IOException
     }
 
     @Test
-    public void invalidFlagThrowsError() {
-        assertThrows(IllegalArgumentException.class, () -> {
-            FlagParser.parseString(getFlagsFromResource(INVALID_FLAG), true);
-        });
+    void validJsonConfigurationWithGlobalMetadataParsing() throws IOException {
```
toddbaert: Tests look good, thanks; I think I see all obvious the cases tested.

toddbaert (approved):

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolver.java
```diff
@@ -178,6 +183,7 @@ private <T> ProviderEvaluation<T> resolve(Class<T> type, String key, EvaluationC
             return ProviderEvaluation.<T>builder()
                     .errorMessage("flag: " + key + " not found")
                     .errorCode(ErrorCode.FLAG_NOT_FOUND)
+                    .flagMetadata(fallBackMetadata)
```
toddbaert: As a corollary to the inheritance of the flag set metadata, for error cases we should make a "best effort" to return whatever metadata we can - so here we should still return the flag set metadata. We've added this to the spec: https://github.com/open-feature/flagd/pull/1505/files#diff-28ba4b4041b6c39c4b9deb542276312bebdd08f45b39b29d860849d9e8b81a6eR297

The simplest way might be to store the flag set metadata on the flag store and expose it there.

This will greatly help in telemetry and debugging, since we'll have some idea of the flag set, it's version, etc in error reporting and metrics.

toddbaert (changes requested):
I found one small issue [here](https://github.com/open-feature/java-sdk-contrib/pull/1122#discussion_r1909417119)

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/StorageQueryResult.java
```diff
@@ -0,0 +1,24 @@
...
+ * To be returned by the storage when a flag is queried. Contains the flag (iff a flag associated with the given key
+ * exists, null otherwise) and global flag metadata
+ */
+@Getter
+@SuppressFBWarnings(
+        value = {"EI_EXPOSE_REP"},
+        justification = "The storage provides access to both feature flags and global metadata")
+public class StorageQueryResult {
+    private final FeatureFlag featureFlag;
+    private final Map<String, Object> globalFlagMetadata;
```
toddbaert:
```suggestion
    private final Map<String, Object> flagSetMetadata;
```
I understand why you prepended `global` to the name (since the metadata doesn't just belong to the _this_ flag, but all of them) but I think global is a bit confusing here. I think this name is perhaps better. WDYT?

If you agree with it here, I think you should change it in all cases.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolver.java
```diff
@@ -232,13 +233,59 @@ private <T> ProviderEvaluation<T> resolve(Class<T> type, String key, EvaluationC
...
                 .variant(resolvedVariant)
-                .reason(reason);
+                .reason(reason)
+                .flagMetadata(getFlagMetadata(storageQueryResult))
+                .build();
+    }
+
+    private ImmutableMetadata getFlagMetadata(StorageQueryResult storageQueryResult) {
+        ImmutableMetadata.ImmutableMetadataBuilder metadataBuilder = ImmutableMetadata.builder();
+        for (Map.Entry<String, Object> entry :
```
toddbaert, line 242: Do you think we should add some some sort of convenient way to do this in the SDK itself? We have [this](https://github.com/open-feature/java-sdk/blob/b997946db1c7663b7ebb775ad45cdb2b0aaeb291/src/main/java/dev/openfeature/sdk/ImmutableContext.java#L46) for the context.

I'm fine merging this as is, but wondering what you think about subsequent work.

toddbaert (approved):

beeme1mr (approved):

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolver.java
chrfwow: I do believe there should be some way to merge two or more `ImmutableMetadata` objects

PR #1153, opened 2025-01-11
test(flagd): add socket path tests

aepfli:
- finding, socket path is not supported for in_process

### DIFF
--- a/providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/RunFlagdInProcessSocketCucumberTest.java
+++ b/providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/RunFlagdInProcessSocketCucumberTest.java
@@ -1,14 +1,17 @@
 package dev.openfeature.contrib.providers.flagd.e2e;
 
-import static io.cucumber.junit.platform.engine.Constants.GLUE_PROPERTY_NAME;
-import static io.cucumber.junit.platform.engine.Constants.PLUGIN_PROPERTY_NAME;
-
 import org.apache.logging.log4j.core.config.Order;
 import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.condition.EnabledOnOs;
+import org.junit.jupiter.api.condition.OS;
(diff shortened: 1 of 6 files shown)


providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/RunFlagdInProcessSocketCucumberTest.java
```diff
@@ -0,0 +1,28 @@
...
+import org.testcontainers.junit.jupiter.Testcontainers;
+
+import static io.cucumber.junit.platform.engine.Constants.GLUE_PROPERTY_NAME;
+import static io.cucumber.junit.platform.engine.Constants.PLUGIN_PROPERTY_NAME;
+
+@Order(value = Integer.MAX_VALUE)
+@Suite(failIfNoTests = false)
+@IncludeEngines("cucumber")
+// Not implemented in flagd
+//@SelectFile("spec/specification/assets/gherkin/evaluation.feature")
```
aepfli, line 20: again, this does not work for in_process connection

toddbaert (dismissed):

toddbaert:
@aepfli I'm not sure if you noticed, but there's an OOM error in the CI.

aepfli:
Yes, I have seen it, and I am currently also investigating. This is an odd issue. How do you simulate a reconnect with an SSL socket? I am not sure this is something testable, as a socket connection is a file connection. and imho, if that gets delete, i don't think we can recover, wdyt?

toddbaert:
> Yes, I have seen it, and I am currently also investigating. This is an odd issue. How do you simulate a reconnect with an SSL socket? I am not sure this is something testable, as a socket connection is a file connection. and imho, if that gets delete, i don't think we can recover, wdyt?

I think we should avoid reconnection tests for the file mode. I don't think they make sense and I don't even know if it's possible. Definitionally, a unix socket should be on a shared disk; if that disk is not accessible I think it's out of our hands and probably we don't need to support reconnection. Really the socket is to enable very fast local communication by avoiding the TCP stack.

aepfli:
Superseded by #1115

PR #1169, opened 2025-01-21
feat: Improve wait logic to a more elegant solution #1160

### LINKED ISSUE #1160: [flagd] improve wait logic to a more elegant solution
Our current waiting logic could be improved. It is a working wait logic, but it might not be the most elegant solution.
https://github.com/open-feature/java-sdk-contrib/blob/cd0ea9e8d3d303b675b5d33287d94394e3c3aa6c/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/common/Util.java#L26-L38

This task aims to migrate to an approach with `notify()` and `wait()`. As suggested by @chrfwow within #1115.

chrfwow:
## This PR
Adds a better wait logic to wait for the initialization of the `FlagdProvider`.

With these changes, it is not advisable to restart a `FlagdProvider` once it has been shut down.

### Related Issues

Fixes #1160


providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/EventsLock.java
```diff
@@ -0,0 +1,100 @@
...
+import dev.openfeature.sdk.exceptions.GeneralError;
+import lombok.Getter;
+import lombok.Setter;
+
+/**
+ * Contains all fields we need to worry about locking, used as intrinsic lock
+ * for sync blocks.
+ */
+@Getter
+public class EventsLock {
```
chrfwow: I moved this class out of the `FlagdProvider` to ensure no private fields are accessed and to keep the size of the file down.
However, I feel like the name is not approrpaite anymore, so please feel free to suggest a better one. Also, this new class is strongly coupled to the `FlagdProvider`, if you have any suggestions to make it cleaner please let me know

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
```diff
@@ -119,27 +118,26 @@ public List<Hook> getProviderHooks() {
...
-            if (eventsLock.initialized) {
+            if (eventsLock.isInitialized()) {
                 return;
             }
 
             flagResolver.init();
+            // block till ready - this works with deadline fine for rpc, but with in_process
+            // we also need to take parsing into the equation
+            // TODO: evaluate where we are losing time, so we can remove this magic number -
+            eventsLock.waitForInitialization(this.deadline * 2);
```
aepfli: We purposely moved this out of the sync block, as the eventhandler also syncs on the event lock. is this now a problem that this is within the sync block?

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/FlagdProviderTest.java
```diff
@@ -5,6 +5,7 @@
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotEquals;
 import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.mockito.AdditionalAnswers.*;
```
aepfli: please try to avoid star imports

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/EventsLock.java
aepfli: This class is only used by our FlagDProvider, and nobody else should use it or instantiate it. Hence, it might be worth making the package private, wdyt?

aepfli: maybe `FlagdProviderSyncState` or similar could be an appropriate name - or just `SyncState` or `SyncResource`  - but to be fair, i am also not hundred percent happy with my suggestions

aepfli (approved):
thank you - i just added some questions and suggestions, nothing crucial except the one about where we are doing this wait and notify

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
chrfwow: The new implementation does a `wait(long)` call, through which it gives up the lock on the object until the thread is woken up. Then, another thread can acquire the lock and do the initialization. 
I moved the `waitForInitialization` call inside the synchronized as there is an additional `synchronize` on the eventsLock object inside the method. This way we do not release the lock and acquire it immediately afterwards, which should be more efficient.

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/FlagdProviderTest.java
chrfwow: My bad. But shouldn't spotless clean such things up?

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/EventsLock.java
chrfwow: I would love to, but then it is not visible in the tests

aepfli: There is a `@VisibleForTesting` annotation, does this maybe help?

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/FlagdProviderTest.java
aepfli: i think, this is one of the things spotless is not able to properly address sadly. need to check, maybe we add a checkstyle rule after all for that

chrfwow:
~~Waiting on [https://github.com/open-feature/java-sdk/issues/1299](https://github.com/open-feature/java-sdk/issues/1299)~~ completed

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/ConfigCucumberTest.java
```diff
@@ -16,7 +16,8 @@
 @Order(value = Integer.MAX_VALUE)
 @Suite
 @IncludeEngines("cucumber")
-@SelectFile("test-harness/gherkin/config.feature")
+// @SelectFile("test-harness/gherkin/config.feature")
+@SelectDirectories("test-harness/gherkin")
```
aepfli: this file is just a unit test based on gherkin tests, and not a e2e test. we purposely selected only the config file. the execution will fail, as it will not be able to find all the steps

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/steps/ProviderSteps.java
```diff
@@ -125,28 +125,26 @@ public void setupProvider(String providerType) throws IOException, InterruptedEx
                 .statusCode(200);
 
         // giving flagd a little time to start
-        Thread.sleep(30);
+        Thread.sleep(100);
         FeatureProvider provider =
                 new FlagdProvider(state.builder.resolverType(State.resolverType).build());
 
         OpenFeatureAPI api = OpenFeatureAPI.getInstance();
-        String providerName = providerType + Math.random();
```
aepfli, line 133: I think the providername adds a little bit more separation to the execution, so each e2e test has its own provider, without any sideeffects from previous tests, wdyt should we maybe keep this?

chrfwow: Since we recreate the api for every test, I don't think that the providers for each test can interfere with each other. If we happen to have more than one provider per api/ more than one api, this would be a whole other problem

aepfli: We do not recreate the API for e2e tests. We still utilize the singleton as the end user would use it.

chrfwow: True, I misread that. But maybe we should?
I will reintroduce names for the provider

aepfli: E2E tests should be closer to the consumer, simulating how they are used, we should not fall back to separate API instances, we might hide problems, which we initially planned to detect with e2e tests

aepfli (approved):

aepfli (approved):

toddbaert:
I'll be sure to review this tomrrow.

pom.xml
```diff
@@ -68,8 +68,8 @@
             <!-- this can be overriden in child POMs to support specific SDK requirements -->
             <groupId>dev.openfeature</groupId>
             <artifactId>sdk</artifactId>
-            <!-- 1.14 <= v < 2.0 (excluding 2.0 pre-releases)-->
-            <version>[1.14,1.99999)</version>
+            <!-- 1.14.1 <= v < 2.0 (excluding 2.0 pre-releases)-->
+            <version>[1.14.1,1.99999)</version>
```
toddbaert, line 78: :+1:

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/FlagdProviderSyncResourcesTest.java
```diff
@@ -0,0 +1,140 @@
...
+import dev.openfeature.sdk.exceptions.GeneralError;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicLong;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.Timeout;
+
+class FlagdProviderSyncResourcesTest {
+    private static final long PERMISSIBLE_EPSILON = 20;
```
toddbaert: Perhaps rename this to for the benefit of those of us who haven't been in a calculus class for a decade :sweat_smile:

toddbaert (approved):
Great work! :pray:

toddbaert:
I'll merge this tomorrow unless I hear objections.

PR #1173, opened 2025-01-22
feat(flagd): migrate file to own provider type

aepfli:
(https://github.com/open-feature/flagd/pull/1525)


providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdOptions.java
```diff
@@ -193,7 +193,20 @@ void prebuild() {
                 resolverType = fromValueProvider(System::getenv);
             }
 
-            if (port == 0) {
+            if (StringUtils.isEmpty(offlineFlagSourcePath)) {
```
chrfwow:
```suggestion
if (StringUtils.isBlank(offlineFlagSourcePath)) {
   if (resolverType == Config.Resolver.FILE) {
      throw new IllegalArgumentException("Resolver Type 'FILE' requires an offlineFlagSourcePath");
   }
   offlineFlagSourcePath = fallBackToEnvOrDefault(Config.OFFLINE_SOURCE_PATH, null);
} else if (resolverType == Config.Resolver.IN_PROCESS) {
   resolverType = Config.Resolver.FILE;
}
```

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/steps/ProviderSteps.java
```diff
@@ -78,10 +78,23 @@ public void setupProvider(String providerType) throws IOException, InterruptedEx
         String flagdConfig = "default";
         state.builder.deadline(1000).keepAlive(0).retryGracePeriod(2);
         boolean wait = true;
+        File flags = new File("test-harness/flags");
```
chrfwow: Why do we need to do this for every provider type? The `offlinePath` seems to be only used in the `default` case

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/steps/ProviderSteps.java
```diff
@@ -103,26 +116,17 @@ public void setupProvider(String providerType) throws IOException, InterruptedEx
...
+                if (State.resolverType == Config.Resolver.FILE) {
+
+                    state.builder
+                            .port(UNAVAILABLE_PORT)
+                            .offlineFlagSourcePath(offlinePath.toAbsolutePath().toString());
+                } else {
+                    state.builder.port(container.getPort(State.resolverType));
+                }
                 break;
         }
```
chrfwow, line 123: I can't add a comment further down, but could you please remove the naming of the provider in lines 142-148? This was added when we fixed the flaky test, but is of no use now

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/steps/ProviderSteps.java
aepfli: i need to drop this, we do not need to create this anymore, we can use a tempfile

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdOptions.java
aepfli: I don't think this is correct behavior. Before I can throw the error, I need to check if there might be an environment variable set for the offline source path.

1. fetch the variables in all kinds of flavors
2. if the path is set, and we do have in-process resolver, reset the resolver type
3. if the path is not set, and we're in the file type throw an error

the second step is an optional one, ensuring backward compatibility; people might use the right provider type already but don't have to.

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/steps/ProviderSteps.java
aepfli: dropped, and added the new file approach

toddbaert (approved):

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/steps/ProviderSteps.java
aepfli: it seems like this makes our tests more stable, i would leave that in

beeme1mr (approved):
Please make sure to update the readme to this "new" provider type.

aepfli:
> Please make sure to update the readme to this "new" provider type.

also added the missing field of poll interval now to the implementation

PR #1205, opened 2025-02-08
build: utilizing new java version, but still building with language level 8

aepfli:
build runs on 2 different java versions:
- 17: all tests and code quality tools
- 8: all tests

build target is version 8, but we use java 17 to generate the packages.

We can at least update checkstyle, spotless, pmd, spotbugs to newer versions, ~~but junit, mockito, assertj, cucumber, can not be updated as we are still running the tests with java 8.~~ Tests are not compiled on java8 so we could update all our test dependencies too. eg. i update mockito to version 5 within this pr

Theoretically, we could not run tests with java8 and only build, to ensure api compatibility :)

as the build is strangely not working now in here - see https://github.com/open-feature-forking/java-sdk-contrib/actions/runs/13217059965 of my fork

## how to test:

```shell
# java 17 and above
mvn verify

#java8 
mvn verify -Pjava8
```

if you want to test, that this wont build successfully on java 8 with newer features, use eg. `String.isBlank()` (java 11) on any string and try to build with java8 :) it will even fail with the default build, informing you, that this is not supported by the language level 8.

## Additional findings:

With this approach we could build everything with Java 17, but target with the release option the desired java version. This would allow different providers to have different baselines. Eg. some providers might want to upgrade to java 11 already. Giving our providers more flexibility overall.

## Todos:

- [x] adapt release please config

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,5 @@
 name: CI
 on:
-  push:
   pull_request:
     types:
       - opened
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,15 +11,23 @@ on:
 jobs:
   main:
     runs-on: ubuntu-latest
-
+    strategy:
+        matrix:
+            build:
+                - java: 17
+                  profile: codequality
+                - java: 11
+                  profile: java11
+                - java: 8
+                  profile: java8
     steps:
       - name: Checkout Repository
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
 
-      - name: Set up JDK 8
+      - name: Set up JDK ${{ matrix.build.java }}
         uses: actions/setup-java@3a4f6e1af504cf6a31855fa899c6aa5355ba6c12 # v4
         with:
-          java-version: '8'
+          java-version: ${{ matrix.build.java }}
           distribution: 'temurin'
           cache: maven
 
@@ -27,9 +35,9 @@ jobs:
         uses: actions/cache@0c907a75c2c80ebcb7f088228285e798b750cf8f # v4
         with:
           path: ~/.m2/repository
-          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+          key: ${{ runner.os }}${{ matrix.build.java }}-maven-${{ hashFiles('**/pom.xml') }}
           restore-keys: |
-            ${{ runner.os }}-maven-
+            ${{ runner.os }}${{ matrix.build.java }}-maven-
 
       - name: Maven Verify
-        run: mvn --batch-mode --activate-profiles e2e clean verify
+        run: mvn --batch-mode --activate-profiles e2e,${{ matrix.build.profile }} clean verify
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,5 +1,6 @@
 name: CI
 on:
+  push:
   pull_request:
     types:
       - opened
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,8 +17,6 @@ jobs:
           build:
               - java: 17
                 profile: codequality
-              - java: 11
-                profile: java11
               - java: 8
                 profile: java8
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,16 +10,19 @@ on:
 
 jobs:
   main:
-    runs-on: ubuntu-latest
     strategy:
         matrix:
-            build:
-                - java: 17
-                  profile: codequality
-                - java: 11
-                  profile: java11
-                - java: 8
-                  profile: java8
+          os: [ubuntu-latest]
+          build:
+              - java: 17
+                profile: codequality
+              - java: 11
+                profile: java11
+              - java: 8
+                profile: java8
+
+    name: with Java ${{ matrix.build.java }}
+    runs-on: ${{ matrix.os}}
     steps:
       - name: Checkout Repository
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -1,5 +1,6 @@
 name: CI
 on:
+  push:
```
aepfli, line 3: - [ ] remove this before merging

pom.xml
```diff
@@ -437,93 +376,227 @@
...
+                            <consoleOutput>true</consoleOutput>
+                            <failsOnError>true</failsOnError>
+                            <linkXRef>false</linkXRef>
+                            <excludeGeneratedSources>true</excludeGeneratedSources>
+                        </configuration>
+                        <dependencies>
+                            <dependency>
+                                <groupId>com.puppycrawl.tools</groupId>
+                                <artifactId>checkstyle</artifactId>
+                                <version>10.21.2</version>
```
aepfli, line 422: newest checkstyle version

pom.xml
```diff
@@ -437,93 +376,227 @@
...
+                                </goals>
+                            </execution>
+                        </executions>
+                    </plugin>
+
+                    <!-- Spotless -->
+                    <plugin>
+                        <groupId>com.diffplug.spotless</groupId>
+                        <artifactId>spotless-maven-plugin</artifactId>
+                        <version>2.43.0</version>
```
aepfli, line 492: new spotless version

pom.xml
```diff
@@ -437,93 +376,227 @@
...
+                                </goals>
+                            </execution>
+                        </executions>
+                    </plugin>
+
+                    <!-- SpotBugs -->
+                    <plugin>
+                        <groupId>com.github.spotbugs</groupId>
+                        <artifactId>spotbugs-maven-plugin</artifactId>
+                        <version>4.8.6.6</version>
```
aepfli, line 458: we can upgrade to version `4.9.1`

pom.xml
```diff
@@ -437,93 +376,227 @@
...
+                                    <version>1.13.0</version>
+                                </plugin>
+                            </plugins>
+                        </configuration>
+                        <dependencies>
+                            <!-- overwrite dependency on spotbugs if you want to specify the version of spotbugs -->
+                            <dependency>
+                                <groupId>com.github.spotbugs</groupId>
+                                <artifactId>spotbugs</artifactId>
+                                <version>4.8.6</version>
```
aepfli, line 474: we can upgrade to version `4.9.1`

```suggestion
                        <version>4.9.1</version>
```

.github/workflows/ci.yml
```diff
@@ -10,26 +11,37 @@ on:
 
 jobs:
   main:
-    runs-on: ubuntu-latest
+    strategy:
+        matrix:
+          os: [ubuntu-latest]
```
aepfli, line 15: we could add eg. windows and macos here, and would have a good validation, if the builds also work on other platforms. especially for development this could be interesting, eg. labeling builds only working on linux, etc.

checkstyle.xml
aepfli, line 1: now we are even closer to the official google java style checkstyle config

.github/workflows/ci.yml
```diff
@@ -10,26 +11,37 @@ on:
...
+          build:
+              - java: 17
+                profile: codequality
+              - java: 11
```
toddbaert: I'm not 100% we need to explicitly test Java 11 at the moment, since we are compiling on 8 and running all tests on 17. Between just those 2, we cover all our relevant checks. However, since we have a desire to make 11 the new minimum, we can keep this for now knowing eventually we will remove 8.

pom.xml
```diff
@@ -437,93 +377,229 @@
...
+                    </plugin>
+                </plugins>
+            </build>
+        </profile>
+
+        <!-- profile for running tests under java 8 (used mostly in CI) -->
+        <!-- selected automatically by JDK activeation (see https://maven.apache.org/guides/introduction/introduction-to-profiles.html#implicit-profile-activation) -->
+        <profile>
+            <id>java8</id>
+            <!-- with the next block we can define a set of sdks which still support java 8, if any of the sdks is not supporting java 8 anymore -->
```
toddbaert:
```suggestion
            <!-- with the next block we can define a set of artifacts which still support java 8, if any of the artifacts is not supporting java 8 anymore -->
```

pom.xml
```diff
@@ -49,16 +49,23 @@
...
         <javadoc.failOnWarnings>true</javadoc.failOnWarnings>
         <!--  This is required for later correct replacement of surefireArgLine -->
         <!-- see surefire-java8 and surefire-java9+ profiles -->
-        <surefireArgLine/>
+        <surefireArgLine>
+            --add-opens java.base/java.util=ALL-UNNAMED
+            --add-opens java.base/java.lang=ALL-UNNAMED
+        </surefireArgLine>
+        <skip.tests>false</skip.tests>
+        <!-- this will throw an error if we use wrong apis -->
```
toddbaert:
```suggestion
        <!-- this will throw an error if we use APIs not available in the specified version -->
```

toddbaert (approved):
So from my perspective, this:

- allows us to support newer Java versions in dev environments
- allows us to use newer Java versions in CI, in order to use later versions of various plugins
- ensures that we build Java8-compatible bytecode in our release
- ensures that we don't use any >Java8 APIs via the new `maven.compiler` option you added

I think this is a great path forward and a huge DX improvement. Thanks a lot!

.github/workflows/ci.yml
aepfli: removed java 11

chrfwow (approved):

aepfli:
@toddbaert @chrfwow should we move forward with this, as the sdk seems to be working with the new build magic, i am pretty sure we can also move forward with this one. wdyt?

chrfwow:
Changes look good and so does the pipeline, I think we are good to go 🚀

PR #1214, opened 2025-02-12
feat: improve error handling

Nikzz3:
## Improve error handling
This PR improves error handling for 400 and 500 error codes. It returns the message to the user from the response payload.

### How to test
When the providers receives an error (400 or 500), the provider must show the message of the response payload.


beeme1mr (approved):
@thomaspoignant any concerns about this?

toddbaert (approved):
Makes sense to me.

cc @thomaspoignant (but I'm confident this is right so I'll merge in a few days unless I hear otherwise)

PR #1220, opened 2025-02-13
fix: selector not being sent in sync call

aepfli:
some how the selector implementation got lost


chrfwow (approved):

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/RunFileTest.java
```diff
@@ -28,7 +28,7 @@
 @ConfigurationParameter(key = GLUE_PROPERTY_NAME, value = "dev.openfeature.contrib.providers.flagd.e2e.steps")
 @ConfigurationParameter(key = OBJECT_FACTORY_PROPERTY_NAME, value = "io.cucumber.picocontainer.PicoFactory")
 @IncludeTags("file")
-@ExcludeTags({"unixsocket", "targetURI", "reconnect", "customCert"})
+@ExcludeTags({"unixsocket", "targetURI", "reconnect", "customCert", "events"})
```
aepfli, line 31: events seem to not work with file currently, we should investigate or improve the file handler to better work with eg. temporarly unavailable files etc.

toddbaert:
> some how the selector implementation got lost

It was lost in a recent (and very nice-to-have) refactor. It's easy to see why it was lost, but now the e2e will protect us from similar regressions.

toddbaert (approved):

PR #1230, opened 2025-02-19
fix: RPC mode does not honor timeout

aepfli:
During the refactoring of the connection the timeout for unary calls got lost. This adds the timeout back into the mix.

TODO:
- [x] adapt gherkin tests for contextEnrichment to do handle stale separately for file and inprocess - https://github.com/open-feature/flagd-testbed/pull/229

## How to test:

currently one e2e test is failing, which should fail, as this one should actually not work for rpc.

blocked by https://github.com/open-feature/flagd-testbed/pull/229

### DIFF
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/common/GrpcConnector.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/common/GrpcConnector.java
@@ -32,7 +32,7 @@ public class GrpcConnector<T extends AbstractStub<T>, K extends AbstractBlocking
     /**
      * The blocking service stub for making blocking GRPC calls.
      */
-    private final K blockingStub;
+    private final Function<ManagedChannel, K> blockingStubFunction;
 
     /**
      * The GRPC managed channel for managing the underlying GRPC connection.
@@ -58,6 +58,8 @@ public class GrpcConnector<T extends AbstractStub<T>, K extends AbstractBlocking
      * A consumer that handles GRPC service stubs for event stream handling.
      */
     private final Consumer<T> streamObserver;
+
+    private final FlagdOptions options;
 
     /**
      * Indicates whether the connector is currently connected to the GRPC service.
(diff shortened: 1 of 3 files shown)


providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/common/GrpcConnector.java
```diff
@@ -126,7 +129,7 @@ public void initialize() throws Exception {
      * @return the blocking service stub
      */
     public K getResolver() {
-        return blockingStub;
+        return blockingStubFunction.apply(channel).withWaitForReady();
```
aepfli, line 132: if we set the deadline here, (which imho sounds reasonable) it will break our inprocess implementation.

chrfwow: How/why would it break the implementation?

aepfli: we get a strange exception and it breaks our logic. for now i did not want to further investigate, but can be handled with another issue

toddbaert:
@aepfli while we are here, can we add a deadline to the getSyncMetadata call in the in-process as well?

aepfli:
> @aepfli while we are here, can we add a deadline to the getSyncMetadata call in the in-process as well?

I tried to add it, but it breaks our reconnection tests in a strange way. 

If we want to add it, I think it is worth to explore this separately. I also opened an issue with discussion points to start with. Eg. The sync metadata does not have to be part of our loop, we could do this everytime the channel is reestablished. Anyways I think we should discuss this, and define this more precisely, as this could be also the way to go for other languages

See: https://github.com/open-feature/java-sdk-contrib/issues/1231

toddbaert (approved):
This seems like a step in the right direction.

I've made a related proposal here: https://github.com/open-feature/java-sdk-contrib/issues/1231

chrfwow (approved):

PR #1251, opened 2025-02-27
build: Ensure compatibility with newer Java versions by increasing our java build version

aepfli:
this pull request updates our build java version to 21 - to enable the compiler option `proc` which we need for lombok to be working with newer java versions, see details further below

Furthermore we changing junit to use the bom, rather then specifying a version per artifact.

Java changed the way how it processes annotations:

```
As of JDK 23, annotation processing is only run with some explicit configuration of annotation processing or with an explicit request to run annotation processing on the javac command line.
```

this messes with lombok, this compiler configuration will change the default behavior, so it works with java 23

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -49,13 +49,10 @@
     </scm>
 
     <properties>
-        <toolchain.jdk.version>[17,)</toolchain.jdk.version>
-        <junit.jupiter.version>5.11.4</junit.jupiter.version>
+        <toolchain.jdk.version>[21,)</toolchain.jdk.version>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
         <module-name>${groupId}.${artifactId}</module-name>
-        <io.cucumber.version>7.21.1</io.cucumber.version>
-        <org.mockito.version>5.2.0</org.mockito.version>
         <javadoc.failOnWarnings>true</javadoc.failOnWarnings>
         <!--  This is required for later correct replacement of surefireArgLine -->
         <!-- see surefire-java8 and surefire-java9+ profiles -->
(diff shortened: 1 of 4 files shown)


aepfli:
okay this will be more complex than expected, the propery was introduced with java 21 - so we do not have this property within our build environment. We could increase the version of our build environment to 21 - than this would not be an issue anymore

toddbaert:
> okay this will be more complex than expected, the propery was introduced with java 21 - so we do not have this property within our build environment. We could increase the version of our build environment to 21 - than this would not be an issue anymore

That seems fine to me

pom.xml
```diff
@@ -189,7 +189,7 @@
         <dependency>
             <groupId>org.mockito</groupId>
             <artifactId>mockito-inline</artifactId>
-            <version>${org.mockito.version}</version>
+            <version>5.2.0</version>
```
chrfwow: Why can't we use the more recent version?

chrfwow (approved):
Looks good. Only one question regarding the mockito version

pom.xml
aepfli: that is just mockito inline, which differs from mockito version, a minor configuration issue ;)

pom.xml
```diff
@@ -182,35 +199,24 @@
         <dependency>
             <groupId>org.mockito</groupId>
             <artifactId>mockito-junit-jupiter</artifactId>
-            <version>5.15.2</version>
-            <scope>test</scope>
-        </dependency>
-
-        <dependency>
-            <groupId>org.mockito</groupId>
```
aepfli, line 190: inline is now the default and part of mockito-core ;) no need for extra dependency

pom.xml
aepfli: even better, this is now part of mockito core

chrfwow (approved):

beeme1mr (approved):

toddbaert (approved):

PR #1259, opened 2025-03-04
feat(flagd): Support supplying providerID for in-process resolver as an option

cupofcat:
## This PR
- When the user is using in-process flagd, allows them to supply a providerID string to be in gRPC connections
- Adds tests and documentation for providerID

### DIFF
--- a/providers/flagd/README.md
+++ b/providers/flagd/README.md
@@ -116,7 +116,7 @@ Given below are the supported configurations:
 | streamDeadlineMs      | FLAGD_STREAM_DEADLINE_MS       | int                      | 600000    | rpc & in-process        |
 | keepAliveTime         | FLAGD_KEEP_ALIVE_TIME_MS       | long                     | 0         | rpc & in-process        |
 | selector              | FLAGD_SOURCE_SELECTOR          | String                   | null      | in-process              |
-| providerId            | FLAGD_SOURCE_PROVIDER_ID       | String                   | null      | in-process              |
+| providerID            | FLAGD_SOURCE_PROVIDER_ID       | String                   | null      | in-process              |
 | cache                 | FLAGD_CACHE                    | String - lru, disabled   | lru       | rpc                     |
 | maxCacheSize          | FLAGD_MAX_CACHE_SIZE           | int                      | 1000      | rpc                     |
 | maxEventStreamRetries | FLAGD_MAX_EVENT_STREAM_RETRIES | int                      | 5         | rpc                     |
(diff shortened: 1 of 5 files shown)


providers/flagd/README.md
```diff
@@ -116,6 +116,7 @@ Given below are the supported configurations:
 | streamDeadlineMs      | FLAGD_STREAM_DEADLINE_MS       | int                      | 600000    | rpc & in-process        |
 | keepAliveTime         | FLAGD_KEEP_ALIVE_TIME_MS       | long                     | 0         | rpc & in-process        |
 | selector              | FLAGD_SOURCE_SELECTOR          | String                   | null      | in-process              |
+| providerId            | FLAGD_SOURCE_PROVIDER_ID       | String                   | null      | in-process              |
```
beeme1mr, line 120: Should we use `providerId` or `providerID`? In Go, the later was used and that's what's defined [here](https://github.com/open-feature/flagd/pull/1567).

aepfli: depending on the decision, we might want to add it [here](https://github.com/open-feature/flagd-testbed/blob/main/gherkin/config.feature) for easier testing throughout out all of our sdks

cupofcat: Oops, fixed to be consistent with Go.

beeme1mr (approved):

providers/flagd/README.md
```diff
@@ -116,6 +116,7 @@ Given below are the supported configurations:
 | streamDeadlineMs      | FLAGD_STREAM_DEADLINE_MS       | int                      | 600000    | rpc & in-process        |
 | keepAliveTime         | FLAGD_KEEP_ALIVE_TIME_MS       | long                     | 0         | rpc & in-process        |
 | selector              | FLAGD_SOURCE_SELECTOR          | String                   | null      | in-process              |
+| providerID            | FLAGD_SOURCE_PROVIDER_ID       | String                   | null      | in-process              |
```
```suggestion
| providerId            | FLAGD_SOURCE_PROVIDER_ID       | String                   | null      | in-process              |
```

I think for consistency this should be fully camel case.

cupofcat: Haha, I think I'm getting conflicting feedback here :) I started with camelcase and then @beeme1mr suggested capitalized ID (to be consistent with Go) and now you are suggesting camelcase again. Personally, I don't have a strong opinion either way.

chrfwow (approved):

providers/flagd/README.md
toddbaert: We've use camel case for "id" elsewhere in general (like flagSetId). I think I'd prefer that @beeme1mr

providers/flagd/README.md
toddbaert: Sorry! https://github.com/open-feature/java-sdk-contrib/pull/1259#discussion_r1981520856

providers/flagd/README.md
beeme1mr: I'm fine with that. I honestly prefer camel case here but I wanted to call out the inconstancy. As long as the environment variable is the same, it's probably better to follow the languages best practices than force a particular style.

toddbaert (approved):

PR #1261, opened 2025-03-04
fix: broken reconnect on some HTTP2 error frames

toddbaert:
This PR fixes an issue where some HTTP2 error frames (for example `RST_STREAM` frames) cause our streams to die and not recover. The `RST_STREAM` in particular is sent by envoy/istio when upstream connections are lost, though the connection stays open on our end (https://github.com/envoyproxy/envoy/issues/30149, https://github.com/grpc/grpc-go/issues/8041). We are required to restart the stream in this case by reacting to the `onCompleted` stream handler, which we never did in the `eventStream` and we recently stopped doing in the `syncStream` (due to a regression).

Specifically, the PR:

- handles `onError` and `onComplete` stream messages by reconnecting the sync (in-process) and event (RPC) streams (this was the main issue)
- adds specific unit tests for the above
- does some renames and refactors
  - anything to do with the sync stream is now called "sync" instead of "grpc" to reduce ambiguity
  - anything to do with RPC mode is now called "rpc" or prefixed with that
  - the RPC `eventStream` now also uses a blocking queue to consume messages, just like the `syncStream`
  
We don't have e2e tests for this issue because we'd need some more feature in the test-harness to do that, but I will create an issue for it.


providers/flagd/pom.xml
```diff
@@ -69,10 +70,22 @@
...
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-databind</artifactId>
-            <version>2.18.3</version>
+            <version>${com.fasterxml.jackson.version}</version>
+        </dependency>
+
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-annotations</artifactId>
+            <version>${com.fasterxml.jackson.version}</version>
```
toddbaert, line 88: unrelated (we were missing some jackson deps that are included in many projects but should be here)

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
```diff
@@ -82,7 +82,7 @@ public FlagdProvider(final FlagdOptions options) {
                 this.flagResolver = new InProcessResolver(options, this::onProviderEvent);
                 break;
             case Config.RESOLVER_RPC:
-                this.flagResolver = new GrpcResolver(
+                this.flagResolver = new RpcResolver(
```
toddbaert, line 85: "RPC" is a better prefix for our RPC mode stuff.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncResponseModel.java
```diff
@@ -1,27 +1,27 @@
...
 
 @Getter
-class GrpcResponseModel {
+class SyncResponseModel {
     private final SyncFlagsResponse syncFlagsResponse;
     private final Throwable error;
     private final boolean complete;
 
-    public GrpcResponseModel(final Throwable error) {
+    public SyncResponseModel(final Throwable error) {
```
toddbaert: All of our "Sync" related stuff is now mostly prefixed with "Sync" instead of "Grpc" which is pretty ambiguous since both modes use grpc.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/grpc/RpcResolver.java
```diff
@@ -61,40 +73,53 @@ public final class GrpcResolver implements Resolver {
...
 
     /**
      * Initialize Grpc resolver.
      */
     public void init() throws Exception {
         this.connector.initialize();
+
+        Thread listener = new Thread(() -> {
+            try {
+                observeEventStream(shutdown);
```
toddbaert: This is a big refactor, because we also had the [same problem](https://github.com/open-feature/java-sdk-contrib/pull/1261/files#r1980448922) with the RPC mode (we were ignoring stream errors and not restarting).

I'm using a blocking queue similar to how we do in the Sync resolver now to handle each even in a simple [while loop](https://github.com/open-feature/java-sdk-contrib/pull/1261/files#diff-bc31296c0793bdd27a0bb88be2ae2914765eb3f9f11db98d7cb3b476be99854dR303), and restart the stream on errors.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamConnector.java
```diff
@@ -109,11 +98,12 @@ void observeEventStream(final BlockingQueue<QueuePayload> writeTo, final AtomicB
             log.debug("Initializing sync stream request");
             final GetMetadataRequest.Builder metadataRequest = GetMetadataRequest.newBuilder();
             GetMetadataResponse metadataResponse = GetMetadataResponse.getDefaultInstance();
+            restart();
```
toddbaert: The main issue this PR fixes is that there was no call to restart the stream in this while loop, and error messages are fatal (meaning the stream needs to be restored).

This restart does that, instead of only starting the stream once it starts it whenever we break out of the inner loop here.

providers/flagd/pom.xml
```diff
@@ -11,12 +11,13 @@
     </parent>
     <groupId>dev.openfeature.contrib.providers</groupId>
     <artifactId>flagd</artifactId>
-    <version>0.11.4</version> <!--x-release-please-version -->
+    <version>0.11.23-local</version> <!--x-release-please-version -->
```
beeme1mr: Revert before merging.

toddbaert: Will do, I just want to do some final manual tests

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/common/ChannelConnector.java
```diff
@@ -120,16 +106,20 @@ public GrpcConnector(
...
      * Returns the blocking service stub for making blocking GRPC calls.
      *
      * @return the blocking service stub
      */
-    public K getResolver() {
-        return blockingStubFunction.apply(channel).withWaitForReady();
+    public K getBlockingStub() {
+        K stub = blockingStubFunction.apply(channel).withWaitForReady();
+        if (this.deadline > 0) {
+            stub = stub.withDeadlineAfter(this.deadline, TimeUnit.MILLISECONDS);
```
aepfli, line 100: I also thought about unfying this here, in my fix for the timeouts for RPC - but somehow this messed with the inprocess resolver connection, and the e2e tests have not been successful - still no clue why, but if we encounter this again, i suggest to maybe tackle this unification separately

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/common/ChannelConnector.java
```diff
@@ -167,19 +156,40 @@ private synchronized void onConnectionLost() {
...
+     */
+    public static void monitorChannelState(
+            ConnectivityState expectedState,
+            ManagedChannel channel,
+            Runnable onConnectionReady,
+            Runnable onConnectionLost) {
+        channel.notifyWhenStateChanged(expectedState, () -> {
+            ConnectivityState currentState = channel.getState(true);
+            log.debug("Channel state changed to: {}", currentState);
+            if (currentState == ConnectivityState.READY) {
```
aepfli: currently we do have `waitForReady` which wlll force communication on an idle channel. But should we maybe also add here the `IDLE` state? - currently i think this path is not used anyways

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamConnector.java
```diff
@@ -161,4 +153,21 @@ void observeEventStream(final BlockingQueue<QueuePayload> writeTo, final AtomicB
...
         log.info("Shutdown invoked, exiting event stream listener");
     }
+
+    private void restart() {
+        log.debug("starting our new restart method");
+        FlagSyncServiceStub stub =
+                FlagSyncServiceGrpc.newStub(this.grpcConnector.getChannel()).withWaitForReady();
+        if (streamDeadline > 0) {
+            stub = stub.withDeadlineAfter(streamDeadline, TimeUnit.MILLISECONDS);
+        }
```
aepfli: as those settings are the same, we might want to extract this stub creation into a util function or something similar. We will most likely improve something for one GRPC Stream connection, but forget about the other one, afterwards. Having this unified might be a good idea.

aepfli (commented):
i know this is still work in progress, but i added some thoughts - if i find the time i am more than willing to support here :) this is a great improvement, and i think normalizing our code base, reducing complexity has a big benefit. .... in the longrun maybe our rpc mode will also use a construct of flagstore and connector (not part of this) so we have one unified approach to this :)

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/common/ChannelConnector.java
```diff
@@ -57,7 +53,7 @@ public class GrpcConnector<T extends AbstractStub<T>, K extends AbstractBlocking
     /**
      * A consumer that handles GRPC service stubs for event stream handling.
      */
-    private final Consumer<T> streamObserver;
+    // private final Consumer<T> streamObserver;
```
chrfwow: Don't forget to remove this

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/common/ChannelConnector.java
```diff
@@ -167,19 +156,40 @@ private synchronized void onConnectionLost() {
...
-            if (streamDeadlineMs > 0) {
-                localServiceStub = localServiceStub.withDeadlineAfter(this.streamDeadlineMs, TimeUnit.MILLISECONDS);
+     * Monitors the state of a gRPC channel and triggers the specified callbacks based on state changes.
+     *
+     * @param expectedState     the initial state to monitor.
+     * @param channel           the ManagedChannel to monitor.
+     * @param onConnectionReady callback invoked when the channel transitions to a READY state.
+     * @param onConnectionLost  callback invoked when the channel transitions to a FAILURE or SHUTDOWN state.
+     */
+    public static void monitorChannelState(
```
chrfwow: Is this method used somewhere else? If not, I suggest making this method private

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/common/ChannelConnector.java
```diff
@@ -167,19 +156,40 @@ private synchronized void onConnectionLost() {
...
+            ConnectivityState expectedState,
+            ManagedChannel channel,
+            Runnable onConnectionReady,
+            Runnable onConnectionLost) {
+        channel.notifyWhenStateChanged(expectedState, () -> {
+            ConnectivityState currentState = channel.getState(true);
+            log.debug("Channel state changed to: {}", currentState);
+            if (currentState == ConnectivityState.READY) {
+                if (onConnectionReady != null) {
```
chrfwow: `onConnectionReady` is currently basically only a setter for `connected`. I suggest we inline setting it to true and making this method non static, which will also get rid of the null check and be more efficient as we get rid of the runnable. We might consider the same for `onConnectionLost`.

If we do that, don't forget to synchronize on `this` or make `connected` volatile

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamConnector.java
```diff
@@ -109,11 +98,12 @@ void observeEventStream(final BlockingQueue<QueuePayload> writeTo, final AtomicB
...
             GetMetadataResponse metadataResponse = GetMetadataResponse.getDefaultInstance();
+            restart();
 
             try (CancellableContext context = Context.current().withCancellation()) {
 
                 try {
-                    metadataResponse = grpcConnector.getResolver().getMetadata(metadataRequest.build());
+                    metadataResponse = grpcConnector.getBlockingStub().getMetadata(metadataRequest.build());
                 } catch (Exception e) {
                     // the chances this call fails but the syncRequest does not are slim
```
chrfwow: Why don't we handle this exception here instead of in the nested while loop, together with another error?

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamConnector.java
```diff
@@ -161,4 +153,21 @@ void observeEventStream(final BlockingQueue<QueuePayload> writeTo, final AtomicB
 
         log.info("Shutdown invoked, exiting event stream listener");
     }
+
+    private void restart() {
+        log.debug("starting our new restart method");
```
chrfwow: Don't forget to remove this log line and the one at the end of the method

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/common/ChannelConnector.java
```diff
@@ -71,46 +67,36 @@ public class GrpcConnector<T extends AbstractStub<T>, K extends AbstractBlocking
      * Constructs a new {@code GrpcConnector} instance with the specified options and parameters.
```
guidobrei: Class renamed

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/common/ChannelConnector.java
```diff
@@ -71,46 +67,36 @@ public class GrpcConnector<T extends AbstractStub<T>, K extends AbstractBlocking
...
         this.blockingStubFunction = blockingStub;
         this.deadline = options.getDeadline();
         this.streamDeadlineMs = options.getStreamDeadlineMs();
         this.onConnectionEvent = onConnectionEvent;
-        this.streamObserver = eventStreamObserver;
         this.options = options;
     }
 
     /**
      * Constructs a {@code GrpcConnector} instance for testing purposes.
```
guidobrei: Class renamed

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamConnector.java
```diff
@@ -27,48 +30,34 @@
 @SuppressFBWarnings(
         value = {"PREDICTABLE_RANDOM", "EI_EXPOSE_REP"},
         justification = "Random is used to generate a variation & flag configurations require exposing")
-public class GrpcStreamConnector implements Connector {
+public class SyncStreamConnector implements Connector {
```
guidobrei: Any chance that some outsider is using this public class?

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/rpc/RpcResolver.java
```diff
@@ -61,40 +73,53 @@ public final class GrpcResolver implements Resolver {
...
+                Thread.currentThread().interrupt();
+            }
+        });
+
+        listener.setDaemon(true);
+        listener.start();
+    }
+
+    private void restartStream() {
+        log.debug("starting our new restart method");
```
guidobrei: Dev log line... to be removed

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/rpc/RpcResolver.java
```diff
@@ -61,40 +73,53 @@ public final class GrpcResolver implements Resolver {
...
+        ServiceStub stub = ServiceGrpc.newStub(this.connector.getChannel()).withWaitForReady();
+        if (options.getStreamDeadlineMs() > 0) {
+            stub = stub.withDeadlineAfter(options.getStreamDeadlineMs(), TimeUnit.MILLISECONDS);
+        }
+
+        stub.eventStream(EventStreamRequest.getDefaultInstance(), new EventStreamObserver(streamReceiver));
+        log.debug("ended our new restart method");
```
guidobrei: Dev log line... to be removed

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/rpc/RpcResolver.java
```diff
@@ -269,4 +294,89 @@ private OpenFeatureError mapError(Exception e) {
...
+
+            try (CancellableContext context = Context.current().withCancellation()) {
+
+                while (!shutdown.get()) {
+                    final EventStreamResponseModel taken = streamReceiver.take();
+                    if (taken.isComplete()) {
+                        log.info("Sync stream completed");
+                        // The stream is complete, this isn't really an error, but we should try to
+                        // reconnect
+                        // TODO: onError
```
guidobrei: Is this a previous todo or something for this PR?

guidobrei (commented):
I've noticed some Javadoc comments that are affected by class renaming of `Grpc*` to `Rpc*`. I did not mark all of them.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/common/ChannelConnector.java
aepfli: I improved the class, without reading the comment, but I think I nailed what you wanted us to do ;)

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamConnector.java
toddbaert: If they are they will be punished with a compilation issue in the next release :imp:

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/rpc/RpcResolver.java
toddbaert: This is from this current work, just a reminder for me. I will clean it before I remove draft.

toddbaert: Removed

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamConnector.java
toddbaert: Good point, I've done this. I think this is left over from when the sync metadata data was optional

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamConnector.java
toddbaert: These methods are a bit too different IMO to commonize, they call different stream methods etc. I've moved the stub creation to the constructor though, there's no reason to create one every time.

providers/flagd/pom.xml
toddbaert: removed.

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/resolver/rpc/RpcResolverTest.java
```diff
@@ -0,0 +1,130 @@
...
+class RpcResolverTest {
+    private ChannelConnector<ServiceStub, ServiceBlockingStub> mockConnector;
+    private ServiceBlockingStub blockingStub;
+    private ServiceStub stub;
+    private EventStreamObserver observer;
+    private Cache cache;
+    private Consumer<FlagdProviderEvent> consumer;
+    private CountDownLatch latch; // used to wait for observer to be initialized
+
+    @BeforeEach
```
toddbaert, line 37: These tests the RPC stream behavior.

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamQueueSourceTest.java
```diff
@@ -0,0 +1,121 @@
...
+import org.mockito.stubbing.Answer;
+
+class SyncStreamQueueSourceTest {
+    private ChannelConnector<FlagSyncServiceStub, FlagSyncServiceBlockingStub> mockConnector;
+    private FlagSyncServiceBlockingStub blockingStub;
+    private FlagSyncServiceStub stub;
+    private SyncStreamObserver observer;
+    private CountDownLatch latch; // used to wait for observer to be initialized
+
+    @BeforeEach
```
toddbaert, line 40: These tests the sync stream behavior.

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamQueueSourceTest.java
```diff
@@ -0,0 +1,121 @@
...
+        // fire onError event
+        observer.onError(new Exception("fake exception"));
+
+        // should enqueue error payload
+        BlockingQueue<QueuePayload> streamQueue = connector.getStreamQueue();
+        QueuePayload payload = streamQueue.poll(1000, TimeUnit.MILLISECONDS);
+        assertNotNull(payload);
+        assertEquals(QueuePayloadType.ERROR, payload.getType());
+        // should have restarted the stream (2 calls)
+        verify(stub, times(2)).syncFlags(any(), any());
```
toddbaert, line 103: Checking the `times(2)` here is what specifically proves we reconnect.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamQueueSource.java
```diff
@@ -0,0 +1,174 @@
...
+            localStub = localStub.withDeadlineAfter(streamDeadline, TimeUnit.MILLISECONDS);
+        }
+
+        final SyncFlagsRequest.Builder syncRequest = SyncFlagsRequest.newBuilder();
+        if (this.selector != null) {
+            syncRequest.setSelector(this.selector);
+        }
+
+        if (this.providerId != null) {
+            syncRequest.setProviderId(this.providerId);
```
toddbaert, line 173: @cupofcat FYI some of your recent changes moved here. I didn't change your tests and I'm confident everything still works, I just didn't want you to be confused if you noticed your changes were immediate moved :sweat_smile:

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/rpc/RpcResolver.java
```diff
@@ -35,66 +37,96 @@
...
             final FlagdOptions options, final Cache cache, final Consumer<FlagdProviderEvent> onProviderEvent) {
         this.cache = cache;
         this.strategy = ResolveFactory.getStrategy(options);
         this.options = options;
-        this.connector = new GrpcConnector<>(
-                options,
-                ServiceGrpc::newStub,
-                ServiceGrpc::newBlockingStub,
-                onProviderEvent,
-                stub -> stub.eventStream(
```
toddbaert, line 74: This was moved into a private method so it can be called within our stream loop (same as `sync`).

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/grpc/GrpcStreamConnector.java
```diff
@@ -1,170 +0,0 @@
...
-    }
-
-    /** Contains blocking calls, to be used concurrently. */
-    void observeEventStream(final BlockingQueue<QueuePayload> writeTo, final AtomicBoolean shutdown, final int deadline)
-            throws InterruptedException {
-
-        log.info("Initializing sync stream observer");
-
-        while (!shutdown.get()) {
-            writeTo.clear();
```
toddbaert, line 112: I've removed this queue clear. It actually caused a race condition because it's possible that the queue was cleared after a stream restart before it was consumed-from, leading to a lost event.

I can't see why it's important anyway - there's no reason to remove queued events from a previous stream if we have them... just let them be dequeued as normal.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/rpc/RpcResolver.java
```diff
@@ -269,4 +301,93 @@ private OpenFeatureError mapError(Exception e) {
         }
         return new GeneralError(e.getMessage());
     }
+
+    private void restartStream() {
+        ServiceStub localStub = stub; // don't mutate the stub
+        if (options.getStreamDeadlineMs() > 0) {
+            localStub = stub.withDeadlineAfter(options.getStreamDeadlineMs(), TimeUnit.MILLISECONDS);
```
aepfli:
```suggestion
            localStub = localStub.withDeadlineAfter(options.getStreamDeadlineMs(), TimeUnit.MILLISECONDS);
```

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncResponseModel.java
```diff
@@ -1,27 +1,27 @@
-package dev.openfeature.contrib.providers.flagd.resolver.process.storage.connector.grpc;
+package dev.openfeature.contrib.providers.flagd.resolver.process.storage.connector.sync;
 
 import dev.openfeature.flagd.grpc.sync.Sync.SyncFlagsResponse;
 import lombok.Getter;
 
 @Getter
-class GrpcResponseModel {
+class SyncResponseModel {
```
aepfli: [suggestion] To reduce duplication, we could make this class generic as it shares the same behavior as EventStreamResponseModel, just with different types

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamObserver.java
```diff
@@ -1,35 +1,35 @@
...
 
 import dev.openfeature.flagd.grpc.sync.Sync.SyncFlagsResponse;
 import io.grpc.stub.StreamObserver;
 import java.util.concurrent.BlockingQueue;
 import lombok.extern.slf4j.Slf4j;
 
 @Slf4j
-class GrpcStreamHandler implements StreamObserver<SyncFlagsResponse> {
-    private final BlockingQueue<GrpcResponseModel> blockingQueue;
+class SyncStreamObserver implements StreamObserver<SyncFlagsResponse> {
```
aepfli: [suggestion] To reduce duplication, we could make this class generic as it shares the same behaviour as EventStreamObserver, just with different types

aepfli (approved):
Thank you. I feel like we are getting closer and closer to an easy to use codebase, in which we share the same approaches in all the providers. Changing to a queue for also RPC normalizes things, and improves the maintainability of our codebase in the long run

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamQueueSource.java
```diff
@@ -0,0 +1,174 @@
...
+            GetMetadataResponse metadataResponse = GetMetadataResponse.getDefaultInstance();
+            restart();
+
+            try (CancellableContext context = Context.current().withCancellation()) {
+
+                try {
+                    metadataResponse = channelConnector.getBlockingStub().getMetadata(metadataRequest.build());
+                } catch (Exception metaEx) {
+                    log.error("Metadata exception: {}", metaEx.getMessage(), metaEx);
+                    context.cancel(metaEx);
```
chrfwow, line 125: Is it ok to proceed with a canceled context into the inner while loop, or should we try to restart (which might lead to problems because we might never get into the inner while loop just because we cannot query metadata)?

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamQueueSource.java
```diff
@@ -0,0 +1,174 @@
...
+
+        // outer loop for re-issuing the stream request
+        while (!shutdown.get()) {
+
+            log.debug("Initializing sync stream request");
+            final GetMetadataRequest.Builder metadataRequest = GetMetadataRequest.newBuilder();
+            GetMetadataResponse metadataResponse = GetMetadataResponse.getDefaultInstance();
+            restart();
+
+            try (CancellableContext context = Context.current().withCancellation()) {
```
chrfwow, line 117: `context` is not used in the inner while loop. Does it make sense to close this try block before the while loop?

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamQueueSource.java
```diff
@@ -0,0 +1,174 @@
...
+
+    /**
+     * Creates a new SyncStreamQueueSource responsible for observing the event stream.
+     */
+    public SyncStreamQueueSource(final FlagdOptions options, Consumer<FlagdProviderEvent> onConnectionEvent) {
+        streamDeadline = options.getStreamDeadlineMs();
+        selector = options.getSelector();
+        providerId = options.getProviderId();
+        channelConnector = new ChannelConnector<>(options, FlagSyncServiceGrpc::newBlockingStub, onConnectionEvent);
+        // TODO: should we do this in init?
```
chrfwow: I think this is fine

toddbaert: Yes you are right. I forgot to remove this.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamQueueSource.java
toddbaert: This shouldn't be different from the previous behavior, as `context.cancel()` will notify the stream listener and put an error event on it, so we will break out of the inner loop after 1 iteration.

However, this would mean that we can get in a fairly tight loop starting the stream over and over if the metadata errors over and over (this is not new). I'm not sure what can be done to improve this since I really do not want to proceed without metadata in case an admin has injected context attrs and there's rules involving them, and I don't want to give up forever on one error either. We could perhaps add a sleep after the metadata error just to make sure the loop isn't super tight... but I don't want to build our own backoff here for this edge case.

I really dislike the fact the metadata is a separate RPC. We need to improve that.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamQueueSource.java
toddbaert: Actually this is a very good catch. The restart needs to be in the context-try-with-resource block so we can cancel the stream appropriately outside like this the cancel will have no impact. I've fixed that (but we need to keep the context for the duration of the stream.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamQueueSource.java
toddbaert: Removed.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamQueueSource.java
toddbaert: cc @cupofcat

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamObserver.java
toddbaert: :100: done

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncResponseModel.java
toddbaert: :100: done

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamConnector.java
toddbaert: Not much we can do about this.

PR #1268, opened 2025-03-07
chore: adjust noisy log lines

toddbaert:
Just tweaking some log lines.

My opinion is:

- anything log lines that are "normal" (expected at runtime) should be `debug`
  - stream ending (this is normal with deadlines)
  - change notifications
- startup/shutdown messages are OK at `info`
- unexpected errors (like disconnections) should be `error`
- "weird things" that should never happen, but aren't catastrophic should be `warn`.

### DIFF
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
@@ -193,7 +193,7 @@ public class FlagdProvider extends EventProvider {
 
     @SuppressWarnings("checkstyle:fallthrough")
     private void onProviderEvent(FlagdProviderEvent flagdProviderEvent) {
-        log.info("FlagdProviderEvent event {} ", flagdProviderEvent.getEvent());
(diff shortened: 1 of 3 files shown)


beeme1mr (approved):

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
```diff
@@ -256,7 +256,7 @@ private void onReady() {
     }
 
     private void onError() {
-        log.info("Connection lost. Emit STALE event...");
+        log.debug("Stream error. Emitting STALE event and scheduling ERROR event...");
```
aepfli: Could we merge this with the line below?

aepfli (approved):

chrfwow (approved):

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
toddbaert:
```java
        log.debug(
                "Stream error. Emitting STALE, scheduling ERROR, and waiting {}s for connection to become available.",
                gracePeriod);
```

PR #1273, opened 2025-03-11
fix: transient error log-spam, add retry policy

toddbaert:
Fixes an issue observed when intervening proxies are used (like Istio), where the proxy's upstream is disconnected, resulting in immediate error responses from the proxy (which we still have a healthy TCP connection to) which causes a tight loop in our retry logic for a brief time until the upstream is available. 

This PR adds a retry policy to retry such problematic requests, which also slows down our retry loop. In general this will make the provider more robust to transient errors in streams and unary calls. I will be adding something similar to the flagd spec.

Other minor refactors as well.

There is one possible enhancement we could consider as a separate issue, mentioned [here](https://github.com/open-feature/java-sdk-contrib/pull/1273/files#r1990354285).

### DIFF
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/common/ChannelBuilder.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/common/ChannelBuilder.java
@@ -5,6 +5,7 @@ import dev.openfeature.contrib.providers.flagd.resolver.common.nameresolvers.Env
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import io.grpc.ManagedChannel;
 import io.grpc.NameResolverRegistry;
+import io.grpc.Status.Code;
 import io.grpc.netty.GrpcSslContexts;
 import io.grpc.netty.NettyChannelBuilder;
 import io.netty.channel.epoll.Epoll;
@@ -15,11 +16,75 @@ import io.netty.handler.ssl.SslContextBuilder;
 import java.io.File;
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.concurrent.TimeUnit;
 import javax.net.ssl.SSLException;
 
 /** gRPC channel builder helper. */
 public class ChannelBuilder {
+
+    static final int MAX_RETRY_ATTEMPTS = 3;
+
+    /**
+     * Controls retry (not-reconnection) policy for failed RPCs.
+     */
+    @SuppressWarnings({"unchecked", "rawtypes"})
+    static final Map<String, ?> SERVICE_CONFIG_WITH_RETRY = new HashMap() {
+        {
+            put("methodConfig", Arrays.asList(new HashMap() {
+                {
+                    put(
+                            "name",
+                            Arrays.asList(
+                                    new HashMap() {
+                                        {
+                                            put("service", "flagd.sync.v1.FlagSyncService");
+                                        }
+                                    },
+                                    new HashMap() {
+                                        {
+                                            put("service", "flagd.evaluation.v1.Service");
+                                        }
+                                    }));
+                    put("retryPolicy", new HashMap() {
+                        {
+                            // 1 + 2 + 4
+                            put("maxAttempts", (double) MAX_RETRY_ATTEMPTS);
+                            put("initialBackoff", "1s");
+                            put("maxBackoff", "5s");
+                            put("backoffMultiplier", 2.0);
+                            // status codes to retry on:
+                            put(
+                                    "retryableStatusCodes",
+                                    Arrays.asList(
+                                            /*
+                                             * All codes are retryable except OK or CANCELLED, since anything
+                                             * not listed here causes a very tight loop of retries.
+                                             * CANCELLED is not retryable because it is a client-side termination.
+                                             */
+                                            Code.UNKNOWN.toString(),
+                                            Code.INVALID_ARGUMENT.toString(),
+                                            Code.DEADLINE_EXCEEDED.toString(),
+                                            Code.NOT_FOUND.toString(),
+                                            Code.ALREADY_EXISTS.toString(),
+                                            Code.PERMISSION_DENIED.toString(),
+                                            Code.RESOURCE_EXHAUSTED.toString(),
+                                            Code.FAILED_PRECONDITION.toString(),
+                                            Code.ABORTED.toString(),
+                                            Code.OUT_OF_RANGE.toString(),
+                                            Code.UNIMPLEMENTED.toString(),
+                                            Code.INTERNAL.toString(),
+                                            Code.UNAVAILABLE.toString(),
+                                            Code.DATA_LOSS.toString(),
+                                            Code.UNAUTHENTICATED.toString()));
+                        }
+                    });
+                }
+            }));
+        }
+    };
 
     private ChannelBuilder() {}
 
@@ -45,6 +110,9 @@ public class ChannelBuilder {
                     .eventLoopGroup(new EpollEventLoopGroup())
                     .channelType(EpollDomainSocketChannel.class)
                     .usePlaintext()
+                    .defaultServiceConfig(SERVICE_CONFIG_WITH_RETRY)
+                    .maxRetryAttempts(MAX_RETRY_ATTEMPTS)
+                    .enableRetry()
                     .build();
         }
 
@@ -89,7 +157,10 @@ public class ChannelBuilder {
                 builder.intercept(new FlagdGrpcInterceptor(options.getOpenTelemetry()));
             }
 
-            return builder.build();
+            return builder.defaultServiceConfig(SERVICE_CONFIG_WITH_RETRY)
+                    .maxRetryAttempts(MAX_RETRY_ATTEMPTS)
+                    .enableRetry()
+                    .build();
         } catch (SSLException ssle) {
             SslConfigException sslConfigException = new SslConfigException("Error with SSL configuration.");
             sslConfigException.initCause(ssle);
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/common/ChannelBuilder.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/common/ChannelBuilder.java
@@ -60,13 +60,14 @@ public class ChannelBuilder {
                                     "retryableStatusCodes",
                                     Arrays.asList(
                                             /*
-                                             * All codes are retryable except OK or CANCELLED, since anything
-                                             * not listed here causes a very tight loop of retries.
+                                             * All codes are retryable except OK, CANCELLED and DEADLINE_EXCEEDED since
+                                             * any others not listed here cause a very tight loop of retries.
                                              * CANCELLED is not retryable because it is a client-side termination.
+                                             * DEADLINE_EXCEEDED is typically a result of a client specified deadline,
+                                             * and definitionally should not result in a tight loop (it's a timeout).
                                              */
                                             Code.UNKNOWN.toString(),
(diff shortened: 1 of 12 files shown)


providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/common/ChannelBuilder.java
```diff
@@ -15,12 +16,64 @@
...
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.concurrent.TimeUnit;
 import javax.net.ssl.SSLException;
 
 /** gRPC channel builder helper. */
 public class ChannelBuilder {
 
+    static final int MAX_RETRY_ATTEMPTS = 3;
```
toddbaert, line 28: Let me know your thoughts on this policy. I will add it to the flagd spec once merged, since this is available in most gRPC implementations.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/common/ChannelConnector.java
```diff
@@ -89,19 +74,6 @@ public void initialize() throws Exception {
...
-     * @return the blocking service stub
-     */
-    public K getBlockingStub() {
-        K stub = blockingStubFunction.apply(channel).withWaitForReady();
-        if (this.deadline > 0) {
-            stub = stub.withDeadlineAfter(this.deadline, TimeUnit.MILLISECONDS);
-        }
-        return stub;
-    }
-
```
toddbaert, line 104: Refactor: as pointed out by @aepfli , this has no business being in this class, which other than this, only handles channel state. I've moved this stubs into their respective correct places which already have the async stubs.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/common/ChannelBuilder.java
```diff
@@ -15,12 +16,64 @@
...
+
+    /**
+     * Controls retry (not-reconnection) policy for failed RPCs.
+     */
+    @SuppressWarnings({"unchecked", "rawtypes"})
+    static final Map<String, ?> SERVICE_CONFIG_WITH_RETRY = new HashMap() {
```
toddbaert, line 32: Let me know your thoughts on this policy. I will add it to the flagd spec once merged, since this is available in most gRPC implementations.

Note that this is used by both RPC and in-process impls, and it impacts both streams and unary RPCs (I tested this).

see: https://grpc.io/docs/guides/retry/

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamQueueSource.java
```diff
@@ -124,15 +132,21 @@ private void observeSyncStream() throws InterruptedException {
...
                 if (!syncMetadataDisabled) {
                     try {
-                        metadataResponse = channelConnector.getBlockingStub().getMetadata(metadataRequest.build());
+                        FlagSyncServiceBlockingStub localStub = blockingStub;
+
+                        if (deadline > 0) {
+                            localStub = localStub.withDeadlineAfter(deadline, TimeUnit.MILLISECONDS);
+                        }
+
+                        metadataResponse = localStub.getMetadata(metadataRequest.build());
```
toddbaert, line 142: https://github.com/open-feature/java-sdk-contrib/pull/1273/files#r1989971943

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/rpc/RpcResolver.java
```diff
@@ -172,11 +177,17 @@ public ProviderEvaluation<Integer> integerEvaluation(String key, Integer default
...
-    private ServiceGrpc.ServiceBlockingStub getResolver() {
-        return connector.getBlockingStub().withDeadlineAfter(options.getDeadline(), TimeUnit.MILLISECONDS);
+    private ServiceGrpc.ServiceBlockingStub getBlockingStub() {
+        ServiceBlockingStub localStub = blockingStub;
+
+        if (options.getDeadline() > 0) {
+            localStub = localStub.withDeadlineAfter(options.getDeadline(), TimeUnit.MILLISECONDS);
+        }
+
+        return localStub;
```
toddbaert, line 190: https://github.com/open-feature/java-sdk-contrib/pull/1273/files#r1989971943

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/FlagdProviderTest.java
```diff
@@ -230,9 +229,9 @@ void zero_value() {
                 .thenReturn(objectResponse);
 
         ChannelConnector grpc = mock(ChannelConnector.class);
-        when(grpc.getBlockingStub()).thenReturn(serviceBlockingStubMock);
+        // when(grpc.getBlockingStub()).thenReturn(serviceBlockingStubMock);
```
toddbaert:
```suggestion
```

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/common/ChannelBuilder.java
```diff
@@ -15,12 +16,76 @@
...
+                                            Code.PERMISSION_DENIED.toString(),
+                                            Code.RESOURCE_EXHAUSTED.toString(),
+                                            Code.FAILED_PRECONDITION.toString(),
+                                            Code.ABORTED.toString(),
+                                            Code.OUT_OF_RANGE.toString(),
+                                            Code.UNIMPLEMENTED.toString(),
+                                            Code.INTERNAL.toString(),
+                                            Code.UNAVAILABLE.toString(),
+                                            Code.DATA_LOSS.toString(),
+                                            Code.UNAUTHENTICATED.toString()));
```
toddbaert: As the comment says:

```java
    /*
    * All codes are retryable except OK or CANCELLED, since anything
    * not listed here causes a very tight loop of retries.
    * CANCELLED is not retryable because it is a client-side termination.
    */
```

We always want our provider to reconnect, even if for some reason a service temporarily responds NOT_FOUND or NOT_IMPLEMENTED, etc. Adding all error codes here not only makes things robust, but it also slows our stream loop in the case of error conditions.

One possible enhancement would be to allow users to specify "fatal" error codes, for which the provider would simply transition to FATAL and no longer make requests, but I'm not sure how useful this actually is, so I haven't implemented it here.

This applies to all other provider implementations as well. I will spec it once this is merged.

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamQueueSourceTest.java
```diff
@@ -64,9 +63,9 @@ public Void answer(InvocationOnMock invocation) {
     @Test
     void onNextEnqueuesDataPayload() throws Exception {
         SyncStreamQueueSource connector =
-                new SyncStreamQueueSource(FlagdOptions.builder().build(), mockConnector, stub);
-        connector.init();
+                new SyncStreamQueueSource(FlagdOptions.builder().build(), mockConnector, stub, blockingStub);
         latch = new CountDownLatch(1);
+        connector.init();
```
toddbaert, line 68: unrelated flaky test fix (the init should be after the latch set.

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamQueueSourceTest.java
```diff
@@ -85,9 +84,9 @@ void onNextEnqueuesDataPayload() throws Exception {
     void onNextEnqueuesDataPayloadMetadataDisabled() throws Exception {
         // disable GetMetadata call
         SyncStreamQueueSource connector = new SyncStreamQueueSource(
-                FlagdOptions.builder().syncMetadataDisabled(true).build(), mockConnector, stub);
-        connector.init();
+                FlagdOptions.builder().syncMetadataDisabled(true).build(), mockConnector, stub, blockingStub);
         latch = new CountDownLatch(1);
+        connector.init();
```
toddbaert, line 89: unrelated flaky test fix (the init should be after the latch set.

aepfli (approved):

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/common/ChannelBuilder.java
```diff
@@ -45,6 +110,9 @@ public static ManagedChannel nettyChannel(final FlagdOptions options) {
                     .eventLoopGroup(new EpollEventLoopGroup())
                     .channelType(EpollDomainSocketChannel.class)
                     .usePlaintext()
+                    .defaultServiceConfig(SERVICE_CONFIG_WITH_RETRY)
+                    .maxRetryAttempts(MAX_RETRY_ATTEMPTS)
```
chrfwow: do we need to specifiy this here a second time, when it is already specified in `SERVICE_CONFIG_WITH_RETRY`?

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamQueueSource.java
chrfwow: What happens if the error code we receive is not retryable? Will we then spam the logs again?

toddbaert: Yes, that's why I've made them all retryable except "cancelled", which would only happen on from a client side shutdown: https://github.com/open-feature/java-sdk-contrib/pull/1273#discussion_r1990354285

IMO this entire situation is rare enough that this is an acceptable solution, but we could also add some kind of sleep here, or queue based back pressure.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/common/ChannelBuilder.java
toddbaert: Yes, it's strange, but all the examples do it like this as well.

I will double check.

chrfwow (approved):

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/common/ChannelBuilder.java
toddbaert: I misunderstood the doc the first time I read this. This actually limits the max possible retry events that can be specified in the service config ... so if the service config specifies 11, but this is set to 10, 10 is used. I will remove it, I don't think we need this assurance.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamQueueSource.java
toddbaert: I also removed `DEADLINE_EXCEEDED from the list. My reasoning is added to the comment:

```java
/*
* All codes are retryable except OK, CANCELLED and DEADLINE_EXCEEDED since
* any others not listed here cause a very tight loop of retries.
* CANCELLED is not retryable because it is a client-side termination.
* DEADLINE_EXCEEDED is typically a result of a client specified deadline,
* and definitionally should not result in a tight loop (it's a timeout).
*/
```

PR #1299, opened 2025-03-30
feat(flagd): add http connector for In-process resolver

liran2000:
add [http connector](https://github.com/liran2000/java-sdk-contrib/blob/bd8d1fc772ec8424f61f5eddf0a6783e863f187e/tools/flagd-http-connector/README.md#http-connector) for In-process resolver.


aepfli:
To be honest, I am a little bit torn by this request. I like the option of a custom connector, and I think an HTTP connector is a nice feature to offer. But the flagd provider is already complex and has much additional logic.

Should we maybe start extracting this custom connector into own artifacts? In this case, the flagd provider code won't gain more complexity, the deliverable size is smaller, and people can opt-in with an additional package. It might be worth talking about the general structure of flagd and maybe extracting all connection modes into their connectors (out of scope here) - but it might be worth starting with the HTTP connector.

wdyt?

liran2000:
> To be honest, I am a little bit torn by this request. I like the option of a custom connector, and I think an HTTP connector is a nice feature to offer. But the flagd provider is already complex and has much additional logic.
> 
> Should we maybe start extracting this custom connector into our artifacts? In this case, the flagd provider code won't gain more complexity, the deliverable size is smaller, and people can opt-in with an additional package. It might be worth talking about the general structure of flagd and maybe extracting all connection modes into their connectors (out of scope here) - but it might be worth starting with the HTTP connector.
> 
> wdyt?

Understood. 

One of the main need of this is to reduce infrastructure/devops work, without additional containers needed. At some dev teams, such things can be a blocker for adopting new components, as without dependency on infrastructure/devops work, dev teams can add it independently, on dev repositories alone.

Regarding how to add it, possibly as some separate extension, similar to [junit-open-feature-extension](https://github.com/open-feature/java-sdk-contrib/tree/main/tools/junit-openfeature#junit-open-feature-extension) ?

aepfli:
> Regarding how to add it, possibly as some separate extension, similar to [junit-open-feature-extension](https://github.com/open-feature/java-sdk-contrib/tree/main/tools/junit-openfeature#junit-open-feature-extension) ?

This sounds like a reasonable solution and the setup here would be similar. but currently this is just my opinion, it would be great to hear what overs think about this topic, before we proceed. @toddbaert @chrfwow @open-feature/sdk-java-maintainers (not sure whom to tag, as this is just a provider, and not the basic SDK, not sure if this is relevant for the tc)

guidobrei:
> Should we maybe start extracting this custom connector into our artifacts?

Sounds reasonable to me to make flagd providers extensible. Like we have [different sync sources for flagd](https://flagd.dev/concepts/syncs/) itself, we could also pull this concept down to the provider level.

What I'm wondering is how we would manage alignment of these sources between different provider implementations. Every source protocol added should ideally be available across provider implementations in different languages. 
Flagd provides this abstraction layer already.

I would see the basic logic to connect to flagd as core part of flagd providers. But something like the `HTTP` or also the `File` connectors could be some type of flagd provider extension with maybe also different SLA from our end attached to it.

toddbaert:
I can see why people would want this!

Like @aepfli , my main concern is maintenance burden. We already have a lot of feature and configuration disparity between flagd providers across languages, and I don't want to add more before we've even resolved what's already a problem.

Similar to what @guidobrei  said, my recommendation would be that we keep all the parts of this PR that allow people to do this sort of extension easily themselves (defining interfaces, etc) so that somebody could make an http-connector (or an s3 connector, etc, etc). Then @liran2000 you could host the HTTP connector yourself if you'd like, or possibly as a separate component in contribs. Right now we maintain these sorts of things in flagd itself so people can use http/s3 and we just have that one repo to maintain. Implementing such things in every flagd provider could mean possibly hundreds of additional modules, and I simply don't think we have the maintainership resources for that.

liran2000:
@aepfli @guidobrei @toddbaert 

Can you have a look on the [new structure](https://github.com/open-feature/java-sdk-contrib/blob/59ffacb41273a3eaa4926df28a7d0fa8b6e37f21/tools/flagd-http-connector/README.md#installation), located under tools, what do you think ?

tools/flagd-http-connector/src/main/java/dev/openfeature/contrib/tools/flagd/resolver/process/storage/connector/sync/http/HttpConnector.java
```diff
@@ -0,0 +1,183 @@
...
+            log.debug("got 304 Not Modified, skipping update");
+            return false;
+        }
+        if (payload == null) {
+            log.debug("payload is null");
+            return false;
+        }
+        log.debug("adding payload to queue");
+        if (!this.queue.offer(new QueuePayload(QueuePayloadType.DATA, payload))) {
+            log.warn("Unable to offer file content to queue: queue is full");
```
chrfwow, line 187: Why don't we purge the queue or at least remove the oldest element? Getting new data sounds more important to me thatn keeping the old one

tools/flagd-http-connector/src/main/java/dev/openfeature/contrib/tools/flagd/resolver/process/storage/connector/sync/http/PayloadCacheWrapper.java
```diff
@@ -0,0 +1,59 @@
...
+    public PayloadCacheWrapper(PayloadCache payloadCache, PayloadCacheOptions payloadCacheOptions) {
+        if (payloadCacheOptions.getUpdateIntervalSeconds() < 1) {
+            throw new IllegalArgumentException("pollIntervalSeconds must be larger than 0");
+        }
+        this.updateIntervalMs = payloadCacheOptions.getUpdateIntervalSeconds() * 1000L;
+        this.payloadCache = payloadCache;
+    }
+
+    public void updatePayloadIfNeeded(String payload) {
+        if ((getCurrentTimeMillis() - lastUpdateTimeMs) < updateIntervalMs) {
```
chrfwow: I'm not sure I understand why we would want to ignore updates to the cache. I understand why we wouldn't want to make requests before the update interval runs out to save resources. But if we happen to make a request and get updated data, why would we ignore it?

tools/flagd-http-connector/src/main/java/dev/openfeature/contrib/tools/flagd/resolver/process/storage/connector/sync/http/HttpConnector.java
liran2000: the behavior is pretty consistent with existing behavior of [FileQueueSource](https://github.com/open-feature/java-sdk-contrib/blob/59ffacb41273a3eaa4926df28a7d0fa8b6e37f21/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/file/FileQueueSource.java#L51).
As I see it, it is a corner case which is not expected to happen. Since updates are per interval, and changes processing is fast, queue size not expected to reach the maximum, even no more than 1 or 2. Then not sure need to handle unexpected case.

tools/flagd-http-connector/src/main/java/dev/openfeature/contrib/tools/flagd/resolver/process/storage/connector/sync/http/PayloadCacheWrapper.java
liran2000: The goal is to avoid overloading fallback cache writes, since the cache serves only as a fallback mechanism. Typically, this value can be tuned to be shorter than the cache's TTL, balancing the need to minimize unnecessary updates while still handling edge cases effectively.  
For example, consider a scenario with 10 microservice instances, each polling every 10 seconds and using a shared Redis cache. In this case, the cache would be updated 10 times every 10 seconds.
By exposing this setting, users gain more control to reduce cache write load, especially since the use case is more of an edge case.
While I could lower the default value, I think it’s valuable to provide this option for users to adjust based on their specific needs.

tools/flagd-http-connector/checkstyle-suppressions.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0"?>
```
liran2000, line 1: both checkstyle files should not be here as part of the parent, but CI build is failing on checkstyle without them

tools/flagd-http-connector/pom.xml
```diff
@@ -38,6 +38,36 @@
             <version>3.17.0</version>
         </dependency>
 
+<!--        <dependency>-->
```
chrfwow: Do we still need this commented out code?

tools/flagd-http-connector/README.md
```diff
@@ -14,35 +14,59 @@ reducing traffic, reducing rate limits effects and changes updates. Can be enabl
...
-down-time to recover from during initialization.
-
-### Sample flow
-Sample flow can use:
-- Github as the flags payload source.
-- Redis cache as a fail-safe initialization cache.
-
-Sample flow of initialization during Github down-time window, showing that application can still use flags
-values as fetched from cache.
+* flagd installation is not required, working independently.
```
chrfwow:
```suggestion
* flagd installation is not required, the Http Connector works independently.  
```

tools/flagd-http-connector/README.md
```diff
@@ -14,35 +14,59 @@ reducing traffic, reducing rate limits effects and changes updates. Can be enabl
...
+  when using polling cache.
+* Can serve as an additional provider for fallback or internal backup scenarios using a multi-provider setup.
+
+### What happens if the Http source is down during application startup?
+
+Http Connector supports optional fail-safe initialization using a cache.
+If the initial fetch fails due to source unavailability, it can load the initial payload from the cache instead of
+falling back to default values.
+This ensures smoother startup behavior until the source becomes available again. To be effective, the fallback cache’s
+TTL should be longer than the expected duration of the source downtime during initialization.
```
chrfwow:
```suggestion
This ensures smoother startup behavior until the source becomes available again. To be effective, the TTL of the fallback cache should be longer than the expected duration of the source downtime during initialization.
```

tools/flagd-http-connector/pom.xml
liran2000: removed

tools/flagd-http-connector/src/main/java/dev/openfeature/contrib/tools/flagd/resolver/process/storage/connector/sync/http/HttpConnectorOptions.java
```diff
@@ -0,0 +1,173 @@
...
+            this.usePollingCache = usePollingCache;
+        }
+    }
+
+    @SneakyThrows
+    private void validate(String url, Integer pollIntervalSeconds, Integer linkedBlockingQueueCapacity,
+            Integer scheduledThreadPoolSize, Integer requestTimeoutSeconds, Integer connectTimeoutSeconds,
+            String proxyHost, Integer proxyPort, PayloadCacheOptions payloadCacheOptions,
+            PayloadCache payloadCache, Boolean useFailsafeCache, Boolean usePollingCache) {
+        new URL(url).toURI();
```
aepfli: [nit] this line will be most likely refactored out by the next person, as it seems like a left over, although it is a validation. We should mark this with a comment, or better handle this validation step

tools/flagd-http-connector/src/main/java/dev/openfeature/contrib/tools/flagd/resolver/process/storage/connector/sync/http/HttpConnector.java
```diff
@@ -0,0 +1,225 @@
...
+        if (payloadCache != null && usePollingCache) {
+            log.debug("checking cache for polling payload");
+            String payload = payloadCache.get(POLLING_PAYLOAD_CACHE_KEY);
+            if (payload != null) {
+                log.debug("got payload from polling cache key, skipping update");
+                return true;
+            }
+        }
+        HttpRequest.Builder requestBuilder = HttpRequest.newBuilder()
+            .uri(URI.create(url))
```
aepfli: [question] the uri will be re generated all the time, is this necessary, or could we do this already within the builder, and just use the generated uri all the time

tools/flagd-http-connector/src/main/java/dev/openfeature/contrib/tools/flagd/resolver/process/storage/connector/sync/http/HttpConnector.java
```diff
@@ -0,0 +1,225 @@
...
+        if (!this.queue.offer(new QueuePayload(QueuePayloadType.DATA, flagData))) {
+            log.warn("init: Unable to offer file content to queue: queue is full");
+        }
+    }
+
+    protected Runnable buildPollTask() {
+        return this::fetchAndUpdate;
+    }
+
+    private boolean fetchAndUpdate() {
```
aepfli, line 146: [nit] As I am a big fan of clean code and readable code, could we maybe split this up into smaller chunked methods, which might be easier to test too?

aepfli (commented):
This pr is a good and valuable addition to the general approach. I already outlined whether we should do this within the provider, and I am thankful that you have migrated it to your deliverable. From my side, this PR is good to go, but we should also get the buy-in from the technical steering committee to see if this is the right approach. Thank you!

tools/flagd-http-connector/src/main/java/dev/openfeature/contrib/tools/flagd/resolver/process/storage/connector/sync/http/HttpConnector.java
liran2000: done

tools/flagd-http-connector/src/main/java/dev/openfeature/contrib/tools/flagd/resolver/process/storage/connector/sync/http/HttpConnector.java
liran2000: done

tools/flagd-http-connector/src/main/java/dev/openfeature/contrib/tools/flagd/resolver/process/storage/connector/sync/http/HttpConnectorOptions.java
liran2000: updated

tools/flagd-http-connector/CHANGELOG.md
```diff
@@ -0,0 +1,35 @@
+# Changelog
```
toddbaert, line 1: I think we can clear this changelog except for the title?

tools/flagd-http-connector/src/test/java/dev/openfeature/contrib/tools/flagd/resolver/process/storage/connector/sync/http/HttpConnectorIntegrationTest.java
```diff
@@ -0,0 +1,57 @@
...
+class HttpConnectorIntegrationTest {
+
+    @SneakyThrows
+    @Test
+    void testGithubRawContent() {
+        assumeTrue(parseBoolean("integrationTestsEnabled"));
+        HttpConnector connector = null;
+        try {
+            String testUrl =
+                    "https://raw.githubusercontent.com/open-feature/java-sdk-contrib/58fe5da7d4e2f6f4ae2c1caf3411a01e84a1dc1a/providers/flagd/version.txt";
```
toddbaert: Let's add a file in this PR that will serve as a new target for this once it's merged. It can be a real payload so that the test is a bit less confusing.

tools/flagd-http-connector/README.md
```diff
@@ -0,0 +1,125 @@
...
+    .build();
+
+FlagdOptions options =
+    FlagdOptions.builder()
+        .resolverType(Config.Resolver.IN_PROCESS)
+        .customConnector(connector)
+        .build();
+
+FlagdProvider flagdProvider = new FlagdProvider(options);
+```
```
toddbaert: This is very clear to me; exactly what I expected.

I'm less sure about how I would use and configure the `FailSafeCache`... could you provide a separate snippet for that (I think it should be a separate example so that people just interested in the basics aren't confused).

tools/flagd-http-connector/README.md
```diff
@@ -0,0 +1,125 @@
...
+    Redis->>micro-service-1: failsafe payload
+    Note right of micro-service-1: polling interval passed
+    micro-service-1->>Github: fetch
+    Github->>micro-service-1: payload
+    micro-service-2->>Redis: fetch
+    Redis->>micro-service-2: payload
+    micro-service-3->>Redis: fetch
+    Redis->>micro-service-3: payload
+
+```
```
toddbaert: This is a good feature, but I think what you have here is more of an advanced use case (with a redis backup, etc). I think you should move this into a more "advanced" section (however you want to title it) so that users trying to understand the basics arent confused.

It might be good to put a reall basic mermaid diagram here just explaining the basic use-case with a single service and and HTTP fetch.

tools/flagd-http-connector/README.md
```diff
@@ -0,0 +1,125 @@
...
+    box Cluster
+    participant micro-service-1
+    participant micro-service-2
+    participant micro-service-3
+    participant Redis
+    end
+    participant Github
+    
+    break source downtime
+    micro-service-1->>Github: initialize
```
toddbaert, line 83:
```suggestion
    service-1->>Github: initialize
```

This is a nit, but I'd recommend just saying "service" instead of "mico-service". It's just more general and "micro-service" architectures sometimes bring in some conceptual baggage that aren't helpful.

toddbaert (commented):
Overall the code looks good and I definitely support this approach.

I left some comments about the doc around the caching mechanisms, because I'm a bit unsure about their usage and I don't think I can do a quality review until I fully understand them. If you update the README and examples a bit to explain those better I think I can dive a bit more into the code and then approve. The basic functionality seems very clear to me.

tools/flagd-http-connector/CHANGELOG.md
liran2000: done

tools/flagd-http-connector/src/test/java/dev/openfeature/contrib/tools/flagd/resolver/process/storage/connector/sync/http/HttpConnectorIntegrationTest.java
liran2000: done

tools/flagd-http-connector/README.md
liran2000: done

tools/flagd-http-connector/README.md
liran2000: done

tools/flagd-http-connector/README.md
liran2000: What I did is according to your recommendations here, added a Basic Simple Configuration example, and moved this example to a section of Configuration with fail-safe cache and polling cache, demonstrates a micro-services architectural. As now there is a simple configuration example, I think that the second example demonstrates a micro-services architectural can be useful using these terms, as its under "Sample flows" section, prefer to leave it using this term, as "service" can have multiple meanings to reduce confusion.

beeme1mr (approved):
Looks good, thanks! Once this has been merged and published, it may be worth mentioning it on the flagd Java Provider readme.

tools/flagd-http-connector/src/main/java/dev/openfeature/contrib/tools/flagd/resolver/process/storage/connector/sync/http/HttpConnector.java
```diff
@@ -0,0 +1,227 @@
...
+    @SuppressFBWarnings(value = "EI_EXPOSE_REP", justification = "parent defines the interface")
+    @Override
+    public BlockingQueue<QueuePayload> getStreamQueue() {
+        boolean success = fetchAndUpdate();
+        if (!success) {
+            log.info("failed initial fetch");
+            updateFromCache();
+        }
+        Runnable pollTask = buildPollTask();
+        scheduler.scheduleWithFixedDelay(pollTask, pollIntervalSeconds, pollIntervalSeconds, TimeUnit.SECONDS);
```
toddbaert, line 113: Unless I'm missing something, I think you should do this in init. As written, aren't we possibly scheduling multiple tasks with this every time getStreamQueue is called? I think it's better to do this just once in init and then we get the benefit of the framework making sure this doesn't happen more than once.

toddbaert (changes requested):
Thanks! I think I have only one remaining concern [here](https://github.com/open-feature/java-sdk-contrib/pull/1299/files?short_path=872efbd#r2085028987). Please let me know if that doesn't make sense, but I think we should start polling as soon as we init.

The README changes are great. I totally understand now.

tools/flagd-http-connector/src/main/java/dev/openfeature/contrib/tools/flagd/resolver/process/storage/connector/sync/http/HttpConnector.java
liran2000: Well, intuitively it is right, and it is confusing.
By what I observed by the call hierarchy, the only place calling it is indeed FlagdProvider.initialize(), so it looks to me that this is by the design of the _QueueSource_.
image

providers/flagd/src/main/resources/simplelogger.properties
beeme1mr, line 1: Could the flagd changes be moved to a separate PR? Including the change in this PR will affect the flagd release notes.

liran2000: removed

aepfli (approved):

toddbaert (approved):

toddbaert:
Thanks @liran2000 !

PR #1336, opened 2025-04-24
feat: migrate to Java 11

toddbaert:
- Migrates all modules to Java 11 (by updating parent)
- Fixes build

This will prompt a release for every artifact here. If you really want to avoid this, you can use the old parent version.


toddbaert:
I'm merging this because it's blocking all builds. Please let me know if you have any issues or concerns.

PR #1421, opened 2025-06-17
chore: added upper bound to parent pom range

### LINKED ISSUE #1417: Maven error: The requested parent version range '[0.2,)' does not specify an upper bound
I am using the env-var provider 0.0.10 in a Maven project, and when calling `mvn install` I get the following error:
```
Caused by: org.apache.maven.project.DependencyResolutionException: Could not collect dependencies for project X
Failed to read artifact descriptor for dev.openfeature.contrib.providers:env-var:jar:0.0.10
        Caused by: The requested parent version range '[0.2,)' does not specify an upper bound
```
I am using Maven 3.9.10.

It looks like it's related to the way the parent pom version is specified in the [env-var pom](https://github.com/open-feature/java-sdk-contrib/blob/main/providers/env-var/pom.xml#L5-L10):
```
    <parent>
        <groupId>dev.openfeature.contrib</groupId>
        <artifactId>parent</artifactId>
        <version>[0.2,)</version>
        <relativePath>../../pom.xml</relativePath>
    </parent>
```

Is this a probelm with my specific setup, or can you also reproduce it?

jarebudev:
## This PR
adds upper bound version for the parent pom

- adds this new feature

### Related Issues

Fixes #1417 

### Notes
see https://maven.apache.org/docs/3.2.2/release-notes.html

```
Support version ranges in parent elements ([MNG-2199](https://issues.apache.org/jira/browse/MNG-2199))
Parent elements can now use bounded ranges in the version specification. You can now consistently use ranges for all intra-project dependencies, of which parents are a special case but still considered a dependency of projects that inherit from them. The following is now permissible:

<project>
  <modelVersion>4.0.0</modelVersion>
  <parent>
    <groupId>org.apache</groupId>
    <artifactId>apache</artifactId>
    <version>[3.0,4.0)</version>
  </parent>
  <groupId>org.apache.maven.its.mng2199</groupId>
  <artifactId>valid</artifactId>
  <version>1</version>
  <packaging>pom</packaging>
</project>

Note this requires Maven 3.2.2 so if you use this new feature it is encouraged you add a Maven Enforcer rule to your build to ensure the use of Maven 3.2.2+.
```

### DIFF
--- a/hooks/open-telemetry/pom.xml
+++ b/hooks/open-telemetry/pom.xml
@@ -5,7 +5,7 @@
     <parent>
         <groupId>dev.openfeature.contrib</groupId>
         <artifactId>parent</artifactId>
-        <version>[0.2,0.3)</version>
+        <version>[0.2,1.0)</version>
         <relativePath>../../pom.xml</relativePath>
     </parent>
     <groupId>dev.openfeature.contrib.hooks</groupId>
--- a/hooks/open-telemetry/pom.xml
+++ b/hooks/open-telemetry/pom.xml
@@ -5,7 +5,7 @@
     <parent>
         <groupId>dev.openfeature.contrib</groupId>
         <artifactId>parent</artifactId>
-        <version>[0.2,)</version>
+        <version>[0.2,0.3)</version>
         <relativePath>../../pom.xml</relativePath>
     </parent>
     <groupId>dev.openfeature.contrib.hooks</groupId>
--- a/providers/configcat/pom.xml
+++ b/providers/configcat/pom.xml
@@ -5,7 +5,7 @@
     <parent>
         <groupId>dev.openfeature.contrib</groupId>
         <artifactId>parent</artifactId>
-        <version>[0.2,)</version>
+        <version>[0.2,0.3)</version>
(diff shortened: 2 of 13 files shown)


toddbaert:
cc @aepfli @chrfwow  @justinabrahms 

I guess this is OK, we'll just have to remember to update these if we release new parents...


We could make the upper bound exclusive of 1.0 actually, maybe that would be safer?

chrfwow:
I agree, let's set it to exclusive 1.0

hooks/open-telemetry/pom.xml
```diff
@@ -5,7 +5,7 @@
     <parent>
         <groupId>dev.openfeature.contrib</groupId>
         <artifactId>parent</artifactId>
-        <version>[0.2,)</version>
+        <version>[0.2,0.3)</version>
```
chrfwow:
```suggestion
        <version>[0.2,1.0)</version>
```
This should be safer for the future

chrfwow (changes requested):

aepfli (approved):

chrfwow (approved):

PR #1429, opened 2025-06-18
feat: ofrep provider

### LINKED ISSUE #1413: [FEATURE] Create OFREP provider
### Requirements

OFREP (OpenFeature Remote Flag Evaluation Protocol), is an API specification for feature flagging that allows the use of generic providers to connect to any feature flag management systems that supports the protocol.

We would like to have an OFREP provider in Java.
The Java SDK is a server SDK, so we should follow those [guidelines](https://github.com/open-feature/protocol/blob/main/guideline/dynamic-context-provider.md) when building the provider.

Rahul-Baradol:
## This PR
- Adds OFREP provider

### Related Issues
Fixes #1413

### Follow-up Tasks
- [x] Dev testing
- [x] Added tests
- [x] Docs


providers/ofrep/src/main/java/dev/openfeature/contrib/providers/ofrep/internal/Resolver.java
```diff
@@ -0,0 +1,184 @@
...
+                        .errorMessage("Type mismatch: expected " + type.getSimpleName() + " but got "
+                                + responseValue.getClass().getSimpleName())
+                        .build();
+            }
+
+            Map<String, Object> metadata = response.getMetadata();
+            return ProviderEvaluation.<T>builder()
+                    .value(type.cast(responseValue))
+                    .variant(response.getVariant())
+                    .flagMetadata(convertToImmutableMetadata(metadata))
```
chrfwow: Should we also try to add the flag metadata to responses in case we fail the evaluation?

pom.xml
```diff
@@ -40,6 +40,7 @@
         <module>providers/configcat</module>
```
liran2000, line 40: can add to component owners and release please config

providers/ofrep/src/main/java/dev/openfeature/contrib/providers/ofrep/OfrepProviderOptions.java
```diff
@@ -0,0 +1,99 @@
...
+    @Getter
+    private Long timeout = null;
+
+    private Map<String, List<String>> headers = null;
+
+    private static final String defaultBaseUrl = "http://localhost:8016";
+    private static final Long defaultTimeout = 10000L; // Default timeout of 10 seconds
+
+    /**
+     * Builder class for constructing an instance of OfrepProviderOptions.
```
liran2000: Can consider using Lombok builder. Can give option to headers, single header etc'.

providers/ofrep/src/main/java/dev/openfeature/contrib/providers/ofrep/OfrepProviderOptions.java
```diff
@@ -0,0 +1,99 @@
...
+            this.baseUrl = baseUrl;
+            return this;
+        }
+
+        public Builder headers(Map<String, List<String>> headers) {
+            this.headers = Collections.unmodifiableMap(headers);
+            return this;
+        }
+
+        public Builder timeout(Long milliseconds) {
```
liran2000: maybe receive Duration instead ?

providers/ofrep/src/main/java/dev/openfeature/contrib/providers/ofrep/internal/OfrepApi.java
```diff
@@ -0,0 +1,153 @@
...
+    private final HttpClient httpClient;
+    private Instant nextAllowedRequestTime = Instant.now();
+
+    /**
+     * Constructs an OfrepApi instance with a default HTTP client and JSON serializers.
+     */
+    public OfrepApi() {
+        httpClient = HttpClient.newHttpClient();
+        serializer = new ObjectMapper();
+        deserializer = new ObjectMapper().configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
```
liran2000, line 50: why FAIL_ON_UNKNOWN_PROPERTIES ?
if some service adding some more, it will fail ?

providers/ofrep/src/main/java/dev/openfeature/contrib/providers/ofrep/internal/OfrepApi.java
```diff
@@ -0,0 +1,153 @@
...
+    private final ObjectMapper serializer;
+    private final ObjectMapper deserializer;
+    private final HttpClient httpClient;
+    private Instant nextAllowedRequestTime = Instant.now();
+
+    /**
+     * Constructs an OfrepApi instance with a default HTTP client and JSON serializers.
+     */
+    public OfrepApi() {
+        httpClient = HttpClient.newHttpClient();
```
liran2000: what about other properties of http client, which can be at the options ?
_connectTimeout, proxy, executor_...

providers/ofrep/src/main/java/dev/openfeature/contrib/providers/ofrep/internal/OfrepApi.java
```diff
@@ -0,0 +1,153 @@
...
+                case HttpURLConnection.HTTP_OK:
+                    return deserializer.readValue(body, OfrepResponse.class);
+                case HttpURLConnection.HTTP_UNAUTHORIZED:
+                case HttpURLConnection.HTTP_FORBIDDEN:
+                    throw new GeneralError("authentication/authorization error");
+                case HttpURLConnection.HTTP_BAD_REQUEST:
+                    throw new InvalidContextError("Invalid context: " + body);
+                case HttpURLConnection.HTTP_NOT_FOUND:
+                    throw new FlagNotFoundError("Flag " + key + " not found");
+                case 429:
```
liran2000: retry also on 502/503 ?
if so, can be configurable

providers/ofrep/src/main/java/dev/openfeature/contrib/providers/ofrep/internal/OfrepRequest.java
```diff
@@ -0,0 +1,24 @@
+package dev.openfeature.contrib.providers.ofrep.internal;
+
+import java.util.Collections;
+import java.util.Map;
+
+/**
```
liran2000, line 6: Can leverage lombok for several stuff here

providers/ofrep/src/main/java/dev/openfeature/contrib/providers/ofrep/internal/OfrepResponse.java
```diff
@@ -0,0 +1,62 @@
+package dev.openfeature.contrib.providers.ofrep.internal;
+
+import java.util.Collections;
+import java.util.Map;
+import lombok.Getter;
+import lombok.Setter;
+
+/**
+ * This class represents the response from an OFREP response.
+ */
```
liran2000, line 11: Can use lombok Data annotation instead of all the others ?

providers/ofrep/src/test/java/dev/openfeature/contrib/AppTest.java
```diff
@@ -0,0 +1,33 @@
+package dev.openfeature.contrib;
```
liran2000: More tests expected ?

liran2000 (commented):
thanks for this :smiley: added some comments

providers/ofrep/src/main/java/dev/openfeature/contrib/providers/ofrep/internal/OfrepApi.java
Rahul-Baradol: It won't fail!
We are configuring it NOT TO FAIL on unknown properties, through the second argument (which is set to `false`)

providers/ofrep/src/main/java/dev/openfeature/contrib/providers/ofrep/internal/OfrepApi.java
Rahul-Baradol: Yeah, this is being set while constructing the HttpRequest.
Please see `prepareHttpRequest`

liran2000: it's not the same.
There are parameters on the HttpRequest like you did, and there are parameters on the [HttpClient itself](https://docs.oracle.com/en/java/javase/11/docs/api/java.net.http/java/net/http/HttpClient.html).

Rahul-Baradol: I see...
So the `timeout` of HttpRequest sets the timeout duration for the full request/response cycle,
whereas the `connectTimeout` of HttpClient is for setting timeout to establish the connection

liran2000: Yes.
Please see other params that can be configurable, like proxy and executor

providers/ofrep/src/main/java/dev/openfeature/contrib/providers/ofrep/internal/OfrepResponse.java
Rahul-Baradol: Using `@Data` would give out the reference of `metadata` , which is vulnerable ( captured by `spotbugs` ) ..

We can instead use `@Getter` and `@Setter` at the class level, and override the getter/setter for metadata...

Will refactor this

providers/ofrep/src/main/java/dev/openfeature/contrib/providers/ofrep/internal/OfrepApi.java
Rahul-Baradol: Could you please elaborate here?

What is the retry policy, and what is expected to be configurable?
Because I couldn't find anything about handling 501/503 in [guidelines](https://github.com/open-feature/protocol/blob/main/guideline/dynamic-context-provider.md)

liran2000: Ok then, can probably ignore

providers/ofrep/src/main/java/dev/openfeature/contrib/providers/ofrep/internal/OfrepResponse.java
liran2000: Yes can return immutable copy at getter

providers/ofrep/README.md
```diff
@@ -0,0 +1 @@
+### Todo
```
liran2000: should add readme with configuration and usage

providers/ofrep/src/main/java/dev/openfeature/contrib/providers/ofrep/OfrepProviderOptions.java
```diff
@@ -0,0 +1,32 @@
...
+public class OfrepProviderOptions {
+
+    @Builder.Default
+    private final String baseUrl = "http://localhost:8016";
+
+    @Builder.Default
+    private final ProxySelector proxySelector = ProxySelector.getDefault();
+
+    @Builder.Default
+    private final Executor executor = Runnable::run;
```
liran2000: I think the default should be something like _Executors.newFixedThreadPool(DEFAULT_THREAD_POOL_SIZE)_ with default value 4-8 and configurable or _Executors.newCachedThreadPool()_, otherwise it can lead to too many threads theoretically.

providers/ofrep/src/main/java/dev/openfeature/contrib/providers/ofrep/OfrepProvider.java
```diff
@@ -0,0 +1,103 @@
...
+import dev.openfeature.sdk.FeatureProvider;
+import dev.openfeature.sdk.Metadata;
+import dev.openfeature.sdk.ProviderEvaluation;
+import dev.openfeature.sdk.Value;
+import org.apache.commons.validator.routines.UrlValidator;
+
+/**
+ * OpenFeature provider for OFREP.
+ */
+public final class OfrepProvider implements FeatureProvider {
```
liran2000, line 19: can add shutdown to close the executor.

providers/ofrep/src/main/java/dev/openfeature/contrib/providers/ofrep/internal/OfrepApi.java
```diff
@@ -0,0 +1,127 @@
...
+     * @return HttpRequest ready to be sent
+     * @throws JsonProcessingException - if an error occurred while processing the json.
+     */
+    private <T> HttpRequest prepareHttpRequest(
+            final URI url, ImmutableMap<String, ImmutableList<String>> headers, final T requestBody)
+            throws JsonProcessingException {
+
+        HttpRequest.Builder reqBuilder = HttpRequest.newBuilder()
+                .uri(url)
+                .header("Content-Type", "application/json; charset=utf-8")
```
liran2000, line 70: can leave _requestTimeout_ here and have both _requestTimeout_ and _connectTimeout_ to be configurable.

providers/ofrep/src/main/java/dev/openfeature/contrib/providers/ofrep/OfrepProvider.java
```diff
@@ -0,0 +1,125 @@
...
+    public Metadata getMetadata() {
+        return () -> OFREP_PROVIDER;
+    }
+
+    @Override
+    public void shutdown() {
+        if (executor instanceof ExecutorService) {
+            try {
+                ExecutorService executorService = (ExecutorService) executor;
+                executorService.shutdownNow();
```
liran2000: can call _shutdown_ first for graceful termination, see [shutdownAndAwaitTermination ](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ExecutorService.html).

liran2000 (approved):
I see the review comments were addressed, thanks.
Added some more review  comments on last commits, approving, can wait for others to review as well.

providers/ofrep/src/main/java/dev/openfeature/contrib/providers/ofrep/internal/Resolver.java
```diff
@@ -0,0 +1,246 @@
...
+                case HttpURLConnection.HTTP_UNAUTHORIZED:
+                case HttpURLConnection.HTTP_FORBIDDEN:
+                    return handleGeneralError(
+                            defaultValue, immutableMetadata, "authentication/authorization error for flag: " + key);
+                case HttpURLConnection.HTTP_BAD_REQUEST:
+                case HttpURLConnection.HTTP_NOT_ACCEPTABLE:
+                    return handleInvalidContext(key, defaultValue, immutableMetadata);
+                case HttpURLConnection.HTTP_NOT_FOUND:
+                    return handleFlagNotFound(key, defaultValue, immutableMetadata);
+                case 429:
```
chrfwow, line 78: Can we also use a constant from `HttpURLConnection` for 429 responses?

Rahul-Baradol: Surprisingly, there is no constant defined in `HttpURLConnection` for status code 429

providers/ofrep/src/main/java/dev/openfeature/contrib/providers/ofrep/internal/OfrepRequest.java
```diff
@@ -0,0 +1,12 @@
+package dev.openfeature.contrib.providers.ofrep.internal;
+
+import com.google.common.collect.ImmutableMap;
+import lombok.Data;
+
+/**
+ * Represents the request body for the OFREP API request.
+ */
+@Data
```
aepfli: [question] As this contains an immutable map, and I think we don't want to tamper with this anyway, maybe `@Value` is better suited, which is more like a record? see https://projectlombok.org/features/Value

providers/ofrep/src/main/java/dev/openfeature/contrib/providers/ofrep/internal/OfrepResponse.java
aepfli: out of curiosity, could the `@value` annotation be helpful here?

aepfli (approved):
Thank you this looks good to me. I am just questioning if we should maybe more often use the `@Value` lombok annotation to generate record like objects. I think this could reduce some of the boilderplate even more here. Thank you

providers/ofrep/src/main/java/dev/openfeature/contrib/providers/ofrep/internal/OfrepRequest.java
Rahul-Baradol: Yeah you are right!

Since we want `OfrepRequest` DTO to be immutable, we can use  `@Value` inplace of `@Data`.
Because we are using this DTO only for serialization and not for deserialization, we can go ahead with this.

Have made the changes.

providers/ofrep/src/main/java/dev/openfeature/contrib/providers/ofrep/internal/OfrepResponse.java
Rahul-Baradol: > could the `@Value` annotation be helpful here?

This will cause problems during deserialization of HTTP response, because `@Value` does not implement setters and NoArgConstructor, and Jackson by default needs them for deserialization.

There are workarounds like `@JsonCreator` and `@JsonProperty`, but I feel sticking with the current implementation makes it more simple.

What do you say !?

Rahul-Baradol:
How can we proceed here?

chrfwow (approved):

aepfli:
from my perspective this is good to go, but I would love to have somebody from the OFREP sig also as part of the reviewers @thomaspoignant @lukas-reining (or maybe somebody else)

providers/ofrep/src/main/java/dev/openfeature/contrib/providers/ofrep/internal/OfrepApi.java
```diff
@@ -0,0 +1,131 @@
...
+     * Constructs an OfrepApi instance with a HTTP client and JSON serializers.
+     *
+     * @param requestTimeout - The request timeout duration for the request.
+     * @param connectTimeout - The connect timeout duration for establishing HTTP connection.
+     * @param proxySelector  - The ProxySelector to use for HTTP requests.
+     * @param executor       - The Executor to use for operations.
+     */
+    public OfrepApi(Duration requestTimeout, Duration connectTimeout, ProxySelector proxySelector, Executor executor) {
+        httpClient = HttpClient.newBuilder()
+                .connectTimeout(connectTimeout)
```
toddbaert, line 44: We could possibly add keepAlive settings, but by default I think there's a 30s pooled connection so I think that's good for now.

providers/ofrep/src/main/java/dev/openfeature/contrib/providers/ofrep/internal/OfrepApi.java
```diff
@@ -0,0 +1,131 @@
...
+            URI uri = URI.create(fullPath);
+
+            HttpRequest request = prepareHttpRequest(uri, headers, requestBody);
+
+            HttpResponse<String> response = httpClient.send(request, HttpResponse.BodyHandlers.ofString());
+            OfrepResponse responseBody = deserializer.readValue(response.body(), OfrepResponse.class);
+
+            return new Resolution(response.statusCode(), response.headers(), responseBody);
+        } catch (JsonProcessingException e) {
+            throw new GeneralError("Error processing JSON: " + e.getMessage());
```
toddbaert:
```suggestion
            throw new ParseError("Error processing JSON: " + e.getMessage());
```

I think this is a good use-case for a `ParseError`.

providers/ofrep/pom.xml
```diff
@@ -0,0 +1,85 @@
+<?xml version="1.0"?>
+<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>dev.openfeature.contrib</groupId>
+        <artifactId>parent</artifactId>
+        <version>0.2.2</version>
```
toddbaert:
```suggestion
        <version>1.0.0</version>
```

.github/component_owners.yml
```diff
@@ -37,6 +37,8 @@ components:
     - liran2000
   providers/multiprovider:
     - liran2000
+  providers/ofrep-provider:
+    - Rahul-Baradol
```
toddbaert, line 41:
```suggestion
    - Rahul-Baradol
    - toddbaert
```

providers/ofrep/src/main/java/dev/openfeature/contrib/providers/ofrep/internal/OfrepApi.java
```diff
@@ -0,0 +1,131 @@
+package dev.openfeature.contrib.providers.ofrep.internal;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.DeserializationFeature;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import dev.openfeature.sdk.exceptions.GeneralError;
```
toddbaert, line 8:
```suggestion
import dev.openfeature.sdk.exceptions.GeneralError;
import dev.openfeature.sdk.exceptions.ParseError;
```

toddbaert (approved):
@Rahul-Baradol I made a few very small changes:

- used a `ParseError` instead of a `GeneralError` where I thought it made sense
- updated the parent pom dependency requiring latest SDK version
- added myself to the owners

If that's all OK with you, don't do anything! I will merge tomorrow. If not, please say so!

Rahul-Baradol:
Ohh no issues!
You merged it already 😄

PR #1478, opened 2025-07-06
feat: Add dedicated annotations for boolean, String, int and double flags in Junit5 extension

### LINKED ISSUE #923: [Junit5 extension] add dedicated Annotations for Number, Boolean and String flags
## Problem

Our annotations for testing are excellent, and they greatly ease the developer experience. Currently, all the values need to be provided as a string. This is a limitation of java annotation as java annotation members can only be (https://docs.oracle.com/javase/specs/jls/se8/html/jls-9.html#jls-9.6.1):
- A primitive type
- String
- Class or an invocation of Class ([§4.5](https://docs.oracle.com/javase/specs/jls/se8/html/jls-4.html#jls-4.5))
- An enum type
- An annotation type
- An array type whose component type is one of the preceding types

Furthermore, annotations can't be generic, nor do they support inheritance.

```java
@Flag(name = "BOOLEAN_FLAG", value = "true")
```

## Goal

Evaluate the feasibility and effort of using meta-annotations or some kind of inheritance. 
We should not make the API more complex than it already is. 
Maybe we only support those convenience annotations for the simple annotation approach and not in the extended configuration.

like instead of 
```java
@Flag(name = "BOOLEAN_FLAG", value = "true")
@Flag(name = "STRING_FLAG", value = "true")
@Flag(name = "Number_FLAG", value = "1")
```
we would love to use
```java
@BooleanFlag(name = "BOOLEAN_FLAG", value = true)
/* optional boolean improvement
@BooleanFlagTrue(name = "BOOLEAN_FLAG")
@BooleanFlagFalse(name = "BOOLEAN_FLAG")
*/
@StringFlag(name = "STRING_FLAG", value = "true")
@NumberFlag(name = "Number_FLAG", value = 1)
```

ssharaev:
## This PR

- adds dedicated annotations `BooleanFlag`, `StringFlag`, `IntegerFlag`, and `DoubleFlag` for more specific and type-safe flag definitions in JUnit tests

### Related Issues

Fixes #923

### Notes
Not sure about tests, feedback on the test coverage is very welcome!

### DIFF
--- a/tools/junit-openfeature/src/main/java/dev/openfeature/contrib/tools/junitopenfeature/OpenFeatureExtension.java
+++ b/tools/junit-openfeature/src/main/java/dev/openfeature/contrib/tools/junitopenfeature/OpenFeatureExtension.java
@@ -3,9 +3,15 @@ package dev.openfeature.contrib.tools.junitopenfeature;
 import dev.openfeature.sdk.OpenFeatureAPI;
 import dev.openfeature.sdk.providers.memory.Flag;
 import java.lang.reflect.Method;
+import java.util.AbstractMap;
 import java.util.Arrays;
 import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import org.apache.commons.lang3.BooleanUtils;
 import org.junit.jupiter.api.extension.AfterEachCallback;
 import org.junit.jupiter.api.extension.BeforeEachCallback;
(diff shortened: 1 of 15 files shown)


tools/junit-openfeature/src/test/java/dev/openfeature/contrib/tools/junitopenfeature/BooleanFlagTest.java
```diff
@@ -48,13 +65,51 @@ void multipleFlagsSimple() {
...
+        void multipleTypedFlagsSimple() {
+            Client client = OpenFeatureAPI.getInstance().getClient();
+            assertThat(client.getBooleanValue(FLAG, false)).isTrue();
+            assertThat(client.getBooleanValue(FLAG + "2", false)).isTrue();
+        }
+
+        @Test
+        @BooleanFlag(name = FLAG, value = true)
+        @BooleanFlag(name = FLAG, value = false)
+        void duplicatedTypedFlagsSimple() {
```
chrfwow: Please use more descriptive names for your tests, maybe something like `duplicateFlagKeyOverridesPreviousDeclaration`

chrfwow:
@ssharaev you will need to force push a commit that is signed off, otherwise the DCO check will not succeed

tools/junit-openfeature/src/test/java/dev/openfeature/contrib/tools/junitopenfeature/DoubleFlagTest.java
```diff
@@ -29,6 +37,13 @@ void existingSimpleFlagIsRetrieved() {
             assertThat(client.getDoubleValue(FLAG, FALLBACK)).isEqualTo(FLAG_VALUE);
         }
 
+        @Test
+        @DoubleFlag(name = FLAG, value = 1.d)
```
chrfwow, line 41:
```suggestion
@DoubleFlag(name = FLAG, value = FLAG_VALUE)
```
also for the other tests

tools/junit-openfeature/src/main/java/dev/openfeature/contrib/tools/junitopenfeature/OpenFeatureExtension.java
```diff
@@ -23,37 +29,127 @@ public class OpenFeatureExtension implements BeforeEachCallback, AfterEachCallba
...
+    private static <T> void addFlag(
+            Map<String, Flag<?>> domainFlags, Set<String> domainFlagNames, String flagName, T value) {
+        if (domainFlagNames.contains(flagName)) {
+            throw new IllegalArgumentException("Flag with name " + flagName + " already exists. "
+                    + "There shouldn't be @Flag and @" + value.getClass().getSimpleName() + "Flag with the same name!");
+        }
+
+        if (domainFlags.containsKey(flagName)) {
+            return;
+        }
```
chrfwow, line 105: One of these checks is redundant. I think we could even remove the `domainFlagNames` set entirely, as it contains the same information as the `domainFlags` map.

tools/junit-openfeature/src/test/java/dev/openfeature/contrib/tools/junitopenfeature/BooleanFlagTest.java
```diff
@@ -48,13 +65,51 @@ void multipleFlagsSimple() {
...
+            assertThat(client.getBooleanValue(FLAG, false)).isTrue();
+            assertThat(client.getBooleanValue(FLAG + "2", false)).isTrue();
+        }
+
+        @Test
+        @BooleanFlag(name = FLAG, value = true)
+        @BooleanFlag(name = FLAG, value = false)
+        void duplicatedTypedFlagDoesntOverridePrevious() {
+            Client client = OpenFeatureAPI.getInstance().getClient();
+            assertThat(client.getBooleanValue(FLAG, false)).isTrue();
```
guidobrei, line 82: Reading the annotations I would have expected the last definition of `FLAG` to win and resolve to false.
Is this the same behavior as with multiple `@Flag` annotations?

tools/junit-openfeature/src/test/java/dev/openfeature/contrib/tools/junitopenfeature/BooleanFlagTest.java
```diff
@@ -81,6 +138,13 @@ void existingSimpleFlagIsRetrieved() {
             assertThat(client.getBooleanValue(FLAG, false)).isTrue();
         }
 
+        @Test
+        @BooleanFlag(name = FLAG, value = true)
+        void existingSimpleTypedFlagIsRetrieved() {
+            Client client = OpenFeatureAPI.getInstance().getClient("testSpecific");
+            assertThat(client.getBooleanValue(FLAG, false)).isTrue();
```
guidobrei, line 145: What does this test do? `FLAG` is defined on `Class` and `Method` level, both with true.

tools/junit-openfeature/src/test/java/dev/openfeature/contrib/tools/junitopenfeature/BooleanFlagTest.java
```diff
@@ -169,17 +310,68 @@ void multipleFlags() {
...
+        }
+
+        @Nested
+        @Disabled
+        class ConfigWithException {
+
+            @Test
+            @Flag(name = FLAG, value = "true")
+            @BooleanFlag(name = FLAG, value = true)
+            void simpleConfigDuplicateFlags() {
```
guidobrei, line 365: Interesting, that defining `FLAG` twice as simple flag and type flag on the method level throws, an exception, but a simple flag `FLAG` on class level and a typed flag `FLAG` on method level does not throw.
See `existingSimpleTypedFlagIsRetrieved()` on [L143](https://github.com/open-feature/java-sdk-contrib/pull/1478/files#diff-944efca46c72af838af3e1225d44d60460efeac6c4fb18853e88dfea4dd5168cR143)

guidobrei: Just saw, this test class is disabled. Is this tech debt?

PR #1511, opened 2025-07-11
feat: allowing null/missing default values

### LINKED ISSUE #1438: [flagd] Update in-process Evaluation To Allow undefined or `null` `defaultValue`
# [flagd] Update in-process Evaluation To Allow undefined or `null` `defaultValue`

We want to support optional/null `defaultValue` for flag configurations. The in-process provider must be updated to support this.

# Steps

- [ ] return `FLAG_NOT_FOUND` for flag configurations missing a `defaultValue` and for which targeting doesn't resolve a variant (This is the critical behavioral change)
- [ ] add associated gherkin tests from the test harness

This issue requires https://github.com/open-feature/flagd-schemas/issues/186 and https://github.com/open-feature/flagd-testbed/issues/268 to be done first.

Rahul-Baradol:
## This PR
- supports missing/null default values, by returning `FLAG_NOT_FOUND`,
  - for flag configurations missing a defaultValue, and 
  - for which targeting doesn't resolve a variant and there is no defaultValue

### Related Issues
Fixes #1438


providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolver.java
```diff
@@ -203,6 +204,12 @@ private <T> ProviderEvaluation<T> resolve(Class<T> type, String key, EvaluationC
         // check variant existence
         Object value = flag.getVariants().get(resolvedVariant);
         if (value == null) {
+            if (resolvedVariant.isEmpty()) {
```
guidobrei: The resolved variant will probably not be an empty string, but `null`. Especially, if the `defaultVariant` property is not set in the data source.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolver.java
```diff
@@ -203,6 +204,12 @@ private <T> ProviderEvaluation<T> resolve(Class<T> type, String key, EvaluationC
...
+                String message = String.format("no default variant found in flag with key %s", key);
+                log.debug(message);
+                throw new FlagNotFoundError(message);
```
guidobrei: [Support Explicit Code Default Values in flagd Configuration](https://github.com/open-feature/flagd/blob/2f330af516376238e889064500a285b13ce8bb9a/docs/architecture-decisions/support-code-default.md?plain=1#L10) will be a default code flow, not an exceptional case. 
Throwing an exception can be problematic in performance critical use-cases. 

Consider returning a `ProviderEvaluaitonResult` with proper reason an error code instead of throwing an exception.

The same holds for the debug log message. The message is surely valuable. Consider checking if debug is enabled before constructing a new message. Or use a `log.debug()` overload that accepts format and arguments but does not concatenate the string beforehand.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolver.java
```diff
@@ -203,6 +204,12 @@ private <T> ProviderEvaluation<T> resolve(Class<T> type, String key, EvaluationC
         // check variant existence
         Object value = flag.getVariants().get(resolvedVariant);
         if (value == null) {
+            if (resolvedVariant.isEmpty()) {
+                String message = String.format("no default variant found in flag with key %s", key);
```
guidobrei: The message could be misleading. The resolved variant could also be the result of a misconfigured targeting rule. E.g., if a targeting rule specifies an empty variant.

guidobrei (commented):
Thanks for the contribution, @Rahul-Baradol. 
I've added some comments on things that cached my attention. 

Wondreing what the [Gherkin tests](https://github.com/open-feature/flagd-testbed/pull/269) will reveal.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolver.java
chrfwow: According to the [spec](https://openfeature.dev/specification/sections/flag-evaluation/#requirement-1411), there should be no logging on hot code paths

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolver.java
Rahul-Baradol: Surprisingly, the value is coming out be `empty string` even if the `defaultVariant` is not set in the config 🤔

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolver.java
Rahul-Baradol: Noted!

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolver.java
Rahul-Baradol: Yeah, actually!
Need to figure out another way to solve this.

toddbaert:
@Rahul-Baradol if it's any help, [here is my JS implementation](https://github.com/open-feature/js-sdk-contrib/pull/1354) of the same thing. I'm happy to add the e2e tests here once your implementation is complete, so you can ignore that part.

beeme1mr:
FYI, I've updated the title to remove the `!` since this isn't a breaking change.

aepfli:
@Rahul-Baradol i hope you dont mind me fixing your conflict, and activating our e2e test suite for your feature. Now this code runs against the e2e tests which we are also using in other languages.

Rahul-Baradol:
Sure, I don't have issues !

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolver.java
```diff
@@ -204,6 +205,12 @@ private <T> ProviderEvaluation<T> resolve(Class<T> type, String key, EvaluationC
...
+                log.debug(message);
+                throw new FlagNotFoundError(message);
```
aepfli:
```suggestion
                return ProviderEvaluation.<T>builder()
                        .reason(Reason.ERROR.toString())
                        .errorCode(ErrorCode.FLAG_NOT_FOUND)
                        .errorMessage("Flag '" + key+ "' has no default variant defined, will use code default")
                        .flagMetadata(getFlagMetadata(storageQueryResult))
                        .build();
```
most likely this will do the trick and is in sync with the js implementation

aepfli (approved):

guidobrei (approved):

PR #1513, opened 2025-07-14
test(flagd): Reduce flakiness of ChannelConnectorTest by using random port

aepfli:
The ChannelConnectorTests were flaky on GitHub actions, as they always used 8080 for their tests. With this change, we will use a random port for each test run.

Theoretically, the same port could still be assigned, but the chance would be much smaller than the fixed port.

## How to test?

Start any software that will use port 8080 (e.g., a Docker web server) and run the test. It should pass without any issues.

```
docker run \
  --rm -it \
  -p 8080:8080 \
  ghcr.io/open-feature/flagd-testbed
```

If you run the same on main, the unit test will fail.


providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/resolver/common/ChannelConnectorTest.java
```diff
@@ -57,12 +58,16 @@ void setUp() throws Exception {
     }
 
     private void setupTestGrpcServer() throws IOException {
+        var testSocket = new ServerSocket(0);
```
aepfli, line 61: this will pick a random available port :)

chrfwow: Can we just assign a random but fixed port that is not 8080?

aepfli: we need a different port per run of the test, else, if the test runs in parallel, we might have the same issue again

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/resolver/common/ChannelConnectorTest.java
```diff
@@ -57,12 +58,16 @@ void setUp() throws Exception {
...
+        var port = testSocket.getLocalPort();
```
chrfwow, line 62:
```suggestion
        var port = new Random().nextInt(49152, 65535);
```
If we just want a random port number, why not use a random number?

aepfli: Because a service can still allocate the port, we don't know what is running on the host, and `ServerSocket(0)` will pick an unallocated port. 

It is like picking a random number and checking if the port is available.

chrfwow (approved):

liran2000 (approved):

PR #1541, opened 2025-07-21
ci(go-feature-flag): Fix flaky maven generate-resources stage

### LINKED ISSUE #1540: [build] flaky builds due to `download-latest-github-asset`
I tried to do a run for dependency updates today, to see and update build deps etc. A lot of prs are not mergable due to the same reason, and the builds are more often failing than succeeding.

```
Error:  Failed to execute goal org.codehaus.mojo:exec-maven-plugin:3.5.1:exec (download-latest-github-asset) on project go-feature-flag: Command execution failed. Process exited with an error: 1 (Exit value: 1) -> [Help 1]
```

we should investigate, if we can cache those dependencies somehow for the build. or if there is a more reliable mechanism to fetch them.

fyi: @thomaspoignant maybe you have some input here ;)

thomaspoignant:
## This PR
This PR fixes a flacky generate-resources stage by using a git submodule instead of downloading it from maven.

### Related Issues
Fixes #1540


toddbaert:
Seeing some issue with this in the CI:

```
/home/runner/work/java-sdk-contrib/java-sdk-contrib/providers/go-feature-flag
mv: cannot stat './wasm-releases/evaluation/gofeatureflag-evaluation_v1.45.4.wasi': No such file or directory
```

thomaspoignant:
Yes I was in rush and I did not finish it.
I am putting back as a draft will work on it tomorrow

.github/workflows/ci.yml
```diff
@@ -24,6 +24,8 @@ jobs:
     steps:
       - name: Checkout Repository
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
+        with:
+          submodules: recursive
```
thomaspoignant, line 28: I am adding that part because without that it was not loading the added submodule.

thomaspoignant:
@toddbaert I am good with the PR now if you want to check.

.gitmodules
```diff
@@ -8,3 +8,6 @@
 [submodule "providers/flagd/spec"]
 	path = providers/flagd/spec
 	url = https://github.com/open-feature/spec.git
+[submodule "providers/go-feature-flag/wasm-releases"]
+	path = providers/go-feature-flag/wasm-releases
+	url = https://github.com/go-feature-flag/wasm-releases.git
```
aepfli, line 13: do we want to fix this to a tag/version of is main always fine? (see a few lines above the `branch` property, renovate will use it for updates)

aepfli (approved):

.gitmodules
thomaspoignant: I am deploying the wasm module on main only for the moment, so I think `main` will always be fine.

PR #1549, opened 2025-07-28
feat(flagd): Adjust to disable-sync-metadata toggle in flagd

### LINKED ISSUE #1509: [flagd] use `sync_context` field instead of `GetMetadata`
# [flagd] use `sync_context` field instead of `GetMetadata`

## Background

Currently, to support static context enrichment via attributes sent from flagd, in-process providers call [sync.GetMetadata](https://buf.build/open-feature/flagd/docs/main:flagd.sync.v1#flagd.sync.v1.FlagSyncService.GetMetadata) along with starting the stream. This is not ideal. The stream isn't valid without the metadata response (since it could result in rules not functioning as expected due to missing attributes) and  the metadata response is useless without the stream. 

This has been implemented in flagd with [this PR](https://github.com/open-feature/flagd/pull/1642).
We need to absorb this change by making use of this new field in in-process providers

## Requirements

- [ ] while initializing the `SyncFlags` stream, continue to call `getMetadata`, but override it with the results from the `sync_context` field, if present (it's optional, so its presence can be checked).
- [ ] associated tests

leakonvalinka:
## This PR

- adjusts the implementation to handle both paths that the `disable-sync-metadata` toggle introduces

### Related Issues

relates: [#1584](https://github.com/open-feature/flagd/issues/1584)
closes: https://github.com/open-feature/java-sdk-contrib/issues/1509

### DIFF
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/FlagStore.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/FlagStore.java
@@ -8,7 +8,6 @@ import dev.openfeature.contrib.providers.flagd.resolver.process.model.FlagParser
 import dev.openfeature.contrib.providers.flagd.resolver.process.model.ParsingResult;
 import dev.openfeature.contrib.providers.flagd.resolver.process.storage.connector.QueuePayload;
 import dev.openfeature.contrib.providers.flagd.resolver.process.storage.connector.QueueSource;
-import dev.openfeature.sdk.ImmutableContext;
 import dev.openfeature.sdk.ImmutableStructure;
 import dev.openfeature.sdk.Structure;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/FlagStore.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/FlagStore.java
@@ -8,6 +8,7 @@ import dev.openfeature.contrib.providers.flagd.resolver.process.model.FlagParser
 import dev.openfeature.contrib.providers.flagd.resolver.process.model.ParsingResult;
 import dev.openfeature.contrib.providers.flagd.resolver.process.storage.connector.QueuePayload;
 import dev.openfeature.contrib.providers.flagd.resolver.process.storage.connector.QueueSource;
+import dev.openfeature.sdk.ImmutableContext;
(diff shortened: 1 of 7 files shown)


providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/FlagStore.java
```diff
@@ -150,9 +150,9 @@ private void streamerListener(final QueueSource connector) throws InterruptedExc
         log.info("Shutting down store stream listener");
     }
 
-    private Structure parseSyncMetadata(GetMetadataResponse metadataResponse) {
+    private Structure parseSyncContextOrMetadata(Struct syncContext) {
```
aepfli: [nit] to future proof this a little, we should just rename it to:

```suggestion
    private Structure parseSyncContext(Struct syncContext) {
```

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/FlagStore.java
```diff
@@ -150,9 +150,9 @@ private void streamerListener(final QueueSource connector) throws InterruptedExc
...
         try {
-            return convertProtobufMapToStructure(metadataResponse.getMetadata().getFieldsMap());
+            return convertProtobufMapToStructure(syncContext.getFieldsMap());
```
aepfli: [Suggestion] as `syncContext` can now be `null`, we should add a null check around this, and maybe return an `ImmutableStructure` early if that is the case. Especially that exceptions are more expensive than a null check.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/FlagStore.java
```diff
@@ -114,7 +114,7 @@ private void streamerListener(final QueueSource connector) throws InterruptedExc
                         Map<String, FeatureFlag> flagMap = parsingResult.getFlags();
                         Map<String, Object> flagSetMetadataMap = parsingResult.getFlagSetMetadata();
 
-                        Structure metadata = parseSyncMetadata(payload.getMetadataResponse());
+                        Structure metadata = parseSyncContextOrMetadata(payload.getSyncContext());
```
aepfli: [suggestion] for consistency we should also rename the variable to `syncContext`

aepfli (approved):
Thank you! Can't wait to fully get rid of the Metadata endpoint, maybe we should also mark the configuration `syncMetadataDisabled` as deprecated and to be removed in the next major version

chrfwow (approved):

PR #1589, opened 2025-09-17
fix(flagd): fix wrong environment variable and test execution

aepfli:
see https://github.com/cucumber/cucumber-jvm/issues/2984


providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdOptions.java
```diff
@@ -123,7 +123,7 @@ public class FlagdOptions {
      * ProviderId to be used with flag sync gRPC contract.
      **/
     @Builder.Default
-    private String providerId = fallBackToEnvOrDefault(Config.SOURCE_PROVIDER_ID_ENV_VAR_NAME, null);
```
beeme1mr, line 126: Should we retain the previous value for backwards compatibility reasons?

aepfli: not sure about this, what is the expectation - should the others then also implement the "non spec conformant" environment name? go lang uses the same env var, python does not - currently this is a mixed setting.

aepfli: added a fallback to the fallback

chrfwow (approved):

PR #1590, opened 2025-09-17
fix(flagd): improve stream observer, refine retry policy; don't use retry to avoid busy loop

### LINKED ISSUE #1583: [flagd] Reconnect Functionality Fails Under Certain Conditions, Transitioning System to Error State
### Bug Summary

- **Frequency**: Occurs irregularly, approximately 1-2 times per week, affecting only one pod out of ~100 in that time frame.
- **Issue**: Reconnect functionality fails under certain conditions, transitioning the system from a stale to an error state.

### Impact

- No negative impact on the running system.
- Feature flag updates stop functioning.
- Typically affects only one pod, not multiple pods.

#### Critical Error
- When the error \"[d.o.c.providers.flagd.FlagdProvider] Provider did not reconnect successfully within 60s Emitting ERROR *\" is thrown, recovery is highly unlikely, if this was not during a development. We have a 60 seconds grace period, which offers a lot of time for resolving connection issues.
- Once this error occurs, the system transitions from stale to error, and the affected pod will most likely remain in the error state indefinitely.

#### Transition Behavior
- The system transitions from a stale to an error state after a 60-second recovery window.
- If recovery does not occur within this time frame, the system will not recover on its own.
- This is especially problematic when only one pod of many is affected.

#### During Deployments
- In rare cases, the issue occurs during deployments, leading to a peak where all pods are affected simultaneously, but they recover.

#### Recovery Attempts
- Restarting the proxies to reset the connection does not resolve the issue. (so it is not on flagd side)
- The problem appears to be related to the connection thread hanging, closing, or entering an undefined state.

#### Tests
- Existing tests for reconnection behavior (e.g., when proxies are unavailable) do not seem to cover this scenario.
- The issue remains hard to debug due to its irregular nature and lack of observable patterns.

#### Hypothesis
- The root cause may involve the connection thread entering an unresponsive or closed state, particularly in edge cases such as deployments or specific pod configurations.

#### Proposed Action
- Implement an additional recovery mechanism:
  - If the system remains in an error state for an extended period (e.g., exceeding the 60-second recovery window), the entire connection will be reset and re-initialized.
  - This mechanism aims to ensure recovery even when individual pods are affected.
- Explore whether the recovery mechanism can account for deployment-related peaks affecting all pods simultaneously.

---

**Please investigate and prioritize accordingly.**

guidobrei:
## This PR
- fixes a potential deadlock in FlagSyncService
- fixes synchronization issues between getMetadata and syncFlags
- instead of using the gRPC retry policy to prevent a busy loop on connection errors, use a simple sleep 

### Related Issues

Fixes #1583 

### Notes

We've identified two related issues in current implementation:

1) It potentially waits forever on receiving a message on a queue by calling `incomingQueue.take()` some message put in the blocking queue.
2) If `getMetadata` fails we want to cancel the stream. However, this never worked properly because the two parts listened on different cancellation tokens.

The new implementation eliminates problem 1) by removing `incomingQueue` at all and mitigates 2) by sharing a cancellation token.

### Follow-up Tasks

Still open is the removal of deprecated getMetadata https://github.com/open-feature/flagd/issues/1584. 

### How to test
Clagd testbed covers FlagSyncService


providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamQueueSource.java
```diff
@@ -115,82 +111,57 @@ public void shutdown() throws InterruptedException {
...
-                        if (!outgoingQueue.offer(new QueuePayload(
-                                QueuePayloadType.ERROR,
-                                String.format("Error from stream: %s", streamException.getMessage())))) {
-                            log.error("Failed to convey ERROR status, queue is full");
-                        }
-                        break;
-                    }
+    // TODO: remove the metadata call entirely after https://github.com/open-feature/flagd/issues/1584
+    private static final GetMetadataRequest.Builder metadataRequest = GetMetadataRequest.newBuilder();
+    private Struct getMetadata(Context.CancellableContext context) {
```
guidobrei: Extracted existing `metadataRequest` to dedicated method.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamQueueSource.java
```diff
@@ -205,6 +176,56 @@ private void restart() {
             syncRequest.setProviderId(this.providerId);
         }
 
-        localStub.syncFlags(syncRequest.build(), new QueueingStreamObserver<SyncFlagsResponse>(incomingQueue));
+        localStub.syncFlags(syncRequest.build(), streamObserver);
+
+        while (!shutdown.get() && !streamObserver.context.isCancelled()) {
+            try {
+                Thread.sleep(500);
```
guidobrei: 500ms is an arbitrary number to regularly check if the observer should shut down or the context was cancelled.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamQueueSource.java
```diff
@@ -115,82 +111,57 @@ public void shutdown() throws InterruptedException {
...
+                    syncFlags(observer);
+                } finally {
+                    context.detach(previous);
                 }
+            }
+        }
 
-                // inner loop for handling messages
-                while (!shutdown.get() && !Context.current().isCancelled()) {
-                    final StreamResponseModel<SyncFlagsResponse> taken = incomingQueue.take();
```
guidobrei, line 154: Moved message handling in `SyncStreamObserver` at the end of this class.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamQueueSource.java
```diff
@@ -205,6 +176,56 @@ private void restart() {
             syncRequest.setProviderId(this.providerId);
         }
 
-        localStub.syncFlags(syncRequest.build(), new QueueingStreamObserver<SyncFlagsResponse>(incomingQueue));
+        localStub.syncFlags(syncRequest.build(), streamObserver);
+
+        while (!shutdown.get() && !streamObserver.context.isCancelled()) {
```
chrfwow: Instead of this loop, couldn't we cancel the context in the `shutdown()` method?

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamQueueSource.java
```diff
@@ -115,82 +111,57 @@ public void shutdown() throws InterruptedException {
...
+            log.debug("Metadata exception: {}, retrying", metaEx.getMessage(), metaEx);
+            context.cancel(metaEx);
         }
 
-        log.info("Shutdown invoked, exiting event stream listener");
+        return null;
     }
 
-    private void restart() {
+    private void syncFlags(SyncStreamObserver streamObserver) {
```
chrfwow, line 180: Should we move this method inside the `SyncStreamObserver`?

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamQueueSource.java
```diff
@@ -115,82 +111,57 @@ public void shutdown() throws InterruptedException {
...
-                        log.debug("Exception in stream RPC, streamException {}, will restart", streamException);
-                        if (!outgoingQueue.offer(new QueuePayload(
-                                QueuePayloadType.ERROR,
-                                String.format("Error from stream: %s", streamException.getMessage())))) {
-                            log.error("Failed to convey ERROR status, queue is full");
-                        }
-                        break;
-                    }
+    // TODO: remove the metadata call entirely after https://github.com/open-feature/flagd/issues/1584
+    private static final GetMetadataRequest.Builder metadataRequest = GetMetadataRequest.newBuilder();
```
chrfwow:
```suggestion
    private static final GetMetadataRequest.Builder metadataRequestBuilder = GetMetadataRequest.newBuilder();
```
This should be at the very top of the file

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamQueueSource.java
```diff
@@ -115,82 +111,57 @@ public void shutdown() throws InterruptedException {
...
-                        // cancel the stream if the getMetadata fails
-                        // we can keep this log quiet since the stream is cancelled/restarted with this exception
-                        log.debug("Metadata exception: {}, cancelling stream", metaEx.getMessage(), metaEx);
-                        context.cancel(metaEx);
+            try (Context.CancellableContext context = Context.current().withCancellation()) {
+                SyncStreamObserver observer = new SyncStreamObserver(outgoingQueue, context);
+                Context previous = context.attach();
+                try {
+                    if (!syncMetadataDisabled) {
+                        observer.metadata = getMetadata(context);
```
chrfwow:
```suggestion
                        observer.getMetadata(context);
```
This goes together with my other comment about moving the `syncFlags` method into the `SyncStreamObserver`. I think it would be cleaner to move the functionality into the object.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamQueueSource.java
guidobrei: This method is executed in a thread. `shutdown` is an `AtomicBoolean` to signal shutdown to the stream RPC and the underlying channel connector. But you're right, we maybe want to cancel the stream when we get a shutdown signal.

Will modify the code.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamQueueSource.java
guidobrei: Don't agree. `SyncStreamObserver` implements `StreamObserver<T>` which is a required argument of the gRPC stub's `syncFlags` method.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamQueueSource.java
```diff
@@ -205,6 +176,56 @@ private void restart() {
...
+        public void onError(Throwable throwable) {
+            try {
+                log.error("Stream error: {}, cancelling stream", throwable.getMessage(), throwable);
+                if (!outgoingQueue.offer(new QueuePayload(QueuePayloadType.ERROR, null, null))) {
+                    log.error("Stream writing failed");
+                }
+            } finally {
+                context.cancel(throwable);
+            }
+        }
```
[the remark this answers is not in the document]
guidobrei: Oopsie. Updated to previous behavior.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamQueueSource.java
guidobrei: I got rid of the whole builder by using `GetMetadataRequest.getDefaultInstance()`.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamQueueSource.java
guidobrei: See other comment

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamQueueSource.java
guidobrei: Got rid of Thread.sleep.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamQueueSource.java
```diff
@@ -115,82 +111,54 @@ public void shutdown() throws InterruptedException {
...
-                        // cancel the stream if the getMetadata fails
-                        // we can keep this log quiet since the stream is cancelled/restarted with this exception
-                        log.debug("Metadata exception: {}, cancelling stream", metaEx.getMessage(), metaEx);
-                        context.cancel(metaEx);
-                    }
-                }
+            SyncStreamObserver observer = new SyncStreamObserver(outgoingQueue);
 
-                // inner loop for handling messages
-                while (!shutdown.get() && !Context.current().isCancelled()) {
```
guidobrei, line 153: Was this call to the current context `Context.current().isCancelled()` intentional or should it have been a call to the context `context.isCancelled()`, created before and used in the metadata block above to also skip flag sync when?

`Context.current() != context`, because the call to `Context.current().withCancellation()` only creates a new context, but does not set it as current context.

In the previous version, flags are always synced, even if getMetadata fails. This behavior is different in my implementation. In the new version, when getMatadata fails, we continue the loop and try again.

@toddbaert I saw this change was introduced by you [here](https://github.com/open-feature/java-sdk-contrib/commit/245e9edace570665df16a57a255aea8e6d34d5dd#diff-fb3e83656a66741223dd3dc27d64daa67a08418fc6c95e5388fe704eab7baf90). Any chance you can recall the intention? Based on the log message in the metadata catch block, I'd say we want the syncFlags stream to start over again.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamQueueSource.java
```diff
@@ -115,82 +111,54 @@ public void shutdown() throws InterruptedException {
...
-
-                    if (!outgoingQueue.offer(new QueuePayload(QueuePayloadType.DATA, data, syncContext))) {
-                        log.error("Stream writing failed");
-                    }
-                }
+            try {
+                observer.metadata = getMetadata();
+            } catch (Exception metaEx) {
+                // retry if getMetadata fails
+                log.warn("Metadata request exception, retrying.", metaEx);
```
guidobrei: Not sure about the log level. Elevating from `debug` to `warn` might be too obtrusive.

toddbaert:
This looks like a good simplification. I will thoroughly review on Monday :pray:

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamQueueSource.java
```diff
@@ -205,6 +168,59 @@ private void restart() {
             syncRequest.setProviderId(this.providerId);
         }
 
-        localStub.syncFlags(syncRequest.build(), new QueueingStreamObserver<SyncFlagsResponse>(incomingQueue));
+        localStub.syncFlags(syncRequest.build(), streamObserver);
+
+        streamObserver.done.await();
```
chrfwow, line 197: Do we also have to wakeup `done` when the `shutdown` method is called?

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamQueueSource.java
toddbaert: Spoke to @guidobrei about this - the intent was to kill the loop if the context was cancelled to make sure things are consistent if the metadata call failed, but this was not working expected.

toddbaert:
@guidobrei I pushed a spotless fix. Doing some manual testing now.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamQueueSource.java
toddbaert: No, because when we shutdown, we call `shutdownNow` on the underlying channel - when this happens, the `onError` handler runs immediately and wakes it up.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamQueueSource.java
toddbaert: Now I think we want to know about these.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamQueueSource.java
```diff
@@ -114,83 +102,58 @@ public void shutdown() throws InterruptedException {
...
+                log.debug("Metadata request error: {}, will restart", message, metaEx);
+                enqueueError(String.format("Error in getMetadata request: %s", message));
+                continue;
+            }
+
+            try {
+                syncFlags(observer);
+            } catch (Exception ex) {
+                log.error("Unexpected sync stream exception, will restart.", ex);
+                enqueueError(String.format("Error in getMetadata request: %s", ex.getMessage()));
```
toddbaert:
```suggestion
                enqueueError(String.format("Error in syncStream: %s", ex.getMessage()));
```

chrfwow (approved):

toddbaert:
@guidobrei There's some e2e tests around context enrichment that are consistently failing since your change:

```
Error:  Errors: 
Error:    RunInProcessTest.Resolve value independent of context » General Initialization timeout exceeded; did not complete within the 2000 ms deadline.
Error:  dev.openfeature.contrib.providers.flagd.e2e.RunInProcessTest.Use enriched context
[INFO]   Run 1: PASS
Error:    Run 2: RunInProcessTest.Use enriched context » General Initialization timeout exceeded; did not complete within the 2000 ms deadline.
[INFO] 
Error:  dev.openfeature.contrib.providers.flagd.e2e.RunInProcessTest.Use enriched context on connection error for IN-PROCESS
[INFO]   Run 1: PASS
Error:    Run 2: RunInProcessTest.Use enriched context on connection error for IN-PROCESS » General Initialization timeout exceeded; did not complete within the 2000 ms deadline.
```

Other than this, things look great. I did some manual testing as well and reconnect, events, etc works fine!

EDIT: I think I found the reason [here](https://github.com/open-feature/java-sdk-contrib/pull/1590#discussion_r2372756450).

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamQueueSource.java
```diff
@@ -114,83 +102,58 @@ public void shutdown() throws InterruptedException {
...
+            SyncStreamObserver observer = new SyncStreamObserver(outgoingQueue);
+
+            try {
+                observer.metadata = getMetadata();
+            } catch (Exception metaEx) {
+                // retry if getMetadata fails
+                String message = metaEx.getMessage();
+                log.debug("Metadata request error: {}, will restart", message, metaEx);
+                enqueueError(String.format("Error in getMetadata request: %s", message));
+                continue;
```
toddbaert: 99% sure this is causing our test failure. @aepfli introduced [this option](https://github.com/open-feature/flagd/blob/aa504f7077093746f886248a4766d9ae5587bf3d/flagd/pkg/service/flag-sync/handler.go#L110) in flagd to help us test the migration from the old `getMetadata` call to the new `syncContext` payload - in some e2e tests, flagd runs with the old `getMetadata` disabled... which is causing an error here immediately, even though we have the new `syncContext`.

We need to "save" this exception state, and only throw if we DON'T get the new alternative `syncContext`.

Does that make sense?

@aepfli please verify what I said :pray:

guidobrei: Either this or we check for [gRPC Status code](https://grpc.io/docs/guides/status-codes/#the-full-list-of-status-codes) `12` `UNIMPLEMENTED`, or `5` `NOT_FOUND`.

Relying on the status code would be the easier implementation.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamQueueSource.java
guidobrei: Exactly, that was the intention.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamQueueSource.java
aepfli: i think the idea with checking the error code is also a good idea based on https://github.com/open-feature/flagd/blob/fcd19b310b319c9837b41c19d6f082ac750cb81d/flagd/pkg/service/flag-sync/handler.go#L111 it is unimplemented

toddbaert: You can get UNIMPLEMENTED for other cases, like missing implementations - so I don't think that will work.

guidobrei: But isn't this what we want to have? If the implementation is missing or disabled, we use whatever we get via the `syncFlags()` stream.

guidobrei: Added an example of how this could look like in [this commit](https://github.com/open-feature/java-sdk-contrib/pull/1590/commits/f2092b999a504f3b0187c352826979a4bdee0847).

toddbaert: Yes I think you're right! lol sorry, I confused myself.

toddbaert: @guidobrei The test is still failing, but if I make the catch block more general, it works. I think either the status code can't be checked the way you're checking it, or the exception is not of the expected type.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/common/ChannelBuilder.java
```diff
@@ -26,17 +26,38 @@
 /** gRPC channel builder helper. */
 public class ChannelBuilder {
 
+    @SuppressWarnings({"unchecked", "rawtypes"})
+    private static final Map<String, ?> DEFAULT_RETRY_POLICY = new HashMap() {
+        {
+            // 1 + 2 + 4
```
chrfwow: What does that comment mean? The number 7 does not show up in the values below

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/common/ChannelBuilder.java
```diff
@@ -46,43 +67,56 @@ public class ChannelBuilder {
...
+                                    new HashMap() {
+                                        {
+                                            put("service", "flagd.sync.v1.FlagSyncService");
+                                            put("method", "SyncFlags");
+                                        }
+                                    }
+                                ));
+                            put("retryPolicy", new HashMap(DEFAULT_RETRY_POLICY) {
+                                {
+                                    // 1 + 2 + 4 + 5 + 5 + 5 + 5 + 5 + 5 + 5
```
chrfwow: What does that comment mean? The result does not show up in the values below

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/common/ChannelBuilder.java
```diff
@@ -46,43 +67,56 @@ public class ChannelBuilder {
...
+                                    put("retryableStatusCodes",
+                                            Arrays.asList(
+                                                    /*
+                                                     * All codes are retryable except OK and DEADLINE_EXCEEDED since
+                                                     * any others not listed here cause a very tight loop of retries.
+                                                     * DEADLINE_EXCEEDED is typically a result of a client specified
+                                                     * deadline,and definitionally should not result in a tight loop
+                                                     * (it's a timeout).
+                                                     */
+                                                    Code.CANCELLED.toString(),
```
chrfwow: Why do we have here so many more status codes than in the other policy?

aepfli (approved):

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamQueueSource.java
guidobrei: Today I learned: https://github.com/open-feature/java-sdk-contrib/pull/1590/commits/0d6c840b9ddee20f2e5a4825d854535d65364ffc

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/common/ChannelBuilder.java
guidobrei: We use the GRPC retry policy in sync flags to not end up in a tight reconnect loop and try to reconnect instantly.

guidobrei:
We've done some manual testing with @aepfli and @chrfwow. The retry policy seems to be broken when connecting via Envoy and Envoy loses the connection to flagd. 

> [!WARNING]
>   In this situation we end up in a tight retry loop. This needs to be fixed.

- [x] Fix retry policies

toddbaert:
> We've done some manual testing with @aepfli and @chrfwow. The retry policy seems to be broken when connecting via Envoy and Envoy loses the connection to flagd.
> 
> Warning
> 
> In this situation we end up in a tight retry loop. This needs to be fixed.
> 
> * [ ]  Fix retry policies

It must be a result of the retry policy changes, right? Should we just avoid changes there for now?

I'm not super happy with it, but I think it's distinct from the deadlock bug you've found.

guidobrei:
> It must be a result of the retry policy changes, right? Should we just avoid changes there for now?

If we don't change the retry policy, can't check for "UNIMPLEMENTED", because we would retry and in the end fail with a DEADLINE_EXCEEDED.

toddbaert:
> We've done some manual testing with @aepfli and @chrfwow. The retry policy seems to be broken when connecting via Envoy and Envoy loses the connection to flagd.
> 
> Warning
> 
> In this situation we end up in a tight retry loop. This needs to be fixed.
> 
> * [ ]  Fix retry policies

Is this only true for the changed introduce in this PR?

> > It must be a result of the retry policy changes, right? Should we just avoid changes there for now?
> 
> If we don't change the retry policy, can't check for "UNIMPLEMENTED", because we would retry and in the end fail with a DEADLINE_EXCEEDED.

ah, right, I forgot how we got here. We could just remove retry for that one though?

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/common/ChannelBuilder.java
```diff
@@ -52,33 +58,17 @@ public class ChannelBuilder {
...
                                     Arrays.asList(
                                             /*
-                                             * All codes are retryable except OK and DEADLINE_EXCEEDED since
-                                             * any others not listed here cause a very tight loop of retries.
-                                             * DEADLINE_EXCEEDED is typically a result of a client specified deadline,
-                                             * and definitionally should not result in a tight loop (it's a timeout).
+                                             * As per gRPC spec, the following status codes are safe to retry:
+                                             * UNAVAILABLE, UNKNOWN,
                                              */
-                                            Code.CANCELLED.toString(),
```
toddbaert, line 67: Removed - these were only here to slow down our loop, which we do with a sleep now.

toddbaert:
@guidobrei I pushed one small change which:

- simplifies our grpc retry config to make it more sensible, and use a `Thread.sleep` instead to slow our busy loop
- associated test to make sure we don't busy loop on connection errors

toddbaert (approved):

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdOptions.java
```diff
@@ -97,10 +97,18 @@ public class FlagdOptions {
     @Builder.Default
     private int deadline = fallBackToEnvOrDefault(Config.DEADLINE_MS_ENV_VAR_NAME, Config.DEFAULT_DEADLINE);
 
+    /**
+     * Max stream retry backoff in milliseconds.
+     */
+    @Builder.Default
+    private int retryBackoffMaxMs =
+            fallBackToEnvOrDefault(Config.FLAGD_RETRY_BACKOFF_MAX_MS_VAR_NAME, Config.DEFAULT_MAX_RETRY_BACKOFF);
```
toddbaert: This is already a defined param we never implemented, so good to have now.

We use this also for the sleep time for non-transient errors.

chrfwow (approved):

alexandraoberaigner (approved):

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdOptions.java
```diff
@@ -98,10 +98,18 @@ public class FlagdOptions {
...
```
guidobrei: We're mixing here `ms` and `s`.

Is `static final int DEFAULT_MAX_RETRY_BACKOFF = 5;` a time unit or count?

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/Config.java
```diff
@@ -14,6 +14,7 @@ public final class Config {
     static final String DEFAULT_HOST = "localhost";
 
     static final int DEFAULT_DEADLINE = 500;
+    static final int DEFAULT_MAX_RETRY_BACKOFF = 5;
```
toddbaert:
```suggestion
    static final int DEFAULT_MAX_RETRY_BACKOFF = 12000;
```

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdOptions.java
toddbaert: It's a time. I will make the default name more clear. It should also be 12000 as per https://flagd.dev/providers/rust/?h=flagd_retry_backoff_max_ms#configuration-options... nice catch.... this was leftover from my initial version before I realized we had an unimplemented var for this.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamQueueSource.java
```diff
@@ -114,84 +109,76 @@ public void shutdown() throws InterruptedException {
...
+                log.debug("Initializing sync stream request");
+                SyncStreamObserver observer = new SyncStreamObserver(outgoingQueue);
+                try {
+                    observer.metadata = getMetadata();
+                } catch (Exception metaEx) {
+                    // retry if getMetadata fails
+                    String message = metaEx.getMessage();
+                    log.debug("Metadata request error: {}, will restart", message, metaEx);
+                    enqueueError(String.format("Error in getMetadata request: %s", message));
+                    Thread.sleep(this.maxBackoffMs);
```
toddbaert, line 129: this sleep helps prevent tight loops during retries, which can be a problem if a upstream connection in a proxy is down.

PR #1595, opened 2025-09-19
fix(go-feature-flag): fix loading wasm files in a jar mode.

thomaspoignant:
Before this pull request the loading of the wasm module was not working when packaged in a jar. 
The actual v1.0.0 is not able to load the wasm file.


thomaspoignant:
I will merge this PR without waiting for a review because the current released version of this morning is not working and I want to have a working version ASAP on mvn repository.

PR #1624, opened 2025-11-10
feat(flagd): introduce fatalStatusCodes option

leakonvalinka:
## This PR

- adds the fatalStatusCode option + env variable

### Related Issues

resulted from [this issue](https://github.com/open-feature/go-sdk-contrib/issues/756) 

### Notes

I'm not too happy with how the fatal error is communicated through the different components (received at `SyncStreamQueueSource` -> `FlagStore` -> `InProcessResolver` -> `FlagdProvider`, respective `RpcResolver` -> `FlagdProvider`). It "misuses" the STALE state to differentiate between normal errors and fatal errors. I couldn't find a cleaner solution for this unfortunately, so feedback on this would be highly appreciated!

Will work on the remaining failing tests once we agree on how to proceed!


providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/steps/config/ConfigSteps.java
```diff
@@ -121,6 +121,7 @@ private static String mapOptionNames(String option) {
         propertyMapper.put("keepAliveTime", "keepAlive");
         propertyMapper.put("retryBackoffMaxMs", "keepAlive");
         propertyMapper.put("cache", "cacheType");
+        propertyMapper.put("fatalStatusCodes", "nonRetryableStatusCodes");
```
aepfli: please also adapt https://github.com/open-feature/flagd-testbed/blob/main/gherkin/config.feature - it seems like there is currently no check for the proper configuration, or am i wrong here?

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/steps/ProviderSteps.java
```diff
@@ -190,4 +199,9 @@ public void the_flag_was_modded() {
                 .then()
                 .statusCode(200);
     }
+
+    @Then("the client is in {} state")
```
aepfli: we should do the renaming of this step as described here https://github.com/open-feature/flagd-testbed/issues/306 before merging

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/steps/Utils.java
```diff
@@ -37,6 +38,8 @@ public static Object convert(String value, String type) throws ClassNotFoundExce
                 }
             case "CacheType":
                 return CacheType.valueOf(value.toUpperCase()).getValue();
+            case "StringList":
+                return List.of(value);
```
aepfli: Not sure about the structure of the step, but i would assume, that there can be multiple values within the string, and that we maybe should split and trim it, before creating a list.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdOptions.java
```diff
@@ -122,6 +123,14 @@ public class FlagdOptions {
     @Builder.Default
     private int retryGracePeriod =
             fallBackToEnvOrDefault(Config.STREAM_RETRY_GRACE_PERIOD, Config.DEFAULT_STREAM_RETRY_GRACE_PERIOD);
+
+    /**
+     * List of grpc response status codes for which failed connections are not retried.
+     * Defaults to empty list
+     */
+    @Builder.Default
+    private List<String> nonRetryableStatusCodes = new ArrayList<>();
```
aepfli: As within the documentation the option has a dedicated name, we should keep to the specification, so this property has the same name in all the providers.

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/steps/config/ConfigSteps.java
aepfli: i am on it :) waiting for my connection flight, so i will try to do this one :)

aepfli: https://github.com/open-feature/flagd-testbed/pull/311

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/steps/Utils.java
leakonvalinka: Yes, i also think so, but I wasn't sure of what the separator to use here. Have we agreed on one?

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdOptions.java
leakonvalinka: Yup, makes sense.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamQueueSource.java
```diff
@@ -47,6 +49,7 @@ public class SyncStreamQueueSource implements QueueSource {
     private final BlockingQueue<QueuePayload> outgoingQueue = new LinkedBlockingQueue<>(QUEUE_SIZE);
     private final FlagSyncServiceStub flagSyncStub;
     private final FlagSyncServiceBlockingStub metadataStub;
+    private final List<String> fatalStatusCodes;
```
chrfwow, line 50: Since we do lots of `.contains` operation on this data structure, a `HashSet` might be more performant. How many entries do we expect in this list?

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/rpc/RpcResolver.java
```diff
@@ -348,13 +352,20 @@ private void observeEventStream() throws InterruptedException {
...
+                    } else {
+                        log.debug(
+                                "Exception in event stream connection, streamException {}, will reconnect",
+                                streamException);
+                        this.handleErrorOrComplete();
+                    }
                     break;
                 }
 
+                successfulConnection.set(true);
```
chrfwow, line 380: Do/Should we ever reset `successfulConnection` to `false`?

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/Config.java
```diff
@@ -91,6 +95,16 @@ static long fallBackToEnvOrDefault(String key, long defaultValue) {
         }
     }
 
+    static List<String> fallBackToEnvOrDefaultList(String key, List<String> defaultValue) {
+        try {
+            return System.getenv(key) != null ? Arrays.stream(System.getenv(key).split(","))
+                    .map(String::trim)
+                    .collect(Collectors.toList()) : defaultValue;
+        } catch (Exception e) {
+            return defaultValue;
```
chrfwow, line 108: We should print an info/warn that the env vars are invalid

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdOptions.java
```diff
@@ -122,6 +123,14 @@ public class FlagdOptions {
     @Builder.Default
     private int retryGracePeriod =
             fallBackToEnvOrDefault(Config.STREAM_RETRY_GRACE_PERIOD, Config.DEFAULT_STREAM_RETRY_GRACE_PERIOD);
+
+    /**
+     * List of grpc response status codes for which failed connections are not retried.
+     * Defaults to empty list
+     */
+    @Builder.Default
+    private List<String> fatalStatusCodes = fallBackToEnvOrDefaultList(Config.FATAL_STATUS_CODES_ENV_VAR_NAME, List.of());
```
chrfwow: Do we really want to retry on every error code per default? How is this handled in our other sdks?

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
```diff
@@ -222,20 +223,25 @@ private void onProviderEvent(FlagdProviderEvent flagdProviderEvent) {
                     onReady();
                     syncResources.setPreviousEvent(ProviderEvent.PROVIDER_READY);
                     break;
-
-                case PROVIDER_ERROR:
-                    if (syncResources.getPreviousEvent() != ProviderEvent.PROVIDER_ERROR) {
-                        onError();
-                        syncResources.setPreviousEvent(ProviderEvent.PROVIDER_ERROR);
+                case PROVIDER_STALE:
```
chrfwow: Please update the javadoc above the `switch`, we do now use the `STALE` state

chrfwow (commented):
Since we do not want to introduce breaking changes into the api by adding a `PROVIDER_FATAL` type to `ProviderEvent`, I have two suggestions how we might be able to work around the "misuse" of the stale event:
We could add a `isFatal` flag to the `FlagdProviderEvent` to track the type of error. I don't really like it because this could also be set when the event is not an error event, and with this we split up information that should be stored in one place into two places.
Or, we create an enum class `ExtendedProviderEvent`, which is a copy of `ProviderEvent` (enums cannot be extended in Java), plus the additional `PROVIDER_FATAL` field. We would then have to map where needed between the two types (not 100% sure if this will work). I don't like this either, because we would duplicate the `ProviderEvent` enum

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/rpc/RpcResolver.java
leakonvalinka: On the issue that started this, the [conclusion](https://github.com/open-feature/go-sdk-contrib/issues/756#issuecomment-3552081857) was that we only check if an error code is considered fatal in the first cycle. So no, I don't think it should ever be set back to false.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/Config.java
leakonvalinka: Just for this method? Or the other ones too? I'd either leave it or add it in all cases to be consistent

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamQueueSource.java
leakonvalinka: That's hard for me to estimate, what do the others think? The [currently defined default](https://github.com/open-feature/flagd/pull/1818/changes#diff-28ba4b4041b6c39c4b9deb542276312bebdd08f45b39b29d860849d9e8b81a6eR316) is an empty list

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdOptions.java
leakonvalinka: No, you're right, I'll rephrase it to "for which the provider transitions into fatal mode upon first connection". The general retry policy is defined [here](https://github.com/open-feature/java-sdk-contrib/blob/main/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/common/ChannelBuilder.java#L33) and is the same for all sdks afaik

guidobrei:
> I'm not too happy with how the fatal error is communicated through the different components (received at SyncStreamQueueSource -> FlagStore -> InProcessResolver -> FlagdProvider...)

This is an implication of our provider design and there is not really something to do about that (in this PR).

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
leakonvalinka: Will do once we agree on a final implementation

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/Config.java
chrfwow: Then we should add it everywhere, but in a different PR

leakonvalinka: Alright, sounds good. Should we create a new issue for this or is that overkill?

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/FlagStore.java
```diff
@@ -138,6 +138,11 @@ private void streamerListener(final QueueSource connector) throws InterruptedExc
                     }
                     break;
                 case ERROR:
+                    if (!stateBlockingQueue.offer(new StorageStateChange(StorageState.STALE))) {
```
toddbaert, line 142: Could we simply add a `FATAL` storage state to resolve this conceptual "STALE" overloading? This is an entirely private enum, so we can add to it without issue.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamQueueSource.java
```diff
@@ -37,6 +38,7 @@ public class SyncStreamQueueSource implements QueueSource {
 
     private final AtomicBoolean shutdown = new AtomicBoolean(false);
     private final AtomicBoolean shouldThrottle = new AtomicBoolean(false);
+    private final AtomicBoolean successfulSync = new AtomicBoolean(false);
```
toddbaert: My biggest question with this whole concept (not your implementation) is whether or not we should care about whether this is the initial sync or not. I'm actually leaning towards "not"... here is my reasoning (anyone feel free to disagree):

- users already fully control what's considered `FATAL`; they can also control whether or not to consider FATAL at init different than FATAL later, using event handlers and the details of the exception
- it's simpler (less conditions/state to handler in our code [this field would disappear] and for users to understand)

WDYT?

toddbaert:
> Since we do not want to introduce breaking changes into the api by adding a `PROVIDER_FATAL` type to `ProviderEvent`, I have two suggestions how we might be able to work around the "misuse" of the stale event: We could add a `isFatal` flag to the `FlagdProviderEvent` to track the type of error. I don't really like it because this could also be set when the event is not an error event, and with this we split up information that should be stored in one place into two places. Or, we create an enum class `ExtendedProviderEvent`, which is a copy of `ProviderEvent` (enums cannot be extended in Java), plus the additional `PROVIDER_FATAL` field. We would then have to map where needed between the two types (not 100% sure if this will work). I don't like this either, because we would duplicate the `ProviderEvent` enum

I could be missing something, but I don't think this is an issue. The "fatalness" (fatality?) of an event is not communicated by the event type, but the error code associated with the event: https://github.com/open-feature/java-sdk/blob/main/src/main/java/dev/openfeature/sdk/ProviderEventDetails.java#L16. All error events are events. Some error events contain with `FATAL` error codes.

This is basically what's in Go, as well, right @alexandraoberaigner ?

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/FlagStore.java
leakonvalinka: Yes, we could, but this would only solve the missuse issue in the communication step from `FlagStore` -> `InProcessResolver` (and not `InProcessResolver` -> `FlagdProvider`)
Also, nit pick: `StorageState.ERROR` is already defined as `/** Storage is in an unrecoverable error stage. */`, which models what FATAL means for us i think

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamQueueSource.java
leakonvalinka: If it's easy to get the same behaviour through event handlers, I think that might be better, because it allows for more customization. I get both sides, that one might not want to completely shut down if a valid flag config was previously received, but also that one might not want to work with stale data given that a non-transient error was received

leakonvalinka:
> I could be missing something, but I don't think this is an issue. The "fatalness" (fatality?) of an event is not communicated by the event type, but the error code associated with the event: https://github.com/open-feature/java-sdk/blob/main/src/main/java/dev/openfeature/sdk/ProviderEventDetails.java#L16. All error events are events. Some error events contain with FATAL error codes.

The issue is that the [`onProviderEvent` method](https://github.com/open-feature/java-sdk-contrib/blob/ba277bf21dc3f4476bae2846a821b4655ede899f/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java#L195) expects a `FlagdProviderEvent`, which doesn't allow a differentiation between different error types (based on the [`ProviderEvent` enum](https://github.com/open-feature/java-sdk/blob/main/src/main/java/dev/openfeature/sdk/ProviderEvent.java)), but we need to tell apart "normal" errors from fatal ones. My implementation does this by its `STALE` misuse, which isn't a super clean approach. @chrfwow's suggestions would be 2 alternatives, but also not 100% clean.

Personally I think I prefer adding a `FATAL` to the `StorageState` enum (even though [this nit pick](https://github.com/open-feature/java-sdk-contrib/pull/1624#discussion_r2634046866) here) and adding an `isFatal` flag to the `FlagdProviderEvent`, but as this also has its disadvantages, so I'd like to hear other opinions as well! :)

alexandraoberaigner:
> I could be missing something, but I don't think this is an issue. The "fatalness" (fatality?) of an event is not communicated by the event type, but the error code associated with the event: https://github.com/open-feature/java-sdk/blob/main/src/main/java/dev/openfeature/sdk/ProviderEventDetails.java#L16. All error events are events. Some error events contain with FATAL error codes.
This is basically what's in Go, as well, right @alexandraoberaigner ?

Yes, in Go its a provider event with error code: see [ProviderInitError](https://github.com/open-feature/go-sdk/blob/4b3d6d953076f3056d78c512b31c9fc2ec29ab89/openfeature/resolution_error.go#L142) with [ErrorCode](https://github.com/open-feature/go-sdk/blob/4b3d6d953076f3056d78c512b31c9fc2ec29ab89/openfeature/resolution_error.go#L8) including PROVIDER_FATAL

Just found a test in the java-sdk that says how the fatal error should look like: see [here](https://github.com/open-feature/java-sdk/blob/0b4325a830997840ae3b622df93ca720e73a9088/src/test/java/dev/openfeature/sdk/FeatureProviderStateManagerTest.java#L149-L151)

```
ProviderEvent.PROVIDER_ERROR,
ProviderEventDetails.builder()
                        .errorCode(ErrorCode.PROVIDER_FATAL)
                        .build()
```

leakonvalinka:
> Just found a test in the java-sdk that says how the fatal error should look like: see [here](https://github.com/open-feature/java-sdk/blob/0b4325a830997840ae3b622df93ca720e73a9088/src/test/java/dev/openfeature/sdk/FeatureProviderStateManagerTest.java#L149-L151)
>```
>ProviderEvent.PROVIDER_ERROR,
>ProviderEventDetails.builder()
>                        .errorCode(ErrorCode.PROVIDER_FATAL)
>                        .build()
>```

Yup, I also implemented it like this when the FlagdProvider emits the fatal error [here](https://github.com/open-feature/java-sdk-contrib/pull/1624/changes#diff-7ac82836e0450dd1f6581536670fba629b6fa21a8ccb4a5ffcdce54a53dd57e3R241)
But that doesnt change anything regarding the question how to cleanly propagate the error through the layers until it reaches the `FlagdProvider` class, unless I am missing something

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/FlagStore.java
guidobrei: Based on the [StorageState existing docs](https://github.com/open-feature/java-sdk-contrib/blob/main/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/StorageState.java) the states are used now as intended:

> `STALE`: Storage has gone stale (most recent sync failed). May get to OK status with next sync.

`QueueSource` encountered an `ERROR` but will try to recover.

> `ERROR`:  Storage is in an unrecoverable error stage.

`QueueSource` encountered an `ERROR` **AND** will not try to recover. It exited the sync loop.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamQueueSource.java
guidobrei: I agree with @toddbaert that fatality does not depend on being the first sync or not.

However, *fatality* from the `SyncStreamQueueSource`'s perspective depends on whether we break out the sync loop and stop the sync thread (`FATAL ERROR`), or continue to reconnect (`ERROR`).

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/StorageState.java
```diff
@@ -1,11 +1,11 @@
...
 public enum StorageState {
     /** Storage is upto date and working as expected. */
     OK,
-    /** Storage has gone stale(most recent sync failed). May get to OK status with next sync. */
-    STALE,
+    /** Storage has gone stale (most recent sync failed). May get to OK status with next sync. */
+    TRANSIENT_ERROR,
     /** Storage is in an unrecoverable error stage. */
-    ERROR,
+    FATAL_ERROR,
```
aepfli: [nit] To be honest, I do not agree with renaming those state adding any benefit to this implementation. The changeset of this pull request is unecessarily "increasing", due to this renames. Furthermore, those renames on the side make it harder and harder to grasp the whole change, because more and more files are touched. Additionally, the known behaviour of the code has shifted, for people familiar witht he code base, renaming the enums and values contain a certain cognitive load, which I do not see the benefit currently in the renaming.

toddbaert: @guidobrei @leakonvalinka and I all did this together in a group programming session, at least we thought it was an improvement.

I don't care _that_ much about it, TBH, but since it's subjective I would go with the majority.

We could also revert it here and do it in a separate PR, because we did find other naming issues, which we thought we should not add here to reduce thrashing.

toddbaert: I will just revert this and we can open another PR for pure name changes and argue about naming there :)

toddbaert: Reverted

toddbaert (approved):
I made a few small test fixes.

Looks good to me, and FATAL works as expected. One thing to note is that as discussed in various places, we DO NOT act differently on FATAL status codes depending on whether this is the initial connection or not; if the stream sends a FATAL code at any time, the provider transitions to FATAL.

See: https://github.com/open-feature/flagd/pull/1818#discussion_r2598681196

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/FlagStore.java
toddbaert: We fixed this by propagating the provider details as well though the consumer, so we don't need to overload or abuse the enums anymore.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java
toddbaert: I don't think this is relevant with our latest solution.

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/steps/ProviderSteps.java
```diff
@@ -200,6 +202,9 @@ public void the_flag_was_modded() {
 
     @Then("the client should be in {} state")
     public void the_client_should_be_in_fatal_state(String clientState) {
-        assertThat(state.client.getProviderState()).isEqualTo(ProviderState.valueOf(clientState.toUpperCase()));
+        await().pollDelay(100, TimeUnit.MILLISECONDS)
```
toddbaert: I had to make this test tolerate a small retry/timeout here (it worked most of the time but was flaky for both in-process and RPC)

I think that there's a small lag time between when we fire events, and when the SDK updates the client status. We should consider whether this is a small bug with the SDK, and if we should guarantee that the state is updated before the event handlers run (or not)

@guidobrei @aepfli @leakonvalinka

guidobrei: The [EventProvider](https://github.com/open-feature/java-sdk/blob/main/src/main/java/dev/openfeature/sdk/EventProvider.java#L95-L100) first changes the state and then emits the event, i.e. runs the event handlers.

Not sure it this is a bug in the SDK or a side-effect of the Gherkin test:
```
  Scenario: Provider forbidden
    Given ...
    Then the error event handler should have been executed within 5000ms
    And the client should be in fatal state
```
If the *Then X And Y* expressions are executed in parallel, this can lead to the described race condition.

@aepfli do you know if this is the case? Or will the *And the client should be in fatal state* be verified after the event handler execution is checked?

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/Config.java
guidobrei: https://github.com/open-feature/java-sdk-contrib/issues/1673

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamQueueSource.java
guidobrei: I agree with @chrfwow, but I don't think the use-case is not performance critical. Maybe we add this in a follow up PR?

guidobrei (approved):

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/steps/ProviderSteps.java
toddbaert: Great point! None of the documentation I can find indicates that `Then` and `And` might run in parallel on a single feature though. Everything I see is mentioning running multiple features at the same time, but don't mention parallelism within a single test/feature :thinking: 

https://cucumber.io/docs/guides/parallel-execution/

https://github.com/cucumber/cucumber-jvm/tree/main/cucumber-junit-platform-engine

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/StorageState.java
aepfli: imho currently we do not know if anybody is also using those enums. so changing them would be a breaking change, and we would need to change the pr title

toddbaert (approved):

toddbaert:
There seems to be some flakiness with the e2e suite, seemingly around the testbed wrapper: 

```
Error:  Failures: 
Error:    RunRpcTest.Flag change event 1 expectation failed.
Expected status code <200> but was <500>.

Error:    RunRpcTest.Flag change event with caching 1 expectation failed.
Expected status code <200> but was <500>.

Error:    RunRpcTest.Use enriched context on RPC connection will not cache the value 1 expectation failed.
Expected status code <200> but was <500>.
```

https://github.com/open-feature/java-sdk-contrib/actions/runs/20664429787/job/59335972841?pr=1624

I think we are calling the /restart /config methods, etc sometimes and getting 500s instead of 200s

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/steps/ProviderSteps.java
leakonvalinka: Should I create a separate issue for this or do we simply accept the current solution as it is? Might be worth looking into it, but also doesn't seem like a high-priority problem right now.

toddbaert:
@leakonvalinka ... I kept seeing intermittent test failures with this PR ([this](https://github.com/open-feature/java-sdk-contrib/pull/1624#discussion_r2641110137) was the biggest example. At first, I thought it was a race-condition in the test itself, but it was not). I have removed this now, because I found the root cause:

The main issue was [here](https://github.com/open-feature/java-sdk-contrib/pull/1624/changes#diff-5651aa7cc591c468780331739eeb2a6316c48cfd11b16479c3608fea1a5d6f34R95-R99). It was not any logical error with the code you added, but your changes surfaced an existing issue. The race condition works like this:

- the `FlagdProvider`'s `initialize` function is called, which calls `flagResolver.init()`... the method then blocks waiting for `syncResources.waitForInitialization(...)`
- normally, in error situations, this method then times out (another thread runs error handlers)
- however, in our situation, the `syncResources.waitForInitialization(...)` method doesn't just time out; due to our changes to shut down when FATAL errors are seen, the method starting throwing `IllegalStateException` (see the condition at the end of the method)
- since this is not a FATAL error, this was thrown by init, and depending on timing, would "override" the FATAL with a non-fatal error (the `IllegalStateException`)

My fix to throw the Fatal error depending on the last event is a bandaid fix, because this PR has already been open for a while.

To really fix the issue, we should improve the way the waiting in the init works, and actually capture the error during init, and not wait for the full timeout every time. I will open a new issue for this.

There was also this totally unrelated bug in the test-harness which was causing some failures as well: https://github.com/open-feature/flagd-testbed/pull/321

EDIT: I've opened the new issue: https://github.com/open-feature/java-sdk-contrib/issues/1675

cc @guidobrei @aepfli

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/steps/ProviderSteps.java
toddbaert: See [here](https://github.com/open-feature/java-sdk-contrib/pull/1624#issuecomment-3715523407). I will create a new issue.

toddbaert: See [here](https://github.com/open-feature/java-sdk-contrib/pull/1624#issuecomment-3715523407). I will create a new issue.

toddbaert: See [here](https://github.com/open-feature/java-sdk-contrib/pull/1624#issuecomment-3715523407). I will create a new issue.

PR #1626, opened 2025-11-12
fix(gofeatureflag): Added Locale.ENGLISH to SimpleDateFormat for parsing the Last-Modified HTTP header

kuaikuai:
Added Locale.ENGLISH to SimpleDateFormat for parsing the Last-Modified HTTP header. This makes the date parsing locale - specific to English, avoiding possible parsing issues caused by default locale differences.


thomaspoignant:
@kuaikuai thanks for this PR, there is a formatter issue that blocks the CI could you have a look?

Also I've seen the gemini review and this implementation is probably better to what I've written.
If you have some time and willing to contribute to the provider by changing that it would be awesome.

If you can't this is not a problem, you can just fix the formatting part, and we can release a new version of the provider, and I will work on updating the way to read the headers later.

kuaikuai:
> @kuaikuai thanks for this PR, there is a formatter issue that blocks the CI could you have a look?
> 
> Also I've seen the gemini review and this implementation is probably better to what I've written. If you have some time and willing to contribute to the provider by changing that it would be awesome.
> 
> If you can't this is not a problem, you can just fix the formatting part, and we can release a new version of the provider, and I will work on updating the way to read the headers later.

I submitted the code directly through the GitHub web interface, so it’s not easy to ensure proper formatting. I’ll make some adjustments to fix this. ^_^

kuaikuai:
`
The DateTimeFormatter.RFC_1123_DATE_TIME in Java does not use ResolverStyle.STRICT by default.

If strict parsing is required , simply using DateTimeFormatter.RFC_1123_DATE_TIME as-is may not be sufficient. 

This means it may accept some lenient or ambiguous inputs that strictly conforming RFC 1123 date-time strings should reject.
Unfortunately, you cannot directly modify the resolver style of the predefined RFC_1123_DATE_TIME formatter, so achieving strict parsing requires additional work — such as creating a custom formatter based on it with ResolverStyle.STRICT.
It is more complex than SimpleDateFormat("EEE, dd MMM yyyy HH:mm:ss zzz", Locale.ENGLISH).

SimpleDateFormat can be sufficient for specific use cases — especially when dealing with HTTP headers (like Last-Modified) where the format is well-known and consistently formatted per RFC 1123.
`

thomaspoignant (approved):

PR #1629, opened 2025-11-14
fix: possible tight busy loop on certain connection errors

toddbaert:
With https://github.com/open-feature/java-sdk-contrib/pull/1590, we fixed a few issues; one was a possible busy loop on certain connection errors. Unfortunately we missed the case that the error actually occurs on the stream (`onError`). This applies the same backoff in that case, and adds an associated test.

I tested this fix manually as well, with a Spring app and a flagd hard-coded to return errors.

Thanks @leakonvalinka and @guidobrei for finding this.

### DIFF
--- a/providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamQueueSourceTest.java
+++ b/providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamQueueSourceTest.java
@@ -36,7 +36,8 @@ class SyncStreamQueueSourceTest {
     private ChannelConnector mockConnector;
     private FlagSyncServiceBlockingStub blockingStub;
     private FlagSyncServiceStub stub;
-    private FlagSyncServiceStub errorStub;
+    private FlagSyncServiceStub syncErrorStub;
+    private FlagSyncServiceStub asyncErrorStub;
     private StreamObserver<SyncFlagsResponse> observer;
     private CountDownLatch latch; // used to wait for observer to be initialized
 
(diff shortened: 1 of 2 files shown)


providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamQueueSourceTest.java
```diff
@@ -60,25 +61,76 @@ public void setup() throws Exception {
...
+                            })
+                            .start();
+
+                    return null;
+                })
+                .when(asyncErrorStub)
+                .syncFlags(any(SyncFlagsRequest.class), any(StreamObserver.class)); // Mock the initialize
+    }
+
+    @Test
```
toddbaert, line 99: Git is rendering this oddly; this is the previous unchanged test (other than the name). The new one is below.

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamQueueSourceTest.java
```diff
@@ -60,25 +61,76 @@ public void setup() throws Exception {
...
+
+        // should have retried the stream (2 calls); initial + 1 retry
+        // it's very important that the retry count is low, to confirm no busy-loop
+        verify(syncErrorStub, times(2)).syncFlags(any(), any());
     }
 
     @Test
-    void initError_DoesNotBusyWait() throws Exception {
-        // make sure we do not spin in a busy loop on errors
+    void asyncInitError_DoesNotBusyWait() throws Exception {
```
toddbaert, line 125: This test exercises the fix (the `onError` runs in the `asyncErrorStub`).

guidobrei (approved):

aepfli (approved):
I can see where we are going with this, and i agree as a fix for now. But i somehow feel like we have to get to the drawing board, and maybe slowly think about this approach in general. Like all the edge cases etc. it feels like we are adding bandaid after bandaid - and i am not sure if it is time to just rethink this for once :)

Issue #1627, opened 2025-11-13
[statsig] Upgrade or provide an alternative Statsig provider based on their new javacore


lindner:
You can find details in the [Java Migration Guide](https://docs.statsig.com/server-core/migration-guides/java#java-migration-steps).

I suspect it might be useful to have two separate providers to allow for people to upgrade in a clean way.

PR #1656, opened 2025-11-23
feat(statsig)!: Migrate to Java Core

liran2000:
Migrated to Java Core according to [Migration guide](https://docs.statsig.com/server-core/migration-guides/java#java-migration-steps).

solves #1627.

@lindner and others you are welcome to review


providers/statsig/src/test/java/dev/openfeature/contrib/providers/statsig/StatsigProviderTest.java
```diff
@@ -60,12 +60,14 @@ class StatsigProviderTest {
     @BeforeAll
     static void setUp() {
         String sdkKey = "test";
-        StatsigOptions statsigOptions = new StatsigOptions();
-        statsigOptions.setLocalMode(true);
+        StatsigOptions statsigOptions = new StatsigOptions.Builder()
+                //                .setLocalMode(true)
```
lindner: Remove?   Or make it such that the test passes with/without?

providers/statsig/pom.xml
```diff
@@ -19,8 +19,9 @@
     <dependencies>
         <dependency>
             <groupId>com.statsig</groupId>
-            <artifactId>serversdk</artifactId>
-            <version>1.18.1</version>
+            <artifactId>javacore</artifactId>
+            <version>0.12.1</version>
+            <classifier>uber</classifier>
```
lindner, line 24: Why is this uber?  Does this include all of the transitive compiled versions for various platforms?

providers/statsig/README.md
```diff
@@ -74,4 +74,7 @@ As it is limited, evaluation context based tests are limited.
 See [statsigProviderTest](./src/test/java/dev/openfeature/contrib/providers/statsig/StatsigProviderTest.java)
 for more information.
 
+## Release Notes
 
+### 0.3.0
+ Migrated to Java Core according to [Migration guide](https://docs.statsig.com/server-core/migration-guides/java#java-migration-steps).
```
lindner: Definitely mention that this is a major breaking change and that there is migration needed.  Also mention that the last version of non-javacore requires a downgrade.

providers/statsig/src/main/java/dev/openfeature/contrib/providers/statsig/StatsigProvider.java
```diff
@@ -198,26 +177,19 @@ public ProviderEvaluation<Value> getObjectEvaluation(String key, Value defaultVa
...
-        return Statsig.getLayerAsync(user, featureConfig.getName()).get();
+        return statsig.getLayer(user, featureConfig.getName());
     }
 
     private Value toValue(DynamicConfig dynamicConfig) {
         MutableContext mutableContext = new MutableContext();
         mutableContext.add("name", dynamicConfig.getName());
         mutableContext.add("value", Structure.mapToStructure(dynamicConfig.getValue()));
         mutableContext.add("ruleID", dynamicConfig.getRuleID());
-        mutableContext.add("groupName", dynamicConfig.getGroupName());
```
lindner, line 214: are these provided in a different way?  If so document in the README.md

lindner (approved):
Overall LGTM, just some minor nits, since I'm starting with JavaCore.  Others who are already using this provider may have issues if the transitive deps have changed significantly.

providers/statsig/src/main/java/dev/openfeature/contrib/providers/statsig/StatsigProvider.java
```diff
@@ -262,8 +228,10 @@ private Value toValue(Layer layer) {
     @SneakyThrows
     @Override
     public void shutdown() {
-        log.info("shutdown");
-        Statsig.shutdown();
+        log.info("shutdown begin");
+        CompletableFuture<Void> shutdownFuture = statsig.shutdown();
```
lindner: check if statsig != null?

providers/statsig/pom.xml
liran2000: From [docs](https://docs.statsig.com/server-core/java-core):
> The uber JAR is recommended for most use cases as it includes native libraries for all popular platforms and simplifies deployment. Use platform-specific JARs only if you need to minimize JAR size or have specific dependency requirements.

> Recommended: Using the Uber JAR (All-in-One)
Since version 0.4.0, Statsig provides an “uber” JAR that contains both the core library and native libraries for popular supported platforms in a single package. This is the recommended approach for most users.

Since we are like a library here for multi-platform, this includes all, and consumers can control it more precisely if wanted.

providers/statsig/README.md
liran2000: marked as a breaking change at the PR title for release notes and added this to the readme:

> The provider usage is basically unchanged, the underlying implementation is changed.
   As the initialization code may change, can refer to the migration guide for details.

providers/statsig/src/main/java/dev/openfeature/contrib/providers/statsig/StatsigProvider.java
liran2000: documented

lindner:
Copilot suggests:

The failing test is FlagdProviderSyncResourcesTest.callingShutdown_wakesUpWaitingThreadWithException (line 133), which expected true but got false. This test creates a thread that waits with flagdProviderSyncResources.waitForInitialization(10000), then calls shutdown and asserts that waitingThread receives an IllegalStateException and was promptly released (< MAX_TIME_TOLERANCE * 2 ms wait).

The likely problem is in the FlagdProviderSyncResources implementation: shutdown() may not notify waiting threads or may not change the internal state in a way that causes waitForInitialization to throw the expected exception and exit quickly.

Solution:
- Ensure that flagdProviderSyncResources.shutdown() notifies all waiting threads (use notifyAll() if using synchronized blocks or the equivalent for your concurrency primitives).
- Ensure that shutdown changes a state variable so that waitForInitialization checks this state and throws IllegalStateException promptly if shutdown has been called.

Example fix (in your FlagdProviderSyncResources class):

```java
private volatile boolean isShutdown = false;

public synchronized void waitForInitialization(long deadlineMs) {
    long end = System.currentTimeMillis() + deadlineMs;
    while (!initialized && !isShutdown && System.currentTimeMillis() < end) {
        try {
            wait(end - System.currentTimeMillis());
        } catch (InterruptedException ignored) {}
    }
    if (isShutdown) {
        throw new IllegalStateException("Provider is shut down");
    }
    if (!initialized) {
        throw new GeneralError("Initialization timeout");
    }
}

    isShutdown = true;
    notifyAll();
}
```

Check that your shutdown logic matches this behavior—both in flag and notification. For reference, see the full test code in FlagdProviderSyncResourcesTest.java (ref: eef84932c5422227c31d383a0dc0b338c4a06677).

lindner:
Any updates here?

liran2000:
> Any updates here?

hi @toddbaert @chrfwow @aepfli can you please review ?

the CI failures are not from this provider.

PR #1661, opened 2025-12-02
feat: Add VMLens Tests

chrfwow:
## This PR
Adds VMLens tests, which should help with finding race conditions and other concurrency issues.

### Follow-up Tasks
Add more VMLens tests


chrfwow:
Waiting for https://github.com/jamsesso/json-logic-java/pull/65, as VMLens found a race condition in the json-logic library

chrfwow:
I disabled the test that finds the race condition in the json logic library. It does not affect us because we do not add json logic operators after application startup, and I don't think the PR to fix the data race will be merged anytime soon

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/FlagdProviderSyncResourcesCTest.java
aepfli, line 1: [Q] is CTest the official vmlens conventions for concurrency tests?

aepfli (approved):

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/FlagdProviderSyncResourcesCTest.java
chrfwow: I don't think there is such a thing as an official naming convention for VMLens tests, but we have to give them some uniquely identifying name, so that we can restrict the vmlens maven plugin to only run those tests. I am happy to use another one if you have one

leakonvalinka (approved):
lgtm!

toddbaert:
@chrfwow thanks a lot for this. There was some nasty conflicts, because of recent changes around the FATAL stuff which involved a lot of the `SyncResources` you changed in this PR. I rebased and resolved them, and moved the new, related tests from main into your new vm-lens enabled tests for the `SyncResources` (`FlagdProviderSyncResourcesCTest`).

VMlens runs as expected and I see VMLens results locally, though I also sometimes see "vmlens not configured" errors - I guess this is a log message we expect, perhaps from other modules? I can't make sense of it, because I certainly see the expected VMLens results.

toddbaert (approved):

chrfwow:
@toddbaert if I understood it correctly, vmlens runs the tests twice. The first time is just the normal test executions without the vmlens agent attached. This is where the warnings come from. And then a second time with vmlens agent, but also with the filter applied (only tests in files that end with `CT.java`). I don't know excatly why it does it like this, but running all the tests with the vmlens agent attached led to problems in our pipeline. The architecture tests sometimes never terminated

PR #1668, opened 2025-12-10
feat: add option to rebuild gRPC connection on error

toddbaert:
This PR adds an experimental option which rebuilds the entire gRPC connection if the provider goes into error state (after grace period). This can be useful for troublehsooting network issues, especially in the context of service meshes/proxies.

I added unit tests, but I also tested locally, and it works as expected. The feature can be enabled through config, but also with an env var: `FLAGD_REINITIALIZE_ON_ERROR`. It defaults to off.

Optionally depends on: https://github.com/open-feature/java-sdk-contrib/pull/1670

PR #1670, opened 2025-12-15
refactor: remove channel monitor

toddbaert:
This PR removes the channel monitor (see thread [here](https://github.com/open-feature/java-sdk-contrib/pull/1668#discussion_r2614877397)).. as Guido put it:

> The current implementation only emits errors messages on transient connection errors or shutdown. Whatever the `ChannelConnector`'s responsibility was, it seems to be handled by other components by now. 

Reasons/justification:

 - any errors seen also fire the error handler in the stream logic (this is why our e2e tests pass without this)
 - it relied on an experimental gRPC API
 - it potentially caused duplicate error events (one from the monitor, one from the stream)

I believe this was previously needed before we made the `eventStream` (RPC mode) required, and before @guidobrei optimized and commonized our streaming logic. Now, I think this code is basically dead.

If this code really is needed for some reason, I think we probably need e2e tests to exercise whatever situation that is, if those can be reasonably created.

I would like to do more refactors around this (for example we can probably delete the entire `ChannelConnector` class now, but I wanted to try to keep this PR as clear as possible.


providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/resolver/common/ChannelConnectorTest.java
```diff
@@ -1,151 +0,0 @@
-package dev.openfeature.contrib.providers.flagd.resolver.common;
```
toddbaert, line 1: Gemini suggested I remove this whole test file. I agree, and I think I should also delete the entire `ChannelConnector` class as I said in my description, but I think I want to do that separately, as that would require small refactors around shutdown which I don't want to add to this PR.

guidobrei (approved):
I like all the red removed lines in the diff ❤️ 

Maybe we should rename the class now to `ChannelDisconnector` 😆

toddbaert:
> I like all the red removed lines in the diff ❤️
> 
> Maybe we should rename the class now to `ChannelDisconnector` 😆

Haha, ya, I think it's awkward now that it just disconnects, but I want to do that separately. I wanted to make it very clear in this PR what was being removed.

chrfwow (approved):

PR #1676, opened 2026-01-08
fix(multiprovider): Properly shutdown multiprovider initialization thread pool

alxckn:
## This PR

This pull request refactors the initialization logic in the `MultiProvider` class to ensure proper resource management for the thread pool used during provider initialization. The main change is moving the creation and shutdown of the `ExecutorService` into a try-finally block, which guarantees that the thread pool is always shut down, even if an exception occurs.

We observed issues with some of our apps that fail to shutdown properly because of a hanging non-daemon thread, the changes proposed in this PR fix the issue (tested on our previously failing app).

### DIFF
--- a/providers/multiprovider/src/main/java/dev/openfeature/contrib/providers/multiprovider/MultiProvider.java
+++ b/providers/multiprovider/src/main/java/dev/openfeature/contrib/providers/multiprovider/MultiProvider.java
@@ -105,10 +105,10 @@ public class MultiProvider extends EventProvider {
         ExecutorService initPool = Executors.newFixedThreadPool(INIT_THREADS_COUNT);
         try {
             List<Future<Boolean>> results = initPool.invokeAll(tasks);
+            // Wait for all provider initializations to complete.
+            // If any provider.initialize() throws, result.get() will throw ExecutionException.
             for (Future<Boolean> result : results) {
-                if (!result.get()) {
-                    throw new GeneralError("init failed");
-                }
+                result.get();
             }
         } finally {
             initPool.shutdown();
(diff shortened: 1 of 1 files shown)


providers/multiprovider/src/main/java/dev/openfeature/contrib/providers/multiprovider/MultiProvider.java
chrfwow, line 97: Since we always return `true` and never use this value, could we get rid of this value entirely? Maybe with a `Future<Void>`?

providers/multiprovider/src/main/java/dev/openfeature/contrib/providers/multiprovider/MultiProvider.java
```diff
@@ -101,12 +100,19 @@ public void initialize(EvaluationContext evaluationContext) throws Exception {
...
-            if (!result.get()) {
-                throw new GeneralError("init failed");
+
+        ExecutorService initPool = Executors.newFixedThreadPool(INIT_THREADS_COUNT);
+        try {
+            List<Future<Boolean>> results = initPool.invokeAll(tasks);
+            // Wait for all provider initializations to complete.
+            // If any provider.initialize() throws, result.get() will throw ExecutionException.
+            for (Future<Boolean> result : results) {
+                result.get();
```
chrfwow, line 110: Maybe we should check for `result.isCancelled()`, and also add a sensible timeout `result.get(timeout, TimeUnit.MILLISECONDS);`.

liran2000:
This was moved, treated and [depecated](https://github.com/open-feature/java-sdk-contrib/tree/main/providers/multiprovider#openfeature-multi-provider-for-java), so can be closed?

[handled](https://github.com/open-feature/java-sdk/blob/main/src/main/java/dev/openfeature/sdk/multiprovider/MultiProvider.java#L126)

chrfwow:
@alxckn The multi provider was moved to the java-sdk repository: https://github.com/open-feature/java-sdk/tree/main/src/main/java/dev/openfeature/sdk/multiprovider
Please migrate to this version. As pointed out by @liran2000, the issues you worked on were fixed there

alxckn:
@chrfwow @liran2000 thanks for reviewing sorry I hadn't noticed the migration, and happy to see it's been internalized and already fixed :) We'll migrate to fix our issue

toddbaert:
@alxckn it's migrated but not released! We have a few new things / fixes in Java though, so I'll try to get a release out soon (if not today): https://github.com/open-feature/java-sdk/pull/1770

PR #1678, opened 2026-01-09
fix: thread memory leak in InProcessResolver

### LINKED ISSUE #1677: Thread memory leak in flagd InProcessResolver
The stateWatcher thread in `InProcessResolver` is not properly closed when the `shutdown()` method is invoked, leading to a memory leak.

Consider implementing a proper shutdown mechanism (similar to that in `FlagStore.shutdown()`).

paul-kraftlauget:
## This PR

- fixes a memory leak in InProcessResolver

### Related Issues

Fixes #1677

### How to test
Please see the unit test in this PR

### DIFF
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolver.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolver.java
@@ -40,6 +40,8 @@ import org.apache.commons.lang3.StringUtils;
  */
 @Slf4j
 public class InProcessResolver implements Resolver {
+
+    static final String STATE_WATCHER_THREAD_NAME = "InProcessResolver.stateWatcher";
     private final Storage flagStore;
     private final TriConsumer<ProviderEvent, ProviderEventDetails, Structure> onConnectionEvent;
     private final Operator operator;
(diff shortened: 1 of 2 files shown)


providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolver.java
```diff
@@ -69,7 +71,7 @@ public void init() throws Exception {
         flagStore.init();
         final Thread stateWatcher = new Thread(() -> {
             try {
-                while (true) {
+                while (!shutdown.get()) {
```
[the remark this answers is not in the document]
toddbaert: While this is true, what's been added here is an objective improvement.

@paul-kraftlauget ... up to you if you'd like to go further. I do think this comment is correct

toddbaert:
@paul-kraftlauget Thanks a lot! can you fix the DCO failure? You can do a `git commit --amend --signoff` which just gives a legal signoff to the commit.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolver.java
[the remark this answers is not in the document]
paul-kraftlauget: Yes, it was correct. And looking at the file sync implementation, there wouldn't be another event sent when shutdown, so the leak still would occur. I have pushed an update now. I hope it is ok.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolver.java
```diff
@@ -105,12 +109,12 @@ public void init() throws Exception {
...
             } catch (InterruptedException e) {
-                log.warn("Storage state watcher interrupted", e);
-                Thread.currentThread().interrupt();
+                log.debug("Storage state watcher interrupted, most likely shutdown was invoked", e);
             }
-        });
+        }, "InProcessResolver.stateWatcher");
         stateWatcher.setDaemon(true);
         stateWatcher.start();
+        this.stateWatcher.set(stateWatcher);
```
chrfwow, line 78: I think we should set the reference before starting the thread

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolverTest.java
```diff
@@ -543,6 +544,35 @@ void flagSetMetadataIsOverwrittenByFlagMetadataToEvaluation() throws Exception {
...
+
+        // when
+        inProcessResolver.init();
+        var threadCountAfterInit = currentDaemonThreadCount();
+        inProcessResolver.shutdown();
+
+        // then
+        assertThat(threadCountAfterInit).isGreaterThan(initialThreadCount);
+        Awaitility.await()
+                .untilAsserted(() -> assertThat(currentDaemonThreadCount()).isEqualTo(initialThreadCount));
```
chrfwow: Could we instead assert that the thread set does not contain a thread named `InProcessResolver.stateWatcher`?

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolver.java
paul-kraftlauget: fixed

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolverTest.java
paul-kraftlauget: Added to the test

paul-kraftlauget: I added it instead of replacing it. I used test-driven development for this, first reproduced the error with a test. It was before I added the thread name. I added the thread name for visibility, not for behavior, although it does make it possiblity for outside code to start interrupting the thread themselves.

I think the daemon thread count check is still valid. In fact, I was debating to start a new thread to first try to let it stop gracefully (and let any final events be sent), then after a given deadline, to forcefully stop it. In any case, the behavior to test is that any daemon threads created, regardless of number, are all stopped after the InProcessResolver is complete.

chrfwow (approved):

aepfli (approved):

toddbaert (approved):

PR #1681, opened 2026-01-15
feat: Propagate fatal error on wait for init

### LINKED ISSUE #1675: [flagd] Improve wait in init
The flagd init code is not ideal.

- the `FlagdProvider`'s `initialize` function is called, which calls `flagResolver.init()`... the method then blocks waiting for `syncResources.waitForInitialization(...)`
- `syncResources.waitForInitialization(...)` always runs until the deadline timeout; it would be ideal if it simply immediately returned if an error occurs, and only timed-out if nothing happens
- the error returned by `initialize` does not return the exception thrown by the stream

Fixing the above should improve the DX of the provider, and ease troubleshooting.

More background from a previous PR is below:
---

@leakonvalinka ... I kept seeing intermittent test failures with this PR ([this](https://github.com/open-feature/java-sdk-contrib/pull/1624#discussion_r2641110137) was the biggest example. At first, I thought it was a race-condition in the test itself, but it was not). I have removed this now, because I found the root cause:

The main issue was [here](https://github.com/open-feature/java-sdk-contrib/pull/1624/changes#diff-5651aa7cc591c468780331739eeb2a6316c48cfd11b16479c3608fea1a5d6f34R95-R99). It was not any logical error with the code you added, but your changes surfaced an existing issue. The race condition works like this:

- the `FlagdProvider`'s `initialize` function is called, which calls `flagResolver.init()`... the method then blocks waiting for `syncResources.waitForInitialization(...)`
- normally, in error situations, this method then times out (another thread runs error handlers)
- however, in our situation, this method doesn't just time out; due to our changes to shut down when FATAL errors are seen, the method starting throwing `IllegalStateException` (see the condition at the end of the method)
- since this is not a FATAL error, this was thrown by init, and depending on timing, would "override" the FATAL with a non-fatal error (the `IllegalStateException`)

My fix to throw the Fatal error depending on the last event is a bandaid fix, because this PR has already been open for a while.

To really fix the issue, we should improve the way the waiting in the init works, and actually capture the error during init, and not wait for the full timeout every time. I will open a new issue for this.

cc @guidobrei @aepfli

_Originally posted by @toddbaert in https://github.com/open-feature/java-sdk-contrib/issues/1624#issuecomment-3715523407_

chrfwow:
## This PR
Propagates fatal stream error to thread waiting for initialization 

### Related Issues

Fixes #1675

### DIFF
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProviderSyncResources.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProviderSyncResources.java
@@ -3,6 +3,7 @@ package dev.openfeature.contrib.providers.flagd;
 import dev.openfeature.sdk.EvaluationContext;
 import dev.openfeature.sdk.ImmutableContext;
 import dev.openfeature.sdk.ProviderEvent;
+import dev.openfeature.sdk.ProviderEventDetails;
 import dev.openfeature.sdk.exceptions.FatalError;
 import dev.openfeature.sdk.exceptions.GeneralError;
 import lombok.Getter;
@@ -15,10 +16,10 @@ import lombok.Setter;
 @Getter
 class FlagdProviderSyncResources {
     @Setter
-    private volatile ProviderEvent previousEvent = null;
+    private volatile ProviderEvent previousEvent;
 
-    @Setter
     private volatile boolean isFatal;
+    private volatile ProviderEventDetails fatalProviderEventDetails;
 
     private volatile EvaluationContext enrichedContext = new ImmutableContext();
     private volatile boolean isInitialized;
@@ -39,7 +40,6 @@ class FlagdProviderSyncResources {
             return false;
         }
         this.isInitialized = true;
-        this.isFatal = false;
         this.notifyAll();
         return true;
     }
@@ -68,7 +68,7 @@ class FlagdProviderSyncResources {
     public void waitForInitialization(long deadline) {
         long start = System.currentTimeMillis();
         long end = start + deadline;
-        while (!isInitialized && !isShutDown) {
+        while (!isInitialized && !isShutDown && !isFatal) {
             long now = System.currentTimeMillis();
             // if wait(0) is called, the thread would wait forever, so we abort when this would happen
             if (now >= end) {
@@ -77,7 +77,7 @@ class FlagdProviderSyncResources {
             }
             long remaining = end - now;
             synchronized (this) {
-                if (isShutDown) {
+                if (isShutDown || isFatal) {
                     break;
                 }
                 if (isInitialized) { // might have changed in the meantime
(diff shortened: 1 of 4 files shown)


providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProviderSyncResources.java
```diff
@@ -91,11 +91,16 @@ public void waitForInitialization(long deadline) {
...
             }
         }
+        if (isFatal) {
+            var fatalEvent = fatalProviderEventDetails;
+            if (fatalEvent != null) {
+                throw new FatalError("Initialization failed due to a fatal error: " + fatalEvent.getMessage());
+            } else {
+                throw new FatalError("Initialization failed due to a fatal error.");
+            }
+        }
```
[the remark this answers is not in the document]
chrfwow: I assume that the user only invokes `fatalError(...)` only with a fatal error. I do not plan to add any checks to that matter, because there is nothing to do in this case.
Therefore I will ignore this comment

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProviderSyncResources.java
```diff
@@ -91,11 +91,16 @@ public void waitForInitialization(long deadline) {
                 }
             }
         }
+        if (isFatal) {
+            var fatalEvent = fatalProviderEventDetails;
+            if (fatalEvent != null) {
```
leakonvalinka, line 96: Question: Do we need this null check here? As far as I can see this is always set to an actual even object except in the tests.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProviderSyncResources.java
```diff
@@ -91,11 +91,16 @@ public void waitForInitialization(long deadline) {
...
+                throw new FatalError("Initialization failed due to a fatal error: " + fatalEvent.getMessage());
+            } else {
+                throw new FatalError("Initialization failed due to a fatal error.");
+            }
+        }
         if (isShutDown) {
             String msg = "Already shut down due to previous error.";
```
leakonvalinka: Nit pick: You could remove the variable and move the string directly into the `GeneralError`

leakonvalinka (approved):
I left 2 comments with a question and a nit, but otherwise it lgtm!

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProviderSyncResources.java
chrfwow: This is true, but I don't want an NPE because of _that_

aepfli (approved):

toddbaert (approved):

PR #1696, opened 2026-02-10
feat: extract core flagd evaluation and API modules

### LINKED ISSUE #1695: [flagd] Extract common Evaluator interface into shared modules
### Background

The flagd in-process resolver contains evaluation logic that could be reused by other implementations. Currently, the evaluation is tightly coupled to the provider, making it difficult to use alternative implementations or share the evaluation logic.

We already have something similar in Go and TS/JS:

See a Go equivalent interface [here](https://github.com/open-feature/flagd/blob/47f904b3a5a5a6e4389de6201bef3c15158dbc97/core/pkg/evaluator/ievaluator.go#L37]).
And a JS equivalent module [here](https://github.com/open-feature/js-sdk-contrib/tree/main/libs/shared/flagd-core)

### Requirements

- [ ] Create `flagd-api` module containing the `Evaluator` interface
- [ ] Create `flagd-core` module with the `FlagdCore` implementation of `Evaluator`
- [ ] Move flag parsing, targeting operators (fractional, semver, string comparison), and model classes to `flagd-core`
- [ ] Update provider to depend on `flagd-core` (which transitively provides `flagd-api`)
- [ ] Ensure provider references only the `Evaluator` interface, not the `FlagdCore` implementation directly
- [ ] Move unit tests for operators and parsing to `flagd-core`

While doing this, all other changes should be avoided if possible.

toddbaert:
Extraction of the core evaluator of flagd into 2 modules:

- flagd-api (interfaces only, so that others can implement alternatives)
- flagd-core (our current implementation)
  - the API is similar to the flagdCore object in JS, but has a smaller surface area for now (only what was currently needed in Java)
  
I've also included some basic automation to make sure the local version of flagd-core is always resolved in development, and at provider release the dep is pinned to exactly what is present locally. 

I've intentionally made as few changes as possible, and made no functional changes; note that a large volume of the changes are `git mv` operations.

Fixes: https://github.com/open-feature/java-sdk-contrib/issues/1695


tools/flagd-core/src/main/java/dev/openfeature/contrib/tools/flagd/core/FlagdCore.java
```diff
@@ -0,0 +1,320 @@
...
+
+    private List<String> getChangedFlagsKeys(Map<String, FeatureFlag> newFlags) {
+        return newFlags.entrySet().stream()
+                .filter(entry -> {
+                    FeatureFlag oldFlag = flags.get(entry.getKey());
+                    return oldFlag == null || !oldFlag.equals(entry.getValue());
+                })
+                .map(Map.Entry::getKey)
+                .collect(Collectors.toList());
+    }
```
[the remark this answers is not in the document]
toddbaert: Refactor gone bad - reverted, and test added.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolver.java
```diff
@@ -206,134 +185,4 @@ static QueueSource getQueueSource(final FlagdOptions options) {
                 ? new FileQueueSource(options.getOfflineFlagSourcePath(), options.getOfflinePollIntervalMs())
                 : new SyncStreamQueueSource(options);
     }
-
-    private <T> ProviderEvaluation<T> resolve(Class<T> type, String key, EvaluationContext ctx) {
```
toddbaert, line 210: This is one of the few things that weren't moved 1:1

This class was eliminated because FlagdCore now handles flag storage internally and includes the metadata in evaluation results directly. The `StorageQueryResult` intermediary is no longer needed.

tools/flagd-api/src/main/java/dev/openfeature/contrib/tools/flagd/api/Evaluator.java
```diff
@@ -0,0 +1,87 @@
...
+import dev.openfeature.sdk.ProviderEvaluation;
+import dev.openfeature.sdk.Value;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Interface for in-process flag evaluation in flagd.
+ * Combines flag storage and evaluation into a single abstraction.
+ */
+public interface Evaluator {
```
toddbaert, line 13: This new interface is equivalent to that in JS' flagd core, and the Go equivalent - though only the minimal features have been implemented (we don't need bulk evaluation here).

.github/workflows/release-please.yml
```diff
@@ -65,11 +65,12 @@ jobs:
         # This means there's no way to skip publishing of a particular module in a multi-module build, so we iterate over each module and publish them individually,
         # letting exists-maven-plugin skip the nexus-staging-maven-plugin's entire deploy goal if the artifact exists.
         run: |
+          FLAGD_CORE_VERSION=$(cat tools/flagd-core/version.txt)
           mvn --non-recursive --batch-mode --settings release/m2-settings.xml -DskipTests -Dcheckstyle.skip clean deploy
           modules=($(cat pom.xml | grep "<module>" | sed 's/\s*<.*>\(.*\)<.*>/\1/'))
           for module in "${modules[@]}"
           do
-              mvn --batch-mode --projects $module --settings release/m2-settings.xml -DskipTests -Dcheckstyle.skip clean deploy
+              mvn --batch-mode --projects $module --settings release/m2-settings.xml -DskipTests -Dcheckstyle.skip -Dflagd-core.version=$FLAGD_CORE_VERSION clean deploy
```
toddbaert, line 73: This pins the flagd-core dep in the flagd-provider to the current version in source, ensuring we never have incompatibility.

chrfwow (approved):
looks good

aepfli (approved):

PR #1709, opened 2026-03-06
fix: fixing null pointer exception error

### LINKED ISSUE #1708: [flagd] [BUG] Null Pointer Dereference in getFlagMetadata()
In https://github.com/open-feature/java-sdk-contrib/blob/7760d098b6c965c5a299e274ddf790e10b967738/tools/flagd-core/src/main/java/dev/openfeature/contrib/tools/flagd/core/FlagdCore.java#L270 `currentFlagSetMetadata` and/or `flag.getMetadata()` can be `null`, causing `entrySet()` to throw NPE.

A defensive check for null should be added to fix the problem for servers that don't suport metadata.

Ayushman-Gaur:
## fix: null pointer exceptionn error
- Adds null check for `flag.getMetadata()` in `getFlagMetadata()` to prevent 
  NullPointerException when a flagd server doesn't support metadata.

### Related Issues
Fixes #1708

### Notes
`currentFlagSetMetadata` is always initialized as a new HashMap so it won't 
be null, but `flag.getMetadata()` can return null on servers that don't support 
metadata. Added a null guard before iterating over it.

### How to test
A new test `resolveBooleanValue_flagWithNullMetadata_doesNotThrowNPE` has been 
added to `FlagdCoreTest` that verifies no NPE is thrown when evaluating a flag 
with no metadata defined.

### DIFF
--- a/tools/flagd-core/src/main/java/dev/openfeature/contrib/tools/flagd/core/FlagdCore.java
+++ b/tools/flagd-core/src/main/java/dev/openfeature/contrib/tools/flagd/core/FlagdCore.java
@@ -274,7 +274,7 @@ public class FlagdCore implements Evaluator {
             addEntryToMetadataBuilder(metadataBuilder, entry.getKey(), entry.getValue());
         }
 
-        if (flag != null && flag.getTargeting() != null) {
+        if (flag != null && flag.getMetadata() != null) {
             for (Map.Entry<String, Object> entry : flag.getMetadata().entrySet()) {
                 addEntryToMetadataBuilder(metadataBuilder, entry.getKey(), entry.getValue());
             }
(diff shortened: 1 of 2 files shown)


chrfwow (approved):

Ayushman-Gaur:
The CI failure is in the go-feature-flag provider which is unrelated to my changes in flagd-core. This appears to be a pre-existing issue.

tools/flagd-core/src/main/java/dev/openfeature/contrib/tools/flagd/core/FlagdCore.java
cupofcat: You should also guard the for loop against `currentFlagSetMetadata` being null.

cupofcat (approved):
Please solve the CI error with formatting too. Thanks for the contribution!

aepfli (approved):

toddbaert (approved):
Thanks @Ayushman-Gaur

PR #1752, opened 2026-03-30
perf(flagd): speed up e2e test execution via container pool and parallel scenarios

aepfli:
## Summary

Reduces flagd provider e2e test wall-clock time by **~75%** through a pre-warmed container pool and Cucumber-level parallel scenario execution — no Surefire fork changes needed.

### 1. Container pool (`ContainerPool` + `ContainerEntry`)

The previous setup used a **single shared Docker Compose stack** for all scenarios within a runner. Since the flagd-testbed launchpad controls a single flagd process via `/start`, `/stop`, `/restart`, and `/change` HTTP endpoints, scenarios sharing one container would race on these operations and could not run concurrently.

The fix is a **pre-warmed container pool**: `@BeforeAll` starts N containers in parallel (~45s, once per JVM), and each Cucumber scenario borrows one `ContainerEntry` for its duration, giving it a fully isolated flagd process. After teardown the entry is returned to the pool.

Pool size is tunable via `-Dflagd.e2e.pool.size=N` (default: 2).

A reference counter ensures that when multiple suite runners share the same JVM (`reuseForks=true`), containers are only started on the first `initialize()` call and stopped on the last `shutdown()` call.

### 2. Parallel Cucumber scenarios

With `cucumber.execution.parallel.enabled=true` and `fixed.parallelism=2` (matching the default pool size), scenarios within each runner execute concurrently.

Correctness safeguards via exclusive resource locks:
- **`@env-var`** scenarios serialised behind `ENV_VARS` lock (requires companion PR [flagd-testbed#359](https://github.com/open-feature/flagd-testbed/pull/359))
- **`@grace`** scenarios (container restart + reconnection timing) serialised behind `CONTAINER_RESTART` lock
- **`ConfigCucumberTest`** disables parallelism entirely (env-var mutations in <0.4s suite — no benefit)

### 3. Per-provider teardown

Replaced `OpenFeatureAPI.getInstance().shutdown()` (global — tears down all providers) with a per-provider `NoOpProvider` swap through the SDK lifecycle. This properly detaches event emitters and is safe for parallel execution.

### 4. Event drain fix

`EventSteps` now drains events up to and including the first match instead of `clear()`-ing the entire queue. This prevents stale events (e.g. a READY from before a disconnect) from satisfying later assertions.

## Architecture

```
Before:
  Runner 1 (sequential) → start 1 container → scenario 1,2,...N (sequential) → stop
  Runner 2 (sequential) → start 1 container → scenario 1,2,...N (sequential) → stop
  Runner 3 (sequential) → start 1 container → scenario 1,2,...N (sequential) → stop

After (pool + parallel):
  Runner 1: start pool(2) → 2 parallel scenarios → ... → defer shutdown
  Runner 2: reuse pool    → 2 parallel scenarios → ... → defer shutdown
  Runner 3: reuse pool    → 2 parallel scenarios → ... → stop pool
```

## Dependencies

- [flagd-testbed#359](https://github.com/open-feature/flagd-testbed/pull/359) — adds `@env-var` tag to config scenarios (submodule temporarily pointed at PR branch)

## Draft — watching CI
Opening as draft to observe CI behaviour before merging.


providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/steps/EventSteps.java
```diff
@@ -60,9 +60,18 @@ public void eventHandlerShouldBeExecutedWithin(String eventType, int ms) {
...
-        state.lastEvent = state.events.stream()
-                .filter(event -> event.type.equals(eventType))
-                .findFirst();
-        state.events.clear();
+        // Drain all events up to and including the first match. This ensures that
+        // older events (e.g. a READY from before a disconnect) cannot satisfy a
+        // later assertion that expects a *new* event of the same type, while still
+        // preserving events that arrived *after* the match for subsequent steps.
+        Event matched = null;
+        while (!state.events.isEmpty()) {
```
chrfwow, line 68: Are w esure that no new events can be emitted while or immediately after this runs? Otherwise this loop might not be sufficient

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/steps/EventSteps.java
```diff
@@ -60,9 +60,18 @@ public void eventHandlerShouldBeExecutedWithin(String eventType, int ms) {
...
+            Event head = state.events.poll();
+            if (head != null && head.type.equals(eventType)) {
+                matched = head;
+                break;
+            }
+        }
+        state.lastEvent = java.util.Optional.ofNullable(matched);
```
chrfwow, line 75: The naming is not ideal, this is not the last event, it's the last event that matches the `eventType`

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/steps/ProviderSteps.java
```diff
@@ -18,92 +21,87 @@
...
+                when().post("http://" + ContainerUtil.getLaunchpadUrl(state.containerEntry.container) + "/stop")
+                        .then()
+                        .statusCode(200);
+            }
+            ContainerPool.release(state.containerEntry);
+            state.containerEntry = null;
+        }
+        // Replace the domain provider with a NoOp through the SDK lifecycle so the SDK
+        // properly calls detachEventProvider (nulls onEmit) and shuts down the emitter
+        // executor — neither of which happens when calling provider.shutdown() directly.
```
chrfwow, line 60: Is this something that should happen when we call `shutdown`?

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/ContainerPool.java
```diff
@@ -0,0 +1,106 @@
...
+            refCount.decrementAndGet();
+            throw e;
+        } finally {
+            executor.shutdown();
+        }
+        log.info("Container pool ready ({} containers).", POOL_SIZE);
+    }
+
+    public static void shutdown() {
+        int remaining = refCount.decrementAndGet();
```
chrfwow, line 87: Could it be possible that all current users call `shutdown`, even though there are still outstanding users, who have not called `initialize` yet? Then we would shutdown the pool, even though there are still tests lined up

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/ContainerPool.java
```diff
@@ -0,0 +1,106 @@
...
+ * are still in use by another class running concurrently in the same JVM.
+ */
+@Slf4j
+public class ContainerPool {
+
+    private static final int POOL_SIZE = Integer.getInteger(
+            "flagd.e2e.pool.size", Math.min(Runtime.getRuntime().availableProcessors(), 4));
+
+    private static final BlockingQueue<ContainerEntry> pool = new LinkedBlockingQueue<>();
+    private static final List<ContainerEntry> all = new ArrayList<>();
```
chrfwow, line 36: I think this needs to be a concurrent data structure too, so that we guarantee that all changes to the list are also visible to another thread calling shutdown

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/ContainerPool.java
aepfli: this is happening on the beforeAll - per test suites, tests suites are not parallel anyways, there is another improvement for this.

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/steps/EventSteps.java
aepfli: what do you mean, we want to specifically keep events which are generated while this loop runs. because an ready event can happen shortly after a disconnect, and while we wait for the disconnect including cleanup, we might even remove the new ready

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/steps/EventSteps.java
aepfli: the last event is for the current test state, the last tracked event. we do not care about the type. this is only used to verify some event information.

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/steps/ProviderSteps.java
aepfli: the tests are now running in parallel, so shutting down the api is a no go, as it is also messing with other tests

chrfwow: No, but I mean in general, not specifically this test case

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/ContainerPool.java
chrfwow: If the test suites do not run in parallel, then we don't need this sync mechanism. If they do run in parallel, the scenrio in my comment could (even though it is unlikely) occur

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/steps/EventSteps.java
chrfwow: If we are worried about events shortly after a disconnect, we should wait for some time and check events afterwards or in the meantime. This loop might be done after 0 or 1 iterations, and might be done before we receive such an event that we would want to wait for

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/steps/ProviderSteps.java
aepfli: maybe, but currently the goal is speeding up tests ;) - when we reset a new provider, we are actually cleaning up

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/ContainerPool.java
aepfli: i ran it shortly in parallel, and the next iteration will add more flexibility to it, where all the tests are actually parallel. but this is in the next follow up pr. i shortly ran all 3 tests in parallel with some hacks, but it was not worth the effort.

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/steps/EventSteps.java
aepfli: this cleans all the events which has happened till our matched event. so cleaning the list till out event. so that if there are events in the meantime, they stay in the list, and we can match in the next check against all of them.

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/ContainerPool.java
```diff
@@ -0,0 +1,106 @@
...
+                try {
+                    entry.stop();
+                } catch (IOException suppressed) {
+                    e.addSuppressed(suppressed);
+                }
+            });
+            pool.clear();
+            all.clear();
+            refCount.decrementAndGet();
+            throw e;
```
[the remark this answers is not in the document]
toddbaert: Applied!

toddbaert (approved):
@aepfli This is amazing. It is much faster on my machine, nearly 25% of the previous time. I also simulated some failures and got all the expected results. Code looks good. I made a few very small tweaks and fixed some conflicts. @chrfwow 's issues seem addressed. Will merge.

aepfli:
nice this means, i can work on the second one, which should make it even faster :D

PR #1753, opened 2026-04-01
perf(flagd): run all 3 e2e resolver modes concurrently via @TestFactory

aepfli:
## Summary

Stacked on top of #1752. Replaces three sequential `@Suite` runner classes with a single `RunE2ETests` class using Jupiter `@TestFactory` methods, cutting wall-clock time from ~3:20 to ~2:00.

### 1. `RunE2ETests` — three concurrent `@TestFactory` methods

Each factory (`rpc()`, `inProcess()`, `file()`) launches a full Cucumber engine for its resolver mode. With `@Execution(CONCURRENT)` and `junit.jupiter.execution.parallel.enabled=true`, all three engines run simultaneously:

```
Before: RPC (sequential) → InProcess → File  ≈ 3:20
After:  RPC ──┐
              ├── all run in parallel ──→ ≈ 2:00
         File ┘
```

Each factory returns a `Stream<DynamicNode>` mirroring the Cucumber TestPlan (engine → feature → scenario), so IDEs show the full expandable tree with accurate pass/fail/skip per scenario — individual scenarios can be re-run from IntelliJ.

### 2. `CucumberResultListener` — full `TestExecutionListener`

Captures every lifecycle event (started, finished, skipped, dynamic) from the Cucumber engine and maps them to `DynamicTest` results. The listener tracks both started and finished state to correctly report scenarios that started but did not complete.

### 3. `ContainerPool` — JVM shutdown hook + restart semaphore

- Lazy `ensureInitialized()` via `AtomicBoolean` — pool starts once on first `acquire()`, shared across all three concurrent engines
- JVM shutdown hook replaces the previous ref-counted `initialize()`/`shutdown()` — no lifecycle calls needed from test classes
- `Semaphore(1)` serialises disruptive container operations (stop/restart) across parallel engines to prevent cascading init timeouts

### 4. Per-resolver glue packages

`RpcSetup`, `InProcessSetup`, `FileSetup` — each a simple `@Before` hook that sets `state.resolverType`, allowing all three engines to share the same step definitions with isolated per-scenario state.

### 5. Deleted `RunRpcTest`, `RunInProcessTest`, `RunFileTest`

No longer needed — `RunE2ETests` replaces all three.

### 6. Envoy cluster improvements (test-harness submodule)

Added `connect_timeout: 0.25s` and active TCP health checks (`interval: 1s`) to both flagd clusters in envoy config. Envoy now detects and recovers from upstream flagd restarts within one health-check cycle.

## Known limitations

- **`@targetURI @in-process` scenarios (7) are excluded** from the parallel run. Root cause: `retryBackoffMaxMs` controls both the initial-connection throttle in `SyncStreamQueueSource` (when `getMetadata()` fails because envoy's upstream isn't ready yet) and the post-disconnect reconnect backoff. These cannot be tuned independently — reducing the backoff for fast initial connection breaks the reconnect event timing tests. Tracked in [flagd#1584](https://github.com/open-feature/flagd/issues/1584) — once `getMetadata()` is removed, these can be re-enabled by removing `"targetURI"` from the `inProcess()` excludeTags.
- **`file()[4][1-3]`** — FILE resolver lacks flag-set metadata support (SDK limitation, pre-existing)
- **`inProcess()[3][2]`, `[6][1-3]`, `[8][2]`** — contextEnrichment failures (pre-existing, also fail on base branch)

## Run from repo root

```bash
./mvnw -pl providers/flagd -Pe2e test
```


aepfli:
## Notes on the concurrent `@TestFactory` approach (tradeoffs for reviewers)

Now that the suite is green against test-harness v3.8.0, a few notes on what this design buys us and what it costs, so it's a conscious choice rather than a hidden one.

### Performance (measured, same machine, test-harness v3.8.0)

| | Old (3 sequential `@Suite` runners) | New (`RunE2ETests`, 3 modes concurrent) |
|---|---|---|
| in-process | 90.1s | — |
| rpc | 69.6s | — |
| file | 47.5s | — |
| **e2e test phase** | **~207s** (sum) | **~130s** (CI: 134s) |

≈ **35–40% faster (~1.6×), ~75s saved per run.**

Worth knowing: the concurrent run (~130s) is actually slower than the slowest single mode (in-process, 90s) — the three engines contend for the shared container pool (`POOL_SIZE = min(cores, 4)`) and CPU, so the gain is ~1.6×, not ~3×. The bottleneck moves from "sum of modes" to "container pool + CPU saturation."

### Developer experience (the real upside)

- Single entry point instead of three runner classes.
- `@TestFactory` produces an expandable per-scenario tree in IDEs — each engine → feature → scenario shows accurate pass/fail/skip, and individual scenarios are re-runnable straight from the IDE.
- Opens the door to richer reporting and per-scenario re-runs that the `@Suite` runners couldn't offer.

### Fragility (the thing to be aware of going forward)

The speedup comes from running **three concurrent Cucumber engines in one JVM**, which is unconventional and has sharp edges. This branch had to fix three contamination bugs that all stem from it:

1. **Resolver-glue recursion** — the per-resolver `@Before` hooks lived under the shared `steps` glue package; Cucumber scans glue recursively, so every engine loaded all three and `resolverType` was set non-deterministically (an `@in-process` scenario could build with the FILE resolver). Fixed by moving them to a sibling package.
2. **Process-global env vars** — `@env-var` scenarios mutate real environment variables that concurrent config-default scenarios read at `build()` time. Cucumber's `exclusive-resources` locks only coordinate **within** a single engine, not across the three, and env vars are JVM-global. Fixed with per-engine read locks on `ENV_VARS`.
3. **Testcontainers init race** — `DockerClientProviderStrategy` resolves via a non-thread-safe `ServiceLoader`; parallel first-time pool startup threw `ConcurrentModificationException`. Fixed by warming the Docker client once before the parallel starts.

Net: the isolation model works, but it's load-bearing and non-obvious. Anyone adding scenarios or tags later needs to keep it in mind, or contamination can quietly return.

### Possibility for later (not for this PR)

Most of the gain is just from parallelizing the three modes. The same win is achievable by running them as **three separate Surefire forks** (JVM-level isolation) instead of three engines in one JVM — that removes the entire class of cross-talk bugs above, at the cost of ~3× container-pool memory. Worth considering if the in-JVM model ever becomes a maintenance burden.

PR #1772, opened 2026-04-09
feat: add GCP Secret Manager OpenFeature provider

mahpatil:
## Summary

- Adds a new OpenFeature provider for **GCP Secret Manager** that reads feature flags from GCP Secret Manager secrets
- Includes flag caching, value conversion for all OpenFeature types (bool, string, int, double, object), and background polling for updates
- Adds a sample application under `samples/gcp-secret-manager-sample/` with setup/teardown scripts

## Provider Details

- **Package**: `providers/gcp-secret-manager`
- **Class**: `GcpSecretManagerProvider`
- Supports all OpenFeature flag types via structured or plain-text secret values
- Configurable poll interval and GCP project settings

## Test plan

- [ ] Unit tests for `FlagCache`, `FlagValueConverter`, and `GcpSecretManagerProvider`
- [ ] Integration test (`GcpSecretManagerProviderIntegrationTest`) requires a real GCP project — set `GCP_PROJECT_ID` env var to run
- [ ] Sample app: follow `samples/gcp-secret-manager-sample/README.md` to run end-to-end


providers/gcp-secret-manager/src/main/java/dev/openfeature/contrib/providers/gcpsecretmanager/FlagValueConverter.java
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    private static Boolean convertBoolean(String raw) {
+        String trimmed = raw.trim();
+        if ("true".equalsIgnoreCase(trimmed)) {
+            return Boolean.TRUE;
+        }
+        if ("false".equalsIgnoreCase(trimmed)) {
+            return Boolean.FALSE;
+        }
```
chrfwow, line 81: Why not `Boolean.parseBoolean()`?

providers/gcp-secret-manager/src/test/java/dev/openfeature/contrib/providers/gcpsecretmanager/FlagCacheTest.java
```diff
@@ -0,0 +1,91 @@
...
+import java.time.Instant;
+import java.time.ZoneOffset;
+import java.util.Optional;
+import java.util.concurrent.atomic.AtomicReference;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+
+@DisplayName("FlagCache")
+class FlagCacheTest {
```
chrfwow, line 16: I'd suggest adding a test that lets a cache entry time out while a get operation and an insert operation are concurrently ongoing. This way we can make sure that a removal of a timed out value does not remove the newly added value by accident. Maybe [VmLens](https://github.com/vmlens/vmlens) can help with that (see our [existing VmLens tests](https://github.com/open-feature/java-sdk-contrib/blob/main/providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/FlagdProviderCTest.java)).

providers/gcp-secret-manager/src/test/java/dev/openfeature/contrib/providers/gcpsecretmanager/FlagValueConverterTest.java
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    @Nested
+    @DisplayName("Integer conversion")
+    class IntegerConversion {
+
+        @Test
+        @DisplayName("converts numeric string to Integer")
+        void numericString() {
+            assertThat(FlagValueConverter.convert("42", Integer.class)).isEqualTo(42);
```
chrfwow, line 51: Maybe we should also add a test for negative numbers

providers/gcp-secret-manager/src/test/java/dev/openfeature/contrib/providers/gcpsecretmanager/FlagValueConverterTest.java
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    @Nested
+    @DisplayName("Double conversion")
+    class DoubleConversion {
+
+        @Test
+        @DisplayName("converts numeric string to Double")
+        void numericString() {
+            assertThat(FlagValueConverter.convert("3.14", Double.class)).isEqualTo(3.14);
```
chrfwow, line 69: Do we also want to support the exponential format, e.g. `3141.5e-3`? We should add tests accordingly

providers/gcp-secret-manager/src/test/java/dev/openfeature/contrib/providers/gcpsecretmanager/GcpSecretManagerProviderIntegrationTest.java
```diff
@@ -0,0 +1,106 @@
...
+        String projectId = System.getenv("GCP_PROJECT_ID");
+        GcpSecretManagerProviderOptions opts =
+                GcpSecretManagerProviderOptions.builder().projectId(projectId).build();
+        provider = new GcpSecretManagerProvider(opts);
+        provider.initialize(new ImmutableContext());
+    }
+
+    @AfterEach
+    void tearDown() {
+        if (provider != null) {
```
chrfwow, line 59: When is the provider `null`?

providers/gcp-secret-manager/src/test/java/dev/openfeature/contrib/providers/gcpsecretmanager/FlagCacheCTest.java
```diff
@@ -0,0 +1,80 @@
...
+    @Test
+    void concurrentExpiryAndInsertDoNotLoseNewEntry() throws Exception {
+        AtomicReference<Instant> now = new AtomicReference<>(Instant.parse("2024-01-01T00:00:00Z"));
+        Clock controllableClock = new Clock() {
+            @Override public ZoneOffset getZone() { return ZoneOffset.UTC; }
+            @Override public Clock withZone(java.time.ZoneId zone) { return this; }
+            @Override public Instant instant() { return now.get(); }
+        };
+
+        FlagCache cache = new FlagCache(Duration.ofSeconds(30), 100, controllableClock);
```
chrfwow, line 41: What do we need this `cache` for?

providers/gcp-secret-manager/src/test/java/dev/openfeature/contrib/providers/gcpsecretmanager/FlagCacheCTest.java
```diff
@@ -0,0 +1,80 @@
...
+                // Make entry expire
+                fresh.set(fresh.get().plus(Duration.ofMinutes(6)));
+
+                Runner.runParallel(
+                        // Thread A: re-insert fresh value for the same key
+                        () -> sharedCache.put("key", "new-value"),
+                        // Thread B: get() triggers expiry-removal of the old entry
+                        () -> sharedCache.get("key"));
+
+                // After both threads complete, either the new value is present or the cache is
```
chrfwow, line 71: After a thread completes that puts a new value to the cache, I would not expect the cache to be empty. Am I missing something here?

providers/gcp-secret-manager/src/test/java/dev/openfeature/contrib/providers/gcpsecretmanager/FlagCacheCTest.java
```diff
@@ -0,0 +1,80 @@
...
+                FlagCache sharedCache = new FlagCache(Duration.ofMinutes(5), 100, freshClock);
+                sharedCache.put("key", "expired-value");
+                // Make entry expire
+                fresh.set(fresh.get().plus(Duration.ofMinutes(6)));
+
+                Runner.runParallel(
+                        // Thread A: re-insert fresh value for the same key
+                        () -> sharedCache.put("key", "new-value"),
+                        // Thread B: get() triggers expiry-removal of the old entry
+                        () -> sharedCache.get("key"));
```
chrfwow, line 69: I believe this should either return nothing or the new value, and we should assert that

chrfwow:
The build pipeline currently fails for two reasons:
The commits are not signed off. To fix this, you will need to sign off all your changes and force push them on this branch
The type check fails, this can be fixed automatically when you run the spotlessApply command

providers/gcp-secret-manager/src/main/java/dev/openfeature/contrib/providers/gcpsecretmanager/GcpSecretManagerProvider.java
```diff
@@ -0,0 +1,176 @@
...
+                .build();
+    }
+
+    private String fetchWithCache(String key) {
+        String secretName = buildSecretName(key);
+        return cache.get(secretName).orElseGet(() -> {
+            String value = fetchFromGcp(secretName);
+            cache.put(secretName, value);
+            return value;
+        });
```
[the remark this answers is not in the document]
mahpatil: Fixed

providers/gcp/src/main/java/dev/openfeature/contrib/providers/gcp/GcpSecretManagerProviderOptions.java
```diff
@@ -0,0 +1,83 @@
...
+    /**
+     * Validates that required options are present and well-formed.
+     *
+     * @throws IllegalArgumentException when {@code projectId} is null or blank
+     */
+    public void validate() {
+        if (projectId == null || projectId.trim().isEmpty()) {
+            throw new IllegalArgumentException("GcpSecretManagerProviderOptions: projectId must not be blank");
+        }
+    }
```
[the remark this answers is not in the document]
mahpatil: fixed

providers/gcp-secret-manager/src/main/java/dev/openfeature/contrib/providers/gcpsecretmanager/FlagValueConverter.java
mahpatil: fixed

providers/gcp/src/main/java/dev/openfeature/contrib/providers/gcp/FlagCache.java
```diff
@@ -0,0 +1,98 @@
...
+        CacheEntry entry = store.get(key);
+        if (entry == null) {
+            return Optional.empty();
+        }
+        if (entry.isExpired()) {
+            store.remove(key);
+            return Optional.empty();
+        }
+        return Optional.of(entry.value);
+    }
```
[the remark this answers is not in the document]
mahpatil: fixed

providers/gcp/src/test/java/dev/openfeature/contrib/providers/gcp/FlagValueConverterTest.java
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    @Nested
+    @DisplayName("Integer conversion")
+    class IntegerConversion {
+
+        @Test
+        @DisplayName("converts numeric string to Integer")
+        void numericString() {
+            assertThat(FlagValueConverter.convert("42", Integer.class)).isEqualTo(42);
```
mahpatil: @chrfwow  there is already negativeNumericString below?

providers/gcp-secret-manager/src/test/java/dev/openfeature/contrib/providers/gcpsecretmanager/GcpSecretManagerProviderIntegrationTest.java
mahpatil: removed unnecessary check

mahpatil:
> The build pipeline currently fails for two reasons: The commits are not signed off. To fix this, you will need to sign off all your changes and force push them on this branch The type check fails, this can be fixed automatically when you run the spotlessApply command

Signed off commits to fix DCO

providers/gcp-secret-manager/src/test/java/dev/openfeature/contrib/providers/gcpsecretmanager/FlagCacheCTest.java
mahpatil: i think this has changed now. getAfterTtlExpiry is the only one that uses timedCache to check if cache is empty after TTL.

providers/gcp-secret-manager/src/test/java/dev/openfeature/contrib/providers/gcpsecretmanager/FlagCacheCTest.java
mahpatil: updated

providers/gcp/src/test/java/dev/openfeature/contrib/providers/gcp/FlagValueConverterTest.java
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    @Nested
+    @DisplayName("Double conversion")
+    class DoubleConversion {
+
+        @Test
+        @DisplayName("converts numeric string to Double")
+        void numericString() {
+            assertThat(FlagValueConverter.convert("3.14", Double.class)).isEqualTo(3.14);
```
mahpatil: added exponentialNotation

providers/gcp-secret-manager/src/test/java/dev/openfeature/contrib/providers/gcpsecretmanager/FlagCacheCTest.java
mahpatil: This may be an outdated comment now based on new quote. Thread A's value should not have expired regardless of the order so it should stay as get-check-remove now synchronized.

providers/gcp/src/test/java/dev/openfeature/contrib/providers/gcp/FlagCacheTest.java
```diff
@@ -0,0 +1,91 @@
...
+import java.time.Instant;
+import java.time.ZoneOffset;
+import java.util.Optional;
+import java.util.concurrent.atomic.AtomicReference;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+
+@DisplayName("FlagCache")
+class FlagCacheTest {
```
mahpatil: added concurrentExpiryAndInsertDoNotLoseNewEntry to FlagCacheCTest.java

mahpatil:
I hope all comments have been addressed now. Please let me know if there's anything outstanding.

providers/gcp-secret-manager/src/test/java/dev/openfeature/contrib/providers/gcpsecretmanager/FlagCacheCTest.java
```diff
@@ -0,0 +1,148 @@
...
+                return this;
+            }
+
+            @Override
+            public Instant instant() {
+                return now.get();
+            }
+        };
+
+        FlagCache cache = new FlagCache(Duration.ofSeconds(30), 100, clock);
```
chrfwow: Why do we create a cache, clock and now when all of that is already done in the setup method?

mahpatil:
Good catch — removed the duplicate `now`, `clock`, and `cache` setup from `concurrentExpiryAndInsertDoNotLoseNewEntry` and switched to the `T0`/`T1` constants and instance fields already initialised in `@BeforeEach`.

toddbaert:
Re: #1771 (cc @aepfli @chrfwow @mahpatil)

Looking at #1771 + #1772 side by side, I think there's been a small misunderstanding of @aepfli's earlier feedback. IIUC he asked for the *PR* to be split for review hygiene ("a pull request is small, and provides just one additional feature"), not necessarily for two separate Maven modules. The split-PRs ask has been honoured; the split-*modules* part was an additional interpretation.

@chrfwow already raised the same question on #1771. Posting here too since this is where active review is happening, and IMO it's much cheaper to settle before #1772 merges than after.

Let's keep 2 PRs but land them in one module.

- The two PRs duplicate ~80% of code (`FlagCache`, `FlagValueConverter`, the provider scaffolding, tests). They're already drifting (Clock injection, `STATIC` vs `CACHED`, exception cause, log4j versions).
- The GCP client deps are technically different artifacts, but their transitive trees overlap heavily (gax, gRPC, protobuf, auth, etc.); the incremental footprint of having both on the classpath is small.
- One module = one CHANGELOG, one release, one set of internals to maintain. Two modules = ongoing duplication tax. A third `gcp-common` module is overkill for what's effectively one shared `FlagCache`.

@mahpatil - to keep things reviewable, suggest stacking the branches: rework this PR (#1772) to add a single `providers/gcp` module with `pom.xml`, shared internals (`FlagCache`, `FlagValueConverter`), and the Secret Manager provider. Then rebase #1771 on top, so it only adds the Parameter Manager provider on top of the shared scaffolding. SM-first makes sense since this PR is closer to approval. That way each PR stays small and focused, and the shared scaffolding only shows up in the diff once.

Suggested module name: `providers/gcp` (artifact `gcp`). Short, accurate, and leaves room if other GCP-backed providers ever land later. `gcp-secret-and-parameter-manager` works too but is mouthful-y. WDYT?

@aepfli - was your earlier ask only about the PR split, or did you also intend two artifacts? Happy to discuss if there's a packaging reason I'm missing.

mahpatil:
> Re: #1771 (cc @aepfli @chrfwow @mahpatil)
> 
> Looking at #1771 + #1772 side by side, I think there's been a small misunderstanding of @aepfli's earlier feedback. IIUC he asked for the _PR_ to be split for review hygiene ("a pull request is small, and provides just one additional feature"), not necessarily for two separate Maven modules. The split-PRs ask has been honoured; the split-_modules_ part was an additional interpretation.
> 
> @chrfwow already raised the same question on #1771. Posting here too since this is where active review is happening, and IMO it's much cheaper to settle before #1772 merges than after.
> 
> Let's keep 2 PRs but land them in one module.
> 
> * The two PRs duplicate ~80% of code (`FlagCache`, `FlagValueConverter`, the provider scaffolding, tests). They're already drifting (Clock injection, `STATIC` vs `CACHED`, exception cause, log4j versions).
> * The GCP client deps are technically different artifacts, but their transitive trees overlap heavily (gax, gRPC, protobuf, auth, etc.); the incremental footprint of having both on the classpath is small.
> * One module = one CHANGELOG, one release, one set of internals to maintain. Two modules = ongoing duplication tax. A third `gcp-common` module is overkill for what's effectively one shared `FlagCache`.
> 
> @mahpatil - to keep things reviewable, suggest stacking the branches: rework this PR (#1772) to add a single `providers/gcp` module with `pom.xml`, shared internals (`FlagCache`, `FlagValueConverter`), and the Secret Manager provider. Then rebase #1771 on top, so it only adds the Parameter Manager provider on top of the shared scaffolding. SM-first makes sense since this PR is closer to approval. That way each PR stays small and focused, and the shared scaffolding only shows up in the diff once.
> 
> Suggested module name: `providers/gcp` (artifact `gcp`). Short, accurate, and leaves room if other GCP-backed providers ever land later. `gcp-secret-and-parameter-manager` works too but is mouthful-y. WDYT?
> 
> @aepfli - was your earlier ask only about the PR split, or did you also intend two artifacts? Happy to discuss if there's a packaging reason I'm missing.

This is a bit frustrating as I have been receiving feedback on the PR and now requested to refactor again. @aepfli please finalize with @chrfwow what you would want, I don't want to be keep coming back and incorporating comments, wasting everyones time. I feel the process very slow even though we are using gemini and im using claude, this PR has been going back and forth for a while. 
Please provide concrete steps that I can incorporate quickly so we can merge this PR.

aepfli:
@mahpatil first of all, I want to apologise. Reading back through the thread, I can absolutely see how
my original comment landed the way it did, and the frustration you are voicing is completely fair. You
have been responsive, you have been thorough, and the last thing I want is for a contributor putting in
this kind of effort to feel that the goalposts keep moving. That is on me for not being clearer the
first time, and I am sorry for the churn it has caused.

Let me try to clear up the misunderstanding, because I think @toddbaert read my intent correctly. When I
 asked you to split the work, I was talking about splitting the pull request for review hygiene —
keeping each PR small and focused on one feature so that we as maintainers can evaluate it properly. I
was not asking for two separate Maven modules or two separate artifacts. That part was an interpretation
 on top of my ask, and I should have spelled it out at the time. The duplication that has surfaced
between #1771 and #1772 is exactly the kind of thing the "one module" approach avoids.

The technique that makes "split the PR" and "share the code" coexist is what is usually called stacked
diffs (or stacked PRs). The idea is simple: instead of two independent branches off main, you stack
them. PR A branches off main and contains the shared scaffolding plus the first provider. PR B then
branches off PR A's branch (not main) and adds only the second provider on top. Each PR stays small and
reviewable on its own, the shared code only shows up in the diff once, and once PR A merges, PR B's diff
 naturally collapses to just its own delta. It is the same review benefit I was originally asking for,
without paying the duplication tax.

Concretely, for your case I think Todd's suggestion is exactly right: rework #1772 to add a single
providers/gcp module containing the shared internals (FlagCache, FlagValueConverter, scaffolding) and
the Secret Manager provider, then rebase #1771 on top of that branch so it only adds the Parameter
Manager provider. SM-first makes sense since this PR is closer to approval anyway.

Again, I am sorry for the back-and-forth — your contribution is genuinely appreciated and we want to
land it. If it would help, I am happy to jump on a quick call or pair on the rebase so we can get this
over the line without another round of comments.

mahpatil:
> @mahpatil first of all, I want to apologise. Reading back through the thread, I can absolutely see how my original comment landed the way it did, and the frustration you are voicing is completely fair. You have been responsive, you have been thorough, and the last thing I want is for a contributor putting in this kind of effort to feel that the goalposts keep moving. That is on me for not being clearer the first time, and I am sorry for the churn it has caused.
> 
> Let me try to clear up the misunderstanding, because I think @toddbaert read my intent correctly. When I asked you to split the work, I was talking about splitting the pull request for review hygiene — keeping each PR small and focused on one feature so that we as maintainers can evaluate it properly. I was not asking for two separate Maven modules or two separate artifacts. That part was an interpretation on top of my ask, and I should have spelled it out at the time. The duplication that has surfaced between #1771 and #1772 is exactly the kind of thing the "one module" approach avoids.
> 
> The technique that makes "split the PR" and "share the code" coexist is what is usually called stacked diffs (or stacked PRs). The idea is simple: instead of two independent branches off main, you stack them. PR A branches off main and contains the shared scaffolding plus the first provider. PR B then branches off PR A's branch (not main) and adds only the second provider on top. Each PR stays small and reviewable on its own, the shared code only shows up in the diff once, and once PR A merges, PR B's diff naturally collapses to just its own delta. It is the same review benefit I was originally asking for, without paying the duplication tax.
> 
> Concretely, for your case I think Todd's suggestion is exactly right: rework #1772 to add a single providers/gcp module containing the shared internals (FlagCache, FlagValueConverter, scaffolding) and the Secret Manager provider, then rebase #1771 on top of that branch so it only adds the Parameter Manager provider. SM-first makes sense since this PR is closer to approval anyway.
> 
> Again, I am sorry for the back-and-forth — your contribution is genuinely appreciated and we want to land it. If it would help, I am happy to jump on a quick call or pair on the rebase so we can get this over the line without another round of comments.

okay thanks for clarifying. I have now consolidated this into single provider for GCP. Do we also want to have single factory and tests to instantiate & test both the providers? Happy to incorporate that here or when merging Parameter manager.

providers/gcp/src/main/java/dev/openfeature/contrib/providers/gcp/GcpSecretManagerProvider.java
```diff
@@ -0,0 +1,185 @@
...
+
+    @Override
+    public Metadata getMetadata() {
+        return () -> PROVIDER_NAME;
+    }
+
+    @Override
+    public void initialize(EvaluationContext evaluationContext) throws Exception {
+        options.validate();
+        if (client == null) {
```
chrfwow, line 81: when would the client not be null here?

mahpatil: Is that the only comment? How about rest of the design/refactor? Or do we need a call to review these changes.

mahpatil: When using this constructor? client might be null
    public GcpSecretManagerProvider(GcpProviderOptions options) {
        this.options = options;
    }

mahpatil:
@chrfwow @aepfli just checking next steps on this? I dont see any responses to my comments or updated PR. Are you still keen to include this contribution?

chrfwow (approved):

chrfwow:
The CI fails because there are some formatting issue. You can fix them by running spotless:apply

mahpatil:
Fixed formatting issues.

providers/gcp/README.md
```diff
@@ -0,0 +1,121 @@
...
+<!-- x-release-please-end-version -->
+
+## Quick Start
+
+```java
+import dev.openfeature.contrib.providers.gcp.GcpSecretManagerProvider;
+import dev.openfeature.contrib.providers.gcp.GcpSecretManagerProviderOptions;
+import dev.openfeature.sdk.OpenFeatureAPI;
+
+GcpSecretManagerProviderOptions options = GcpSecretManagerProviderOptions.builder()
```
toddbaert:
```suggestion
GcpProviderOptions options = GcpSecretManagerProviderOptions.builder()
```

I think this name is wrong.

providers/gcp/src/main/java/dev/openfeature/contrib/providers/gcp/GcpProviderOptions.java
```diff
@@ -0,0 +1,86 @@
...
+     * Validates that required options are present and well-formed.
+     *
+     * @throws IllegalArgumentException when {@code projectId} is null or blank
+     */
+    public void validate() {
+        if (projectId == null || projectId.trim().isEmpty()) {
+            throw new IllegalArgumentException("GcpSecretManagerProviderOptions: projectId must not be blank");
+        }
+        if (version == null || version.trim().isEmpty()) {
+            throw new IllegalArgumentException("GcpSecretManagerProviderOptions: secretVersion must not be blank");
```
toddbaert:
```suggestion
            throw new IllegalArgumentException("GcpSecretManagerProviderOptions: version must not be blank");
```

providers/gcp/README.md
```diff
@@ -0,0 +1,121 @@
...
+    .build();
+```
+
+## Configuration Options
+
+| Option | Type | Default | Description |
+|--------|------|---------|-------------|
+| `projectId` | `String` | *(required)* | GCP project ID that owns the secrets |
+| `credentials` | `GoogleCredentials` | `null` (ADC) | Explicit credentials; falls back to Application Default Credentials when null |
+| `secretVersion` | `String` | `"latest"` | Secret version to access. Use `"latest"` for the current version or a numeric string (e.g. `"3"`) to pin to a specific version |
```
toddbaert:
```suggestion
| `version` | `String` | `"latest"` | Secret version to access. Use `"latest"` for the current version or a numeric string (e.g. `"3"`) to pin to a specific version |
```

samples/gcp/setup.sh
```diff
@@ -0,0 +1,70 @@
...
+# Object flag: structured checkout configuration (JSON)
+create_secret "checkout-config" '{"paymentMethods":["card","paypal"],"expressCheckout":true,"maxRetries":3}'
+
+echo ""
+echo "✓ All secrets created successfully."
+echo ""
+echo "Next steps:"
+echo "  1. Authenticate: gcloud auth application-default login"
+echo "  2. Run the sample:"
+echo "       cd gcp-sample"
```
toddbaert:
```suggestion
echo "       cd samples/gcp"
```

I think?

providers/gcp/pom.xml
```diff
@@ -0,0 +1,116 @@
...
+            <url>https://openfeature.dev/</url>
+        </developer>
+    </developers>
+
+    <dependencies>
+        <!-- GCP Secret Manager client -->
+        <dependency>
+            <groupId>com.google.cloud</groupId>
+            <artifactId>google-cloud-secretmanager</artifactId>
+            <version>2.57.0</version>
```
toddbaert, line 43: nit: consider importing the [GCP Libraries BOM](https://cloud.google.com/java/docs/bom) in `<dependencyManagement>` instead of pinning `google-cloud-secretmanager` directly. The BOM aligns the entire Google Cloud transitive tree (gax, grpc, protobuf, google-auth, and Jackson via `google-http-client-jackson2`), avoiding version skew. `jackson-databind` below could then drop its pin too. Optional now, more compelling once Parameter Manager lands in the same module.

toddbaert (approved):
Approved. I have some recommendations that I think are mostly around doc, and one about potentially improving the deps with a BOM ref, but I will leave that up to you. If I don't hear anything in the next 24 hrs, I will merge; otherwise feel free to make those minor fixes or tell me to hold off.

Thanks again for your contribution and patience.

PS: I made a small change to revert whitespace noise in the parent POM.

providers/gcp/pom.xml
mahpatil: Thank you. I will incorporate it as part of parameter manager PR coming next.

mahpatil:
> Approved. I have some recommendations that I think are mostly around doc, and one about potentially improving the deps with a BOM ref, but I will leave that up to you. If I don't hear anything in the next 24 hrs, I will merge; otherwise feel free to make those minor fixes or tell me to hold off.
> 
> Thanks again for your contribution and patience.
> 
> PS: I made a small change to revert whitespace noise in the parent POM.

Thank you @toddbaert, only one outstanding option on use of BOM, will incorporate it as part of parameter mgr, which is coming next. If you can now merge I can start on Parameter mgr which is really the more common usecase for us.

toddbaert:
FYI I added you as a "component_owner" for this, which will add you as assignee/reviewer to PRs in this folder.

PR #1773, opened 2026-04-11
feat(flagd): log error on invalid env var parsing

### LINKED ISSUE #1673: [flagd] Add log output on invalid env vars
When flagd options in env vars cannot be parsed, we should log.
Currently we silently fall back to a default value.

TBD: Which log level is appropriate?

I'm in favor for `ERROR`. Reason: It's a misconfiguration that can lead to unexpected behavior and requires user intervention.

_Originally posted by @chrfwow in https://github.com/open-feature/java-sdk-contrib/pull/1624#discussion_r2626741473_

buildingisfun23:
## Summary
- Adds ERROR-level log output when environment variables cannot be parsed, instead of silently falling back to defaults
- Covers integer, long, and list parsing in `Config.java`
- Uses the existing `@Slf4j` logger already present on the class

Fixes #1673

### DIFF
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/Config.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/Config.java
@@ -85,6 +85,11 @@ public final class Config {
         try {
             return System.getenv(key) != null ? Integer.parseInt(System.getenv(key)) : defaultValue;
         } catch (Exception e) {
+            log.error(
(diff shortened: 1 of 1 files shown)


leakonvalinka (commented):
Hi, this looks good, please consider integrating the changes gemini suggested (i.e. adding the exception to the log message). Also, all commits need to be signed off for the DCO check to pass, should be fixed if you do `git commit --amend --signoff` :)

buildingisfun23:
Closing as per request.

PR #1789, opened 2026-05-08
fix: tight loop in RPC mode event stream reconnect

toddbaert:
* the in-process sync stream has a [spec](https://flagd.dev/reference/specifications/providers/#stream-reconnection)-compliant delay of maxStreamBackoff, but the RPC does not
* this backoff is crucial to prevent tight loops sometimes cauesd by immediate error responses, often from proxies
* this adds the same backoff to RPC, plus tests, and adds same naming improvements to existing tests

FYI: a lot of the diff in non-test code is whitespace

### DIFF
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/rpc/RpcResolver.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/rpc/RpcResolver.java
@@ -12,6 +12,7 @@ import dev.openfeature.contrib.providers.flagd.resolver.Resolver;
 import dev.openfeature.contrib.providers.flagd.resolver.common.ChannelBuilder;
 import dev.openfeature.contrib.providers.flagd.resolver.common.ChannelConnector;
 import dev.openfeature.contrib.providers.flagd.resolver.common.QueueingStreamObserver;
+import dev.openfeature.contrib.providers.flagd.resolver.common.ShutdownUtils;
 import dev.openfeature.contrib.providers.flagd.resolver.common.StreamResponseModel;
 import dev.openfeature.contrib.providers.flagd.resolver.rpc.cache.Cache;
 import dev.openfeature.contrib.providers.flagd.resolver.rpc.strategy.ResolveFactory;
@@ -144,6 +145,8 @@ public final class RpcResolver implements Resolver {
             return;
         }
         this.retryScheduler.shutdownNow();
+        ShutdownUtils.awaitTerminationQuietly(
+                () -> retryScheduler.awaitTermination(options.getDeadline(), TimeUnit.MILLISECONDS));
         this.connector.shutdown();
     }
 
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/rpc/RpcResolver.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/rpc/RpcResolver.java
@@ -46,7 +46,10 @@ import io.grpc.StatusRuntimeException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.Executors;
 import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.RejectedExecutionException;
+import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Function;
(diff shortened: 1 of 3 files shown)


providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/rpc/RpcResolver.java
```diff
@@ -141,6 +143,7 @@ public void shutdown() throws Exception {
         if (shutdown.getAndSet(true)) {
             return;
         }
+        this.retryScheduler.shutdownNow();
```
chrfwow, line 147: do we also want to `awaitTermination` on the `retryScheduler`

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/rpc/RpcResolver.java
```diff
@@ -337,63 +340,81 @@ private void restartStream() {
...
-                    this.handleErrorOrComplete(false);
-                    // The stream is complete, we still try to reconnect
-                    break;
+            try {
+                // explicit backoff after stream errors to prevent tight loops when errors are returned immediately
+                // (e.g., by intervening proxies like Envoy)
+                if (shouldThrottle.getAndSet(false)) {
+                    log.debug("Previous stream ended with error, waiting {} ms before retry", this.maxBackoffMs);
+                    try {
+                        retryScheduler.schedule(this::observeEventStream, this.maxBackoffMs, TimeUnit.MILLISECONDS);
```
chrfwow, line 359: In case this `schedule` fails, we will still return and never attempt to call `observeEventStream` again. Should we only return in the non-exceptional case?

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/resolver/rpc/RpcResolverTest.java
```diff
@@ -61,6 +62,29 @@ public Void answer(InvocationOnMock invocation) {
...
+
+        // stub that immediately fires onError on every eventStream call
+        errorStub = mock(ServiceStub.class);
+        when(errorStub.withDeadlineAfter(anyLong(), any())).thenReturn(errorStub);
+        doAnswer((Answer<Void>) invocation -> {
+                    @SuppressWarnings("unchecked")
+                    QueueingStreamObserver<EventStreamResponse> obs = (QueueingStreamObserver<EventStreamResponse>)
+                            invocation.getArguments()[1];
+                    latch.countDown();
+                    // immediately fire error on a separate thread
```
chrfwow: Waiting 10ms is not exactly `immediately`

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/resolver/rpc/RpcResolverTest.java
```diff
@@ -61,6 +62,29 @@ public Void answer(InvocationOnMock invocation) {
...
+                    new Thread(() -> {
+                                try {
+                                    Thread.sleep(10);
+                                    obs.onError(new Exception("immediate error"));
```
chrfwow: Should we move the `obs.onError` out of the `try` block, so that we still emit the error even if the thread gets interrupted?

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/resolver/rpc/RpcResolverTest.java
```diff
@@ -61,6 +62,29 @@ public Void answer(InvocationOnMock invocation) {
...
                 .eventStream(any(), any()); // Mock the initialize method
+
+        // stub that immediately fires onError on every eventStream call
+        errorStub = mock(ServiceStub.class);
+        when(errorStub.withDeadlineAfter(anyLong(), any())).thenReturn(errorStub);
+        doAnswer((Answer<Void>) invocation -> {
+                    @SuppressWarnings("unchecked")
+                    QueueingStreamObserver<EventStreamResponse> obs = (QueueingStreamObserver<EventStreamResponse>)
+                            invocation.getArguments()[1];
+                    latch.countDown();
```
chrfwow: Why is the latch counted down here? Wouldn't it make more sense to count it down after the `onError` was invoked

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/rpc/RpcResolver.java
toddbaert: I could be missing something, but I think the _only_ way this could fail (we get a `RejectedExecutionException`) is if we are shut down... so there's no reason to do anything here but return. The behavior is the same in the [SyncStreamQueueSource](https://github.com/open-feature/java-sdk-contrib/blob/c2fb0fee110090406ee6f0a8adfa333e6af9e690/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamQueueSource.java#L193-L199).

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/resolver/rpc/RpcResolverTest.java
toddbaert: I think if the thread is interrupted, something odd is happening, so I'm just throwing a runtime ex now to fail fast.

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/resolver/rpc/RpcResolverTest.java
toddbaert: Changed to "shortly" in all places.

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/resolver/rpc/RpcResolverTest.java
toddbaert: I agree. Updated

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/rpc/RpcResolver.java
toddbaert: Good catch. I've used `ShutdownUtils.awaitTerminationQuietly`, a util we had from the `SyncStreamQueueSource`.

I should say - I tried to make that and this common, based on some base-class, but it got messy and I just decided the small amount of duplication here wasn't worth the refactor. The classes aren't 1:1.

jonathannorris (approved):

PR #1790, opened 2026-05-09
fix(flagd): add logging for invalid env var parsing failures

IshitaSharma08:
Added error logging when environment variable parsing fails in Config.java.

Previously invalid environment variables silently fell back to default values without notifying the user.
This change logs the parsing failure before returning the default value.

### DIFF
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/Config.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/Config.java
@@ -1,11 +1,10 @@
 package dev.openfeature.contrib.providers.flagd;
 
+import dev.openfeature.contrib.providers.flagd.resolver.rpc.cache.CacheType;
 import java.util.Arrays;
 import java.util.List;
 import java.util.function.Function;
 import java.util.stream.Collectors;
-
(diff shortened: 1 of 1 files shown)


providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/Config.java
```diff
@@ -1,10 +1,11 @@
 package dev.openfeature.contrib.providers.flagd;
 
-import dev.openfeature.contrib.providers.flagd.resolver.rpc.cache.CacheType;
 import java.util.Arrays;
 import java.util.List;
 import java.util.function.Function;
 import java.util.stream.Collectors;
+
+import dev.openfeature.contrib.providers.flagd.resolver.rpc.cache.CacheType;
```
chrfwow: Moving this import violates a checkstyle rule and makes the CI pipeline build fail. Please revert it or run `spotless:apply`

chrfwow:
@IshitaSharma08 it seems like you forgot to sign-off at least one commit, that's why the DCO action fails. Please sign off your commits and force push them

toddbaert:
Looks abandoned, closing for now.

PR #1651, opened 2025-11-20
feat: Add FLAGD_SYNC_PORT support for in-process providers with backwards compatibility

carla-was-not-available:
## This PR

- adds this new feature

### Related Issues

Fixes #1234523

### How to test

resolves: #1619

PR #1798, opened 2026-05-18
fix(flagd): validate FLAGD_SYNC_PORT before parse to avoid kubelet service-link collision

### LINKED ISSUE #1797: flagd-provider-java reads FLAGD_SYNC_PORT in in-process mode, colliding with Kubernetes service-link env injection
## Summary

`FlagdOptions.prebuild()` (introduced in #1651, released in `dev.openfeature.contrib.providers.flagd` v0.11.19) reads `FLAGD_SYNC_PORT` first when `FLAGD_RESOLVER=in-process`, falling back to `FLAGD_PORT` only when `FLAGD_SYNC_PORT` is unset. The value is passed to `Integer.parseInt` without validation.

When the SDK is used in a Kubernetes pod and a `Service` named `flagd-sync` exists in the same namespace, kubelet's [service-link environment variable injection](https://kubernetes.io/docs/concepts/services-networking/service/#environment-variables) populates the workload container with `FLAGD_SYNC_PORT=tcp://<clusterIP>:<port>`. The SDK then crashes at startup with:

```
java.lang.NumberFormatException: For input string: "tcp://172.20.136.125:8015"
	at java.base/java.lang.Integer.parseInt(Integer.java:662)
	at dev.openfeature.contrib.providers.flagd.FlagdOptions$FlagdOptionsBuilder.prebuild(FlagdOptions.java:321)
	at dev.openfeature.contrib.providers.flagd.FlagdProvider.<init>(FlagdProvider.java:71)
```

This affects every JVM consumer using `FLAGD_RESOLVER=in-process` when deployed alongside the upstream flagd Helm chart or via the [OpenFeature Operator's `InProcessConfiguration` pattern](https://github.com/open-feature/open-feature-operator/blob/main/docs/in_process_configuration.md) — i.e. the documented happy-path deployment topologies.

## Reproduction

1. Deploy flagd in namespace `N` using the upstream flagd Helm chart (creates `Service/flagd-sync` on port 8015).
2. Deploy a JVM workload in namespace `N` with `FLAGD_RESOLVER=in-process`, `FLAGD_HOST=flagd-sync.N.svc.cluster.local`, `FLAGD_PORT=8015`, and `enableServiceLinks: true` (the default).
3. Start the workload.

**Expected:** provider initializes against `flagd-sync:8015`.
**Actual:** `NumberFormatException` from `FlagdOptions$FlagdOptionsBuilder.prebuild` because kubelet has populated `FLAGD_SYNC_PORT` with the Service's URL before the SDK reads it.

## Affected versions

- v0.11.19 (first release after #1651, 2026-01-06) through v0.13.3 (current latest at filing, 2026-05-11).
- v0.11.18 and earlier read only `FLAGD_PORT` and are unaffected by this specific bug. (Note: v0.11.18 has a separate in-process cold-start hang fixed in v0.11.19, so rolling back is not a viable workaround.)

## Why this is a bug, not a config issue

1. **It collides with the upstream chart's own Service name.** `flagd-sync` is the name produced by the flagd Helm chart and the OpenFeature Operator. The SDK is incompatible with its own deployment story.
2. **The env var name is already reserved.** `FLAGD_SYNC_PORT` is the flagd *daemon's* env var, derived from its `--sync-port` CLI flag (see the [flagd Options wiki](https://github.com/open-feature/flagd/wiki/Options)). Using it for a client-side config too creates a server/client semantic collision in any config system that sets env vars uniformly.
3. **Other-language SDKs don't do this.** The Node.js and Go in-process providers don't read `FLAGD_SYNC_PORT` at all — they read `FLAGD_PORT` only. PR #1651 was a Java-only addition not mirrored in sibling SDKs.
4. **No integer validation before parse.** Even setting the name issue aside, `Integer.parseInt` on an env var of unknown provenance is fragile; a URL-format value is a strong indicator of service-link injection, not user intent.

## Suggested fix (PR forthcoming)

Validate that `FLAGD_SYNC_PORT` parses as an integer in `[1, 65535]` before using it. If it doesn't, log at WARN with a pointer to the most likely root cause (service-link injection) and fall back to `FLAGD_PORT`. Minimal change, backwards-compatible for users who legitimately set `FLAGD_SYNC_PORT`.

Longer-term considerations worth a separate discussion (not in this PR):
- Whether the SDK should read `FLAGD_SYNC_PORT` at all, given the name is already reserved by the flagd daemon and the JS/Go SDKs don't read it.
- Cross-language alignment in the OpenFeature spec for in-process sync port configuration.

## Environment

- `dev.openfeature.contrib.providers:flagd` 0.11.19 — 0.13.3
- OpenFeature Java SDK 1.19.0 (any version in the provider's `[1.0, 2.0)` range)
- Kubernetes ≥ v1.6 (any version with service-link env injection)
- Reproduces independent of OS / JVM version / flagd server version

## Workarounds

Until the SDK is fixed, affected users can:
- Set `enableServiceLinks: false` on JVM workload pod templates (per-pod, easy to lose on chart changes).
- Rename the upstream `flagd-sync` Service so its derived env var prefix doesn't match `FLAGD_SYNC_*` (chart values-only change if the chart exposes the name).
- Deploy flagd in a sibling namespace from JVM workloads (restores the standard cross-namespace topology, no env vars synthesized).

jabenedicic:
Closes #1797

## Summary

`FlagdOptions.prebuild()` parses `FLAGD_SYNC_PORT` directly with `Integer.parseInt` when `FLAGD_RESOLVER=in-process`. When a `Service` named `flagd-sync` shares the pod's namespace (the exact topology produced by the upstream flagd Helm chart and the OpenFeature Operator's `InProcessConfiguration` pattern), kubelet's service-link env injection populates the workload with `FLAGD_SYNC_PORT=tcp://<clusterIP>:<port>`, the SDK calls `Integer.parseInt("tcp://...")`, and the consumer crashes at startup with `NumberFormatException`.

Full repro, affected versions, and root-cause analysis in #1797.

## Change

Validate that `FLAGD_SYNC_PORT` parses as an integer in `[1, 65535]` before using it. On invalid input, log at WARN with a pointer to the most likely cause (service-link injection) and fall back to `FLAGD_PORT`. Behaviour for valid, user-set `FLAGD_SYNC_PORT` is unchanged.

```java
// Before
String portValue = resolverType == Config.Resolver.IN_PROCESS
        ? fallBackToEnvOrDefault(Config.SYNC_PORT_ENV_VAR_NAME, fromPortEnv)
        : fromPortEnv;
port = Integer.parseInt(portValue);

// After (sketch — see diff for full code)
String portValue = fromPortEnv;
if (resolverType == Config.Resolver.IN_PROCESS) {
    String syncPortValue = fallBackToEnvOrDefault(Config.SYNC_PORT_ENV_VAR_NAME, null);
    if (syncPortValue != null) {
        if (isValidPort(syncPortValue)) {
            portValue = syncPortValue;
        } else {
            log.warn("Ignoring {} value '{}' (not a valid port); falling back to {} ('{}'). ...",
                    Config.SYNC_PORT_ENV_VAR_NAME, syncPortValue,
                    Config.PORT_ENV_VAR_NAME, fromPortEnv);
        }
    }
}
port = Integer.parseInt(portValue);
```

## Why this is the minimal change

The narrower fix (validation only) was preferred over alternatives:

- **Stop reading `FLAGD_SYNC_PORT` entirely.** Would match Node/Go SDK behaviour and avoid the env-var name reservation issue, but is a behaviour change for users who legitimately set `FLAGD_SYNC_PORT` after #1651. Worth discussing separately.
- **Rename the env var.** Cleanest but breaking; needs deprecation path.

Validation is fully backwards-compatible: legitimately set `FLAGD_SYNC_PORT=NNNN` continues to take precedence over `FLAGD_PORT` exactly as today.

## Tests

Three new cases in `FlagdOptionsTest`, all using the existing `@SetEnvironmentVariable` pattern:

- `testInProcessProvider_invalidSyncPortFallsBackToFlagdPort` — `FLAGD_SYNC_PORT=tcp://10.0.0.1:8015` + `FLAGD_PORT=8888` → port 8888 (the exact service-link injection scenario).
- `testInProcessProvider_invalidSyncPortWithNoFlagdPortUsesDefault` — invalid `FLAGD_SYNC_PORT`, no `FLAGD_PORT` → `DEFAULT_IN_PROCESS_PORT`.
- `testInProcessProvider_outOfRangeSyncPortFallsBackToFlagdPort` — `FLAGD_SYNC_PORT=99999` → falls back (port out of valid range).

Existing tests for `FLAGD_SYNC_PORT` precedence and fallback are unchanged and continue to pass. Full suite: 21 tests, 0 failures, 0 errors.

`mvn spotless:apply` reports no formatting changes required.

## Out of scope

- Whether the SDK should read `FLAGD_SYNC_PORT` at all (cross-language alignment / env-var name reservation discussion) — flagged in #1797 for a separate conversation.

### DIFF
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdOptions.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdOptions.java
@@ -342,6 +342,19 @@ public class FlagdOptions {
                     }
                 }
 
+                if (!isValidPort(portValue)) {
+                    // Last-line-of-defence: FLAGD_PORT itself can be polluted by Kubernetes
+                    // service-link injection too if a Service named `flagd` exists in the
+                    // pod's namespace (FLAGD_PORT=tcp://<clusterIP>:8013), which would
+                    // affect RPC-mode consumers identically. Fall back to the resolver's
+                    // default port rather than throwing at parse time.
+                    log.warn(
+                            "Configured port value '{}' is not a valid port; falling back to default '{}'.",
+                            portValue,
+                            defaultPort);
+                    portValue = defaultPort;
+                }
+
                 port = Integer.parseInt(portValue);
             }
         }
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdOptions.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdOptions.java
@@ -17,6 +17,7 @@ import java.util.function.Function;
 import jdk.jfr.Experimental;
 import lombok.Builder;
 import lombok.Getter;
+import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang3.StringUtils;
 
 /**
@@ -24,6 +25,7 @@ import org.apache.commons.lang3.StringUtils;
  */
 @Builder(toBuilder = true)
 @Getter
+@Slf4j
 @SuppressWarnings("PMD.TooManyStaticImports")
 public class FlagdOptions {
 
@@ -314,9 +316,31 @@ public class FlagdOptions {
                 String defaultPort = determineDefaultPortForResolver();
                 String fromPortEnv = fallBackToEnvOrDefault(Config.PORT_ENV_VAR_NAME, defaultPort);
 
-                String portValue = resolverType == Config.Resolver.IN_PROCESS
-                        ? fallBackToEnvOrDefault(Config.SYNC_PORT_ENV_VAR_NAME, fromPortEnv)
-                        : fromPortEnv;
+                String portValue = fromPortEnv;
+                if (resolverType == Config.Resolver.IN_PROCESS) {
+                    String syncPortValue = fallBackToEnvOrDefault(Config.SYNC_PORT_ENV_VAR_NAME, null);
+                    if (syncPortValue != null) {
+                        if (isValidPort(syncPortValue)) {
+                            portValue = syncPortValue;
+                        } else {
+                            // FLAGD_SYNC_PORT is unset by the user but populated with a non-numeric
+                            // value — most commonly Kubernetes service-link env injection from a
+                            // Service named `flagd-sync` in the pod's namespace, which produces
+                            // `FLAGD_SYNC_PORT=tcp://<clusterIP>:<port>`. Fall back to FLAGD_PORT
+                            // rather than failing at parse time.
+                            log.warn(
+                                    "Ignoring {} value '{}' (not a valid port); falling back to {} ('{}'). "
+                                            + "This commonly indicates Kubernetes service-link environment "
+                                            + "variable injection from a Service named 'flagd-sync' in the "
+                                            + "pod's namespace; set enableServiceLinks: false on the pod "
+                                            + "template or rename the Service to avoid the collision.",
+                                    Config.SYNC_PORT_ENV_VAR_NAME,
+                                    syncPortValue,
+                                    Config.PORT_ENV_VAR_NAME,
+                                    fromPortEnv);
+                        }
+                    }
+                }
 
                 port = Integer.parseInt(portValue);
             }
@@ -327,6 +351,18 @@ public class FlagdOptions {
                 return Config.DEFAULT_RPC_PORT;
             }
             return Config.DEFAULT_IN_PROCESS_PORT;
+        }
+    }
+
+    private static boolean isValidPort(String value) {
+        if (value == null) {
+            return false;
+        }
+        try {
+            int parsed = Integer.parseInt(value);
+            return parsed > 0 && parsed <= 65535;
+        } catch (NumberFormatException e) {
+            return false;
         }
     }
 }
--- a/providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/FlagdOptionsTest.java
+++ b/providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/FlagdOptionsTest.java
@@ -274,6 +274,29 @@ class FlagdOptionsTest {
 
     @Test
     @SetEnvironmentVariable(key = RESOLVER_ENV_VAR, value = RESOLVER_RPC)
+    @SetEnvironmentVariable(key = PORT_ENV_VAR_NAME, value = "tcp://10.0.0.1:8013")
+    void testRpcProvider_invalidFlagdPortFallsBackToDefault() {
+        // RPC-mode equivalent of the in-process collision: if a Service named `flagd`
+        // shares the pod's namespace, kubelet injects FLAGD_PORT=tcp://<clusterIP>:8013.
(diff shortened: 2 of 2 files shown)


providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdOptions.java
```diff
@@ -314,9 +316,31 @@ void prebuild() {
...
+                                            + "template or rename the Service to avoid the collision.",
+                                    Config.SYNC_PORT_ENV_VAR_NAME,
+                                    syncPortValue,
+                                    Config.PORT_ENV_VAR_NAME,
+                                    fromPortEnv);
+                        }
+                    }
+                }
 
                 port = Integer.parseInt(portValue);
```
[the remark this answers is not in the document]
jabenedicic: Addressed in 5db2d50 — guarded the final `Integer.parseInt(portValue)` with `isValidPort` and a WARN-log fallback to the resolver's default port, matching the existing pattern in the in-process branch. The same service-link collision can hit RPC-mode pods if a `flagd` Service shares the namespace, so the final check is genuinely useful (not just defensive). Added test cases for: RPC with invalid `FLAGD_PORT` → default RPC port; in-process with both env vars invalid → default in-process port.

aepfli (approved):

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdOptions.java
```diff
@@ -314,9 +316,44 @@ void prebuild() {
                 String defaultPort = determineDefaultPortForResolver();
                 String fromPortEnv = fallBackToEnvOrDefault(Config.PORT_ENV_VAR_NAME, defaultPort);
 
-                String portValue = resolverType == Config.Resolver.IN_PROCESS
-                        ? fallBackToEnvOrDefault(Config.SYNC_PORT_ENV_VAR_NAME, fromPortEnv)
-                        : fromPortEnv;
+                String portValue = fromPortEnv;
```
chrfwow, line 319: Do we need this alias?

jabenedicic: Yes — `portValue` and `fromPortEnv` are intentionally distinct:

- `portValue` is the mutable "chosen port" — it gets reassigned to `syncPortValue` on the valid-`FLAGD_SYNC_PORT` path (line 324) and to `defaultPort` on the final invalid-port fallback (line 355).
- `fromPortEnv` is the stable original `FLAGD_PORT` value, still referenced in the WARN log on line 340 so the operator can see the exact fallback value being chosen.

Collapsing them would either lose that detail from the log message or require recomputing/`getenv`-ing it a second time. Happy to rename `portValue` → `chosenPort` if that makes the intent clearer, but I'd prefer to keep the two variables distinct.

chrfwow (approved):

Issue #1420, opened 2025-06-17
Add Provider Support for Cloud-based Parameter Stores (e.g., AWS SSM, GCP Parameter Manager, Azure App Config)


mahpatil:
**Description:**
OpenFeature currently supports various providers for feature flag management. It would be valuable to extend this support by adding a provider (or a set of providers) for cloud-native parameter stores, which are commonly used in enterprise environments for storing configuration and secrets.

These parameter/configuration stores include:
- AWS Systems Manager Parameter Store
- Google Cloud Parameter/Secret Manager Manager
- Azure App Configuration / Key Vault

Adding support for these services would enable OpenFeature users to:
1. Leverage existing cloud-native infrastructure
2. Dynamically fetch feature flags or configuration values stored in these systems
3. Improve consistency and reduce duplication across infrastructure and application layers

**Proposed Solution:**
Implement a separate providers per cloud that can:

1. Connect securely to the respective parameter/config store
2. Fetch values at runtime (optionally cache them)
3. Map keys to OpenFeature flag format
4. Handle authentication via cloud-native SDKs (IAM roles, service accounts, etc.)

**Benefits:**
1. Seamless integration with existing cloud infrastructure
2. Improved developer experience for cloud-first teams
3. Encourages unified configuration management practices

**Potential Challenges:**

- Differences in data models and access patterns across cloud platforms
- Managing credentials securely across environments
- Standardizing behavior for flag evaluation

**Related Links:**
[AWS Parameter Store](https://docs.aws.amazon.com/systems-manager/latest/userguide/systems-manager-parameter-store.html)
[GCP Parameter Manager](https://cloud.google.com/secret-manager/parameter-manager/docs/overview)
[Azure App Configuration](https://learn.microsoft.com/en-us/azure/azure-app-configuration/overview)

**Notes:**
If there's alignment with the OpenFeature roadmap, I’d be happy to collaborate on a proposal or contribute a reference implementation.

PR #1808, opened 2026-06-05
feat(gcp): add parameter manager support

mahpatil:
## This PR
Adds a GCP Parameter Manager support to the GCP Provider that reads feature flags from GCP Parameter Manager config.
Includes updates to sample application under samples/gcp/ with setup/teardown script
Package: providers/gcp
Class: GcpParameterManagerProvider
Supports all OpenFeature flag types via structured or plain-text secret values
Configurable poll interval and GCP project settings

### Related Issues

Fixes GCP services covered under https://github.com/open-feature/java-sdk-contrib/issues/1420

### Notes

This builds on top of the previous PR that added support for GCP Secret manager.

### How to test

- [ ] Unit tests for GcpParameterManagerProvider
- [ ] Integration test (GcpParameterManagerProviderIntegrationTest) requires a real GCP project — set GCP_PROJECT_ID env var to run
- [ ] Sample app: follow samples/gcp/README.md to run end-to-end


mahpatil:
@toddbaert @aepfli this now adds parameter manager support to gcp provider.

providers/gcp/src/main/java/dev/openfeature/contrib/providers/gcp/AbstractGcpProvider.java
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    protected String fetchWithCache(String key) {
+        String name = buildName(key);
+        Optional<String> cached = cache.get(name);
+        if (cached.isPresent()) {
+            log.debug("Fetching from cache name '{}'", key);
+            return cached.get();
+        }
+        synchronized (cache) {
```
toddbaert, line 112: Two related things:

1. `cache` is reassigned in `initialize()` (line 39) but used as the lock in `fetchWithCache` (line 105). Call `initialize()` twice and readers end up holding different locks. (`FlagCache.get` also locks internally, so the outer `synchronized` is partly redundant.)

2. Nothing stops `initialize()` from being called twice, if I'm reading correctly... if it is, the first client just leaks. The convention elsewhere in this repo is to guard with an `AtomicBoolean` and throw `GeneralError("already initialized")`.

Aadd the guard, and make `cache` `final` (or use a dedicated lock) to fix both... and maybe add a test. WDYT?

providers/gcp/src/main/java/dev/openfeature/contrib/providers/gcp/GcpSecretManagerProvider.java
```diff
@@ -44,130 +40,56 @@
...
-    /**
-     * Accesses the configured version of the named secret from GCP Secret Manager.
-     *
-     * @param secretName the GCP secret name (without project path)
-     * @return the UTF-8 string value of the secret payload
-     * @throws FlagNotFoundError when the secret does not exist
-     * @throws GeneralError      for any other GCP API error
-     */
-    private String fetchFromGcp(String secretName) {
+    protected String fetchFromGcp(String secretName) {
```
toddbaert, line 92: The `catch (Exception e)` below drops the cause; sibling `GcpParameterManagerProvider#fetchFromGcp` keeps it. Let's stay consistent and preserve it (helps when debugging GCP errors).

toddbaert:
Approving; pushed a small fixup to revert the manual version bumps (release-please owns those) and wire samples/gcp/pom.xml into the release-please config so the sample's dep version bumps automatically.

Also opened 2 recommendations... [this one](https://github.com/open-feature/java-sdk-contrib/pull/1808#discussion_r3383095649) in particular I'd really like to see done before release (unless I'm missing something).

Once this is merged, I can merge the pending "release please" PR, which will release the package.

toddbaert (approved):

mahpatil:
> toddbaert

thank you @toddbaert ive incorporated all comments

toddbaert:
Thanks @mahpatil . Will merge and release.

PR #1819, opened 2026-06-30
feat(codereadiness): add hook for version-based flag validation

marcin11858:
Introduce the codereadiness hook to control feature flag evaluation by comparing the application's current version with a required minimum version specified in the flag's metadata. If the comparator returns false the hook returns an error to trigger fallback to the default flag value.

## This PR
Introduces the codereadiness hook under hooks/codereadiness to control feature flag evaluation based on the application's running code version.

### How it works
- Compares the application's current version with the required minimum version specified in the flag's metadata (by default, in the `minCodeVersion` key).
- If the comparison fails (by default, if the current version in SemVer format is lower than the minimum or invalid), the hook throws a `GeneralError` to trigger the SDK's fallback mechanism, returning the flag's default value.
- The hook supports custom comparison logic (via `.comparator(...)`), custom metadata keys (via `.metadataMinVerKey(...)`), and optional strict validation (via `.validationRequired(...)`).

### Notes
- Includes comprehensive JUnit 5 unit tests in `CodeReadinessHookTest.java` and `SemVerComparatorTest.java`.
- Includes a `README.md` with setup, configuration examples, and dependency setup.

### Follow-up Tasks
None.

### How to test
Run the module unit tests using Maven from the project root:
```bash
mvn test -pl hooks/codereadiness
```


hooks/codereadiness/src/main/java/dev/openfeature/contrib/hooks/codereadiness/CodeReadinessHook.java
```diff
@@ -0,0 +1,115 @@
...
+    private static final String DEFAULT_MIN_CODE_VERSION_KEY = "minCodeVersion";
+    private static final boolean DEFAULT_VALIDATION_REQUIRED = false;
+    private static final VersionComparator DEFAULT_VERSION_COMPARATOR = new SemVerComparator();
+
+    private final String currentVersion;
+    private final boolean validationRequired;
+    private final String metadataMinVerKey; 
+    private final VersionComparator comparator;
+
+    private CodeReadinessHook (Builder builder) {
```
oxddr: Consider using AutoValue for building data classes.

hooks/codereadiness/src/main/java/dev/openfeature/contrib/hooks/codereadiness/CodeReadinessHook.java
```diff
@@ -0,0 +1,115 @@
...
+        }
+        if (currentVersion == null || currentVersion.isEmpty()) {
+            if (validationRequired) {
+                throw new GeneralError("missing current version error");
+            }
+            return;
+        }
+        try {
+            if (!comparator.isValid(currentVersion, minCodeVersion)) {
+                throw new GeneralError(String.format("current version: \"%s\" required minimum version: \"%s\" check failed", currentVersion, minCodeVersion));
```
oxddr: The exception thrown when there is version mismach is going to be caught by the catch clause below, which is unusual pattern. What about storing result in a variable and throwing outside try-catch block?

hooks/codereadiness/src/main/java/dev/openfeature/contrib/hooks/codereadiness/VersionComparator.java
```diff
@@ -0,0 +1,13 @@
+package dev.openfeature.contrib.hooks.codereadiness;
+
+public interface VersionComparator {
```
oxddr: Provide documentation what this interface represent and why one might want to implement it.

hooks/codereadiness/src/main/java/dev/openfeature/contrib/hooks/codereadiness/VersionComparator.java
```diff
@@ -0,0 +1,13 @@
...
+  /**
+   * Compare current version with required version.
+   *
+   * @param currentVersion of the application
+   * @param minCodeVersion required minimum version
+   * @return true if currentVersion is greater than or equal to minCodeVersion
+   * @throws Exception if there is a parsing error
+   */
+  boolean isValid(String currentVersion, String minCodeVersion) throws Exception;
```
oxddr: This method doesn't only check whether versions are valid, but actually checks whether runtime is at least at the required version. I'd rename this to accordingly, as the current name is a bit misleading. I'd go for for the explicit `isGreaterOrEqual` or `compare`.

hooks/codereadiness/src/main/java/dev/openfeature/contrib/hooks/codereadiness/CodeReadinessHook.java
marcin11858: Lombok was used for building data classes

hooks/codereadiness/src/main/java/dev/openfeature/contrib/hooks/codereadiness/CodeReadinessHook.java
marcin11858: Fixed

hooks/codereadiness/src/main/java/dev/openfeature/contrib/hooks/codereadiness/VersionComparator.java
marcin11858: JavaDoc added

hooks/codereadiness/src/main/java/dev/openfeature/contrib/hooks/codereadiness/VersionComparator.java
marcin11858: Changed to `compare`

hooks/codereadiness/src/main/java/dev/openfeature/contrib/hooks/codereadiness/CodeReadinessHook.java
```diff
@@ -0,0 +1,95 @@
...
+    private final VersionComparator comparator = DEFAULT_VERSION_COMPARATOR;
+
+    CodeReadinessHook(String currentVersion, boolean strictValidation, String metadataMinVerKey, VersionComparator comparator) {
+        this.currentVersion = Objects.requireNonNull(currentVersion, "codereadiness: currentVersion cannot be null");
+        this.strictValidation = strictValidation;
+        this.metadataMinVerKey = Objects.requireNonNull(metadataMinVerKey, "codereadiness: metadataMinVerKey cannot be null");
+        this.comparator = Objects.requireNonNull(comparator, "codereadiness: comparator cannot be null");
+    }
+
+    public static Builder builder(String currentVersion) {
```
chrfwow, line 61: Neither of the parameters are nullable, so why do we enforce non-null only on this one through the builder factory method?

hooks/codereadiness/src/main/java/dev/openfeature/contrib/hooks/codereadiness/CodeReadinessHook.java
```diff
@@ -0,0 +1,95 @@
...
+    public void after(HookContext ctx, FlagEvaluationDetails details, Map hints) {
+        ImmutableMetadata metadata = details != null ? details.getFlagMetadata() : null;
+        if (metadata == null || metadata.isEmpty()) {
+            if (strictValidation) {
+                throw new GeneralError(String.format("flag metadata is null for flag \"%s\"", ctx.getFlagKey()));
+            }
+            log.debug("flag metadata is null for flag \"{}\", skipping validation", ctx.getFlagKey());
+            return;
+        }
+        Object minVerObj = metadata.asUnmodifiableMap().get(metadataMinVerKey);
```
chrfwow:
```suggestion
        Object minVerObj = metadata.getValue(metadataMinVerKey, Object.class);
```
would this also work? It would reduce the object churn, becaus we don't need to create a view on the HashMap in the background.
Or maybe
```suggestion
        Object minVerObj = metadata.getString(metadataMinVerKey);
```

hooks/codereadiness/src/main/java/dev/openfeature/contrib/hooks/codereadiness/SemVerComparator.java
```diff
@@ -0,0 +1,29 @@
...
+import dev.openfeature.sdk.HookContext;
+import dev.openfeature.sdk.ImmutableMetadata;
+import dev.openfeature.sdk.exceptions.GeneralError;
+import java.util.Map;
+import org.semver4j.Semver;
+
+public class SemVerComparator implements VersionComparator {
+  @Override
+  public boolean compare(String currentVersion, String minCodeVersion) throws Exception {
+    String currentFormatted = currentVersion != null && currentVersion.startsWith("v") ? currentVersion : "v" + currentVersion;
```
chrfwow: I think we should guard the parameters with an `Objects.requireNonNull` or similar before we do any operations on them

hooks/codereadiness/src/main/java/dev/openfeature/contrib/hooks/codereadiness/CodeReadinessHook.java
```diff
@@ -0,0 +1,95 @@
...
+
+@Builder(builderMethodName = "", builderClassName = "Builder")
+@Slf4j
+public class CodeReadinessHook implements Hook {
+
+    private static final String DEFAULT_MIN_CODE_VERSION_KEY = "minCodeVersion";
+    private static final boolean DEFAULT_STRICT_VALIDATION = false;
+    private static final VersionComparator DEFAULT_VERSION_COMPARATOR = new SemVerComparator();
+
+    private final String currentVersion;
```
chrfwow, line 21: Since this version cannot change, we should store the parsed version instead, and do comparisons against this stored version. This will save us some resources, becuase we don't need to parse this immutable version on every hook execution

hooks/codereadiness/src/main/java/dev/openfeature/contrib/hooks/codereadiness/CodeReadinessHook.java
marcin11858: You are right, I've removed this additional check

hooks/codereadiness/src/main/java/dev/openfeature/contrib/hooks/codereadiness/CodeReadinessHook.java
marcin11858: I changed it to the second option. Thanks for the tip!

hooks/codereadiness/src/main/java/dev/openfeature/contrib/hooks/codereadiness/SemVerComparator.java
marcin11858: Done, added check, thanks!

hooks/codereadiness/src/main/java/dev/openfeature/contrib/hooks/codereadiness/CodeReadinessHook.java
marcin11858: SemVerComparator was modified to store currentVersion  value

hooks/codereadiness/src/main/java/dev/openfeature/contrib/hooks/codereadiness/CodeReadinessHook.java
```diff
@@ -0,0 +1,89 @@
...
+    private final boolean strictValidation;
+    private final String metadataMinVerKey;
+    private final VersionComparator comparator;
+
+    public static class Builder {
+        private String currentVersion;
+        private boolean strictValidation = DEFAULT_STRICT_VALIDATION;
+        private String metadataMinVerKey = DEFAULT_MIN_CODE_VERSION_KEY;
+        private Function<String, VersionComparator> comparator = SemVerComparator::new;
```
chrfwow: I see why it was done, but it feels a bit of a weird interface that one has to supply a lambda to the builder

marcin11858: Done, changed to a generic interface

hooks/codereadiness/src/main/java/dev/openfeature/contrib/hooks/codereadiness/VersionComparator.java
```diff
@@ -0,0 +1,33 @@
...
+
+    /**
+     * Compare current version with required version.
+     *
+     * @param currentVersion current parsed version
+     * @param minCodeVersion required minimum parsed version
+     * @return true if currentVersion is greater than or equal to minCodeVersion
+     * @throws Exception if there is a comparison error
+     */
+    boolean compare(T currentVersion, T minCodeVersion) throws Exception;
```
chrfwow, line 23: To make this interface even more generic, also for other use cases, this method should probably be called `isGreaterThanOrEqualTo`.
I also wonder if we should add a default method to this interface:
```
default boolean isGreaterThanOrEqualTo(String a, String b) {
   return isGreaterThanOrEqualTo(parse(a), parse(b));
}
```

chrfwow (commented):
You will need to run the spotlessApply maven target to fix the formating issues

hooks/codereadiness/src/main/java/dev/openfeature/contrib/hooks/codereadiness/VersionComparator.java
marcin11858: In my option `compare` is more general name and doesn't suggest the implementation. Some software engineer might decide to say that only with higher or quals version numbers are valid to enable new feature. 

When we will add a default method we will add parsing `currentVersion` value during every execution the method.

chrfwow: Hmm... I see your point.

chrfwow (approved):
I'll approve this because the code looks good. I have not followed the conversation about this hook in the spec repo. Please make sure you align with that outcome

hooks/codereadiness/src/main/java/dev/openfeature/contrib/hooks/codereadiness/VersionComparator.java
```diff
@@ -0,0 +1,33 @@
...
+ * @param <T> The domain object type representing a parsed version (e.g., Semver, LocalDate, Integer).
+ */
+public interface VersionComparator<T> {
+
+    /**
+     * Parse version string into domain object.
+     *
+     * @param versionString version string to parse
+     * @return parsed version object
+     * @throws Exception if there is a parsing error
```
oxddr: nit: Unless this is enforced by some checkers, I'd remove @clauses which don't provide any extra details over what the bare javadoc provides (here and in other places)

oxddr (approved):

hooks/codereadiness/src/main/java/dev/openfeature/contrib/hooks/codereadiness/VersionComparator.java
marcin11858: Done, removed @clauses

toddbaert (approved):
Thanks @marcin11858 . I made some minor changes (adding the flag key to the warning message) but LGTM. I will merge and release. Sorry about the delay - We'll be faster on other implementations; feel free to ping me on them directly.

PR #1831, opened 2026-08-07
feat: Update fractional logic to support hashing consistency ADR.

### LINKED ISSUE #1662: [flagd] [FEATURE] Harden Hashing Consistency And Add Support For Non-string Attributes in Fractional Evaluation
Implement [this ADR](https://github.com/open-feature/flagd/blob/main/docs/architecture-decisions/fractional-non-string-rand-units.md) for the flagd provider.

NeaguGeorgiana23:
## This PR
- Updates the fractional targeting evaluation logic in `flagd-core` to adhere to the hashing consistency ADR.
- Adds `com.upokecenter:cbor` dependency (v4.5.6) to serialize bucketing keys (flag key, targeting key, primitives, maps, and lists) to canonical CBOR format before MurmurHash3 (32-bit x86) hashing.
- Implements canonical CBOR sorting (`KEY_COMPARATOR`) for map keys based on byte length and lexicographical byte order.
- Updates `test-harness` submodules in both `providers/flagd` and `tools/flagd-api-testkit` to commit `82ba89e`.
- Enhances E2E test step definitions (`ContextSteps`, `ConfigSteps`, `EvaluationSteps`, `Utils`) with DataTable context key support, improved type conversions (handling `null` strings, fallback from Integer to Long), and error reason handling.
- Updates `FlagdProviderSyncResources` to properly reset `isInitialized` on fatal errors and shutdowns.
- Updates test expectations, exclusion tags (`fractional-v1` -> `fractional-v2`), and test fixture JSONs (`selfContainedFractional0.json`, `string.json`) to reflect the new hashing results.

### Related Issues
Fixes #1662 

### Notes
- Fixture test values were updated to match the deterministic hashes produced by canonical CBOR + MurmurHash3.

### Follow-up Tasks
- For all new Gherkin tests to run properly, Java needs to add Long support, as discussed [on slack](https://cloud-native.slack.com/archives/C03J36ZP020/p1779268348946649).

### DIFF
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProviderSyncResources.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProviderSyncResources.java
@@ -18,7 +18,6 @@ class FlagdProviderSyncResources {
     @Setter
     private volatile ProviderEvent previousEvent;
 
-    @Setter
     private volatile boolean isFatal;
 
     private volatile ProviderEventDetails fatalProviderEventDetails;
@@ -111,12 +110,22 @@ class FlagdProviderSyncResources {
      */
     public synchronized void shutdown() {
         isShutDown = true;
+        isInitialized = false;
         this.notifyAll();
     }
 
     public synchronized void fatalError(ProviderEventDetails providerEventDetails) {
         isFatal = true;
+        isInitialized = false;
         fatalProviderEventDetails = providerEventDetails;
+        this.notifyAll();
+    }
+
+    public synchronized void setFatal(boolean fatal) {
+        isFatal = fatal;
+        if (fatal) {
+            isInitialized = false;
+        }
         this.notifyAll();
     }
 }
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProviderSyncResources.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProviderSyncResources.java
@@ -110,22 +110,12 @@ class FlagdProviderSyncResources {
      */
     public synchronized void shutdown() {
         isShutDown = true;
-        isInitialized = false;
         this.notifyAll();
     }
 
     public synchronized void fatalError(ProviderEventDetails providerEventDetails) {
         isFatal = true;
-        isInitialized = false;
         fatalProviderEventDetails = providerEventDetails;
-        this.notifyAll();
-    }
-
-    public synchronized void setFatal(boolean fatal) {
-        isFatal = fatal;
-        if (fatal) {
-            isInitialized = false;
-        }
         this.notifyAll();
     }
 }
--- a/providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/steps/ContextSteps.java
+++ b/providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/steps/ContextSteps.java
@@ -4,8 +4,11 @@ import dev.openfeature.contrib.providers.flagd.e2e.State;
 import dev.openfeature.sdk.ImmutableStructure;
 import dev.openfeature.sdk.MutableContext;
 import dev.openfeature.sdk.Value;
+import io.cucumber.datatable.DataTable;
 import io.cucumber.java.en.Given;
+import java.io.IOException;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 
 public class ContextSteps extends AbstractSteps {
@@ -14,32 +17,24 @@ public class ContextSteps extends AbstractSteps {
         super(state);
     }
 
-    @Given("a context containing a key {string}, with type {string} and with value {string}")
+    @Given("^a context containing a key \"([^\"]*)\", with type \"([^\"]*)\" and with value \"(.*)\"$")
     public void a_context_containing_a_key_with_type_and_with_value(String key, String type, String value)
-            throws ClassNotFoundException, InstantiationException {
-        Map<String, Value> map = state.context.asMap();
-        Value typedValue;
-        switch (type) {
-            case "Integer":
-                long longVal = Long.parseLong(value);
-                if (longVal >= Integer.MIN_VALUE && longVal <= Integer.MAX_VALUE) {
-                    typedValue = new Value((int) longVal);
-                } else {
-                    // value exceeds int range; store as string to preserve precision
-                    typedValue = new Value(value);
-                }
-                break;
-            case "Float":
-                typedValue = new Value(Double.parseDouble(value));
-                break;
-            case "Boolean":
-                typedValue = new Value(Boolean.parseBoolean(value));
-                break;
-            default:
-                typedValue = new Value(value);
-                break;
+            throws ClassNotFoundException, IOException {
+        Map<String, Value> map = new HashMap<>(state.context.asMap());
+        map.put(key, Value.objectToValue(Utils.convert(value, type)));
+        state.context = new MutableContext(state.context.getTargetingKey(), map);
+    }
+
+    @Given("a context with the following keys:")
+    public void a_context_with_the_following_keys(DataTable dataTable) throws ClassNotFoundException, IOException {
+        List<Map<String, String>> rows = dataTable.asMaps(String.class, String.class);
+        Map<String, Value> map = new HashMap<>(state.context.asMap());
+        for (Map<String, String> row : rows) {
+            String key = row.get("key");
+            String type = row.get("type");
+            String value = row.get("value");
+            map.put(key, Value.objectToValue(Utils.convert(value, type)));
         }
-        map.put(key, typedValue);
         state.context = new MutableContext(state.context.getTargetingKey(), map);
     }
 
--- a/tools/flagd-api-testkit/src/main/java/dev/openfeature/contrib/tools/flagd/api/testkit/ContextSteps.java
+++ b/tools/flagd-api-testkit/src/main/java/dev/openfeature/contrib/tools/flagd/api/testkit/ContextSteps.java
@@ -4,9 +4,11 @@ import dev.openfeature.sdk.ImmutableStructure;
 import dev.openfeature.sdk.MutableContext;
 import dev.openfeature.sdk.Value;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import io.cucumber.datatable.DataTable;
 import io.cucumber.java.en.Given;
 import java.io.IOException;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 
 /**
@@ -25,10 +27,24 @@ public class ContextSteps {
     }
 
     /** Adds a typed key/value pair to the evaluation context. */
-    @Given("a context containing a key {string}, with type {string} and with value {string}")
+    @Given("^a context containing a key \"([^\"]*)\", with type \"([^\"]*)\" and with value \"(.*)\"$")
     public void contextKeyWithTypeAndValue(String key, String type, String value) throws IOException {
         Map<String, Value> map = new HashMap<>(state.context.asMap());
         map.put(key, Value.objectToValue(EvaluatorUtils.convert(value, type)));
+        state.context = new MutableContext(state.context.getTargetingKey(), map);
+    }
+
+    /** Adds multiple context keys from a data table. */
+    @Given("a context with the following keys:")
+    public void contextWithFollowingKeys(DataTable dataTable) throws IOException {
+        List<Map<String, String>> rows = dataTable.asMaps(String.class, String.class);
+        Map<String, Value> map = new HashMap<>(state.context.asMap());
+        for (Map<String, String> row : rows) {
+            String key = row.get("key");
+            String type = row.get("type");
+            String value = row.get("value");
+            map.put(key, Value.objectToValue(EvaluatorUtils.convert(value, type)));
+        }
         state.context = new MutableContext(state.context.getTargetingKey(), map);
     }
 
--- a/tools/flagd-api-testkit/src/main/java/dev/openfeature/contrib/tools/flagd/api/testkit/ContextSteps.java
+++ b/tools/flagd-api-testkit/src/main/java/dev/openfeature/contrib/tools/flagd/api/testkit/ContextSteps.java
@@ -46,7 +46,7 @@ public class ContextSteps {
             map.put(key, Value.objectToValue(EvaluatorUtils.convert(value, type)));
         }
         state.context = new MutableContext(state.context.getTargetingKey(), map);
-    }
+        }
 
     /** Sets the targeting key on the evaluation context. */
     @Given("a context containing a targeting key with value {string}")
--- a/tools/flagd-api-testkit/src/main/java/dev/openfeature/contrib/tools/flagd/api/testkit/ContextSteps.java
+++ b/tools/flagd-api-testkit/src/main/java/dev/openfeature/contrib/tools/flagd/api/testkit/ContextSteps.java
@@ -46,7 +46,7 @@ public class ContextSteps {
             map.put(key, Value.objectToValue(EvaluatorUtils.convert(value, type)));
         }
         state.context = new MutableContext(state.context.getTargetingKey(), map);
-        }
+    }
 
     /** Sets the targeting key on the evaluation context. */
     @Given("a context containing a targeting key with value {string}")
--- a/tools/flagd-core/src/main/java/dev/openfeature/contrib/tools/flagd/core/targeting/Fractional.java
+++ b/tools/flagd-core/src/main/java/dev/openfeature/contrib/tools/flagd/core/targeting/Fractional.java
@@ -20,6 +20,7 @@ import org.apache.commons.codec.digest.MurmurHash3;
 @Slf4j
 class Fractional implements PreEvaluatedArgumentsExpression {
 
+    private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
     static final int MAX_WEIGHT = Integer.MAX_VALUE;
 
     @Override
@@ -59,9 +60,6 @@ class Fractional implements PreEvaluatedArgumentsExpression {
             // fallback to targeting key if present
             if (properties.getTargetingKey() == null) {
                 log.debug("Missing fallback targeting key");
-                // if (arguments.size() == 2) {
-                //     throw new dev.openfeature.sdk.exceptions.GeneralError("Missing fallback targeting key");
-                // }
                 return null;
             }
 
@@ -85,7 +83,7 @@ class Fractional implements PreEvaluatedArgumentsExpression {
                 totalWeight += fractionProperty.getWeight();
             } catch (JsonLogicException e) {
                 if ("Property is not an array".equals(e.getMessage())) {
-                    throw new io.github.jamsesso.jsonlogic.evaluator.JsonLogicEvaluationException(
+                    throw new JsonLogicEvaluationException(
                             "Error parsing fractional targeting rule: " + e.getMessage(), jsonPath);
                 }
                 return null;
@@ -105,8 +103,6 @@ class Fractional implements PreEvaluatedArgumentsExpression {
         // find distribution
         return distributeValue(bucketBy, propertyList, (int) totalWeight, jsonPath);
     }
-
-    private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
 
     private static Object distributeValue(
             final Object hashKey,
--- a/tools/flagd-core/src/main/java/dev/openfeature/contrib/tools/flagd/core/targeting/Fractional.java
+++ b/tools/flagd-core/src/main/java/dev/openfeature/contrib/tools/flagd/core/targeting/Fractional.java
@@ -59,9 +59,9 @@ class Fractional implements PreEvaluatedArgumentsExpression {
             // fallback to targeting key if present
             if (properties.getTargetingKey() == null) {
                 log.debug("Missing fallback targeting key");
-                if (arguments.size() == 2) {
-                    throw new dev.openfeature.sdk.exceptions.GeneralError("Missing fallback targeting key");
-                }
+                // if (arguments.size() == 2) {
+                //     throw new dev.openfeature.sdk.exceptions.GeneralError("Missing fallback targeting key");
+                // }
                 return null;
             }
 
--- a/tools/flagd-core/src/main/java/dev/openfeature/contrib/tools/flagd/core/targeting/Fractional.java
+++ b/tools/flagd-core/src/main/java/dev/openfeature/contrib/tools/flagd/core/targeting/Fractional.java
@@ -1,10 +1,16 @@
 package dev.openfeature.contrib.tools.flagd.core.targeting;
 
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.upokecenter.cbor.CBORObject;
 import io.github.jamsesso.jsonlogic.JsonLogicException;
 import io.github.jamsesso.jsonlogic.evaluator.JsonLogicEvaluationException;
 import io.github.jamsesso.jsonlogic.evaluator.expressions.PreEvaluatedArgumentsExpression;
+import java.io.Serializable;
 import java.nio.charset.StandardCharsets;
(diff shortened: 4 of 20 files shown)


toddbaert:
@NeaguGeorgiana23 - Long support was released last week in Java SDK `v1.22.0` - You should be able to finish this now I think.

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProviderSyncResources.java
```diff
@@ -111,12 +110,22 @@ public void waitForInitialization(long deadline) {
      */
     public synchronized void shutdown() {
         isShutDown = true;
+        isInitialized = false;
```
chrfwow: Why do we set `isInitialized` to false? I think this would make it possible to re-init the provider afterwards, which, if I remember correctly, could lead to a race condition. Same for `fatalError(...)`

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProviderSyncResources.java
```diff
@@ -111,12 +110,22 @@ public void waitForInitialization(long deadline) {
...
     }
 
     public synchronized void fatalError(ProviderEventDetails providerEventDetails) {
         isFatal = true;
+        isInitialized = false;
         fatalProviderEventDetails = providerEventDetails;
         this.notifyAll();
     }
+
+    public synchronized void setFatal(boolean fatal) {
```
chrfwow: Why do we need this method? One could use `fatalError` instead

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/steps/ContextSteps.java
```diff
@@ -14,32 +17,24 @@ public ContextSteps(State state) {
...
-                typedValue = new Value(Double.parseDouble(value));
-                break;
-            case "Boolean":
-                typedValue = new Value(Boolean.parseBoolean(value));
-                break;
-            default:
-                typedValue = new Value(value);
-                break;
+            throws ClassNotFoundException, IOException {
+        Map<String, Value> map = new HashMap<>(state.context.asMap());
```
chrfwow, line 23: We didn't create a new Map wrapping the result of state.context.asMap before, why do we need to do it now?

tools/flagd-api-testkit/src/main/java/dev/openfeature/contrib/tools/flagd/api/testkit/ContextSteps.java
```diff
@@ -25,13 +27,27 @@ public ContextSteps(EvaluatorState state) {
...
-    @Given("a context containing a key {string}, with type {string} and with value {string}")
+    @Given("^a context containing a key \"([^\"]*)\", with type \"([^\"]*)\" and with value \"(.*)\"$")
     public void contextKeyWithTypeAndValue(String key, String type, String value) throws IOException {
         Map<String, Value> map = new HashMap<>(state.context.asMap());
         map.put(key, Value.objectToValue(EvaluatorUtils.convert(value, type)));
         state.context = new MutableContext(state.context.getTargetingKey(), map);
     }
 
+    /** Adds multiple context keys from a data table. */
+    @Given("a context with the following keys:")
```
chrfwow, line 38: Duplicate of providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/steps/ContextSteps.java ?

tools/flagd-core/src/main/java/dev/openfeature/contrib/tools/flagd/core/targeting/Fractional.java
```diff
@@ -42,34 +46,50 @@ public Object evaluate(List arguments, Object data, String jsonPath) throws Json
...
+                || arguments.get(0) instanceof Number
+                || arguments.get(0) instanceof java.util.Map) {
+            // first arg is a primitive or Map, use for bucketing
+            bucketBy = arguments.get(0);
             distributions = arguments.subList(1, arguments.size());
         } else {
             // fallback to targeting key if present
             if (properties.getTargetingKey() == null) {
                 log.debug("Missing fallback targeting key");
+                // if (arguments.size() == 2) {
```
chrfwow: Do we need this commented out code?

tools/flagd-core/src/main/java/dev/openfeature/contrib/tools/flagd/core/targeting/Fractional.java
```diff
@@ -42,34 +46,50 @@ public Object evaluate(List arguments, Object data, String jsonPath) throws Json
...
-        } else if (arguments.get(0) instanceof String) {
-            // first arg is a String, use for bucketing
-            bucketBy = (String) arguments.get(0);
+        } else if (arguments.get(0) instanceof String
+                || arguments.get(0) instanceof Boolean
+                || arguments.get(0) instanceof Number
+                || arguments.get(0) instanceof java.util.Map) {
+            // first arg is a primitive or Map, use for bucketing
+            bucketBy = arguments.get(0);
             distributions = arguments.subList(1, arguments.size());
```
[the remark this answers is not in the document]
chrfwow: Do we want to support list bucket keys?

tools/flagd-core/src/main/java/dev/openfeature/contrib/tools/flagd/core/targeting/Fractional.java
```diff
@@ -42,34 +46,50 @@ public Object evaluate(List arguments, Object data, String jsonPath) throws Json
...
-        try {
-            for (Object dist : distributions) {
+        for (Object dist : distributions) {
+            try {
                 FractionProperty fractionProperty = new FractionProperty(dist, jsonPath);
                 propertyList.add(fractionProperty);
                 totalWeight += fractionProperty.getWeight();
+            } catch (JsonLogicException e) {
+                if ("Property is not an array".equals(e.getMessage())) {
+                    throw new io.github.jamsesso.jsonlogic.evaluator.JsonLogicEvaluationException(
```
chrfwow: Can we do a proper import?

tools/flagd-core/src/main/java/dev/openfeature/contrib/tools/flagd/core/targeting/Fractional.java
```diff
@@ -86,13 +106,23 @@ public Object evaluate(List arguments, Object data, String jsonPath) throws Json
         return distributeValue(bucketBy, propertyList, (int) totalWeight, jsonPath);
     }
 
+    private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
```
chrfwow: Should be at the top of the file, together with other fields

tools/flagd-api-testkit/src/main/java/dev/openfeature/contrib/tools/flagd/api/testkit/ContextSteps.java
NeaguGeorgiana23: They have matching Cucumber step annotations because both suites run the evaluator feature scenarios from test-harness, but they serve two distinct purposes across different modules:

1. providers/flagd/.../ContextSteps.java (under src/test) is for end-to-end testing of FlagdProvider and depends on the provider e2e State (managing clients, resolvers, and test containers).

2. tools/flagd-api-testkit/.../ContextSteps.java (under src/main) is part of the published, reusable flagd-api-testkit library, which depends on EvaluatorState and allows any standalone Evaluator SPI implementation (like flagd-core) to verify compliance against the test suite.
Because flagd-api-testkit is a standalone reusable testkit distributed in src/main, it maintains its own step definitions.

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/steps/ContextSteps.java
NeaguGeorgiana23: AFAIK `state.context.asMap()` returns an unmodifiable map (Collections.unmodifiableMap(...)), so calling `.put()` directly on it throws an UnsupportedOperationException. Also, `MutableContext` does not provide an `add(String, Value)` overload that accepts a generic Value instance. Wrapping the existing attributes in new `HashMap<>(state.context.asMap())` and recreating the `MutableContext` is necessary to safely append the converted Value attributes.

toddbaert:
@NeaguGeorgiana23 the in-process failures on the timestamp scenario trace back to a gap in the SDK's `Long` support added in open-feature/java-sdk#1985; `Structure.convertValue` still only handles `Integer`/`Double`, so a `Long` context value throws `ValueNotConvertableError` and falls back to the default; not a fractional/CBOR issue, and we'll open a fix upstream. Tracked in open-feature/java-sdk#2019.

toddbaert:
> @NeaguGeorgiana23 the in-process failures on the timestamp scenario trace back to a gap in the SDK's `Long` support added in [open-feature/java-sdk#1985](https://github.com/open-feature/java-sdk/pull/1985); `Structure.convertValue` still only handles `Integer`/`Double`, so a `Long` context value throws `ValueNotConvertableError` and falls back to the default; not a fractional/CBOR issue, and we'll open a fix upstream. Tracked in [open-feature/java-sdk#2019](https://github.com/open-feature/java-sdk/issues/2019).

Opened: https://github.com/open-feature/java-sdk/pull/2020. I will rebase on that after merge.

toddbaert:
Hey @NeaguGeorgiana23; I pushed a change to correct the fractional test tags (v3 for in-process, observing your new hash assertions) and v2 for RPC (since the published flagd still doesn't have those new behaviors).

Overall this implementation looks good to me except perhaps [this](https://github.com/open-feature/java-sdk-contrib/pull/1831#issuecomment-5441589248), and I think if we implement the rest of the providers similarly we can merge them all and release in a tight timeframe. The challenge will be making sure everything stays consistent across languages and modes - we need consistency here or users get bucketed differently in different apps for the same flag. I think we have 2 choices:

1. Add a switch/option for the new fractional behavior everywhere, defaulting to `v2` for now, and switching the default to `v3` in a later release; or
2. Keep this PR and all the others just pending, and merge/release all of its siblings and the flagd-core CBOR change at the same time.

I'm fine with either; it's really just a question of what's better for you in terms of maintenance. What do you think?

cc @cupofcat

toddbaert:
@NeaguGeorgiana23 one concern on the CBOR number handling: I don't think numeric bucketing is provably consistent cross-language yet. Floats always encode as f64 (the ADR wants shortest-form), and whole floats like `1.0` collapse to a CBOR int (type 0 vs type 7) which may not match Go ([open-feature/flagd#2011](https://github.com/open-feature/flagd/pull/2011)). And none of the `@fractional-v3` testbed scenarios feed a number/map as the bucket input, so green CI doesn't actually prove Java and Go agree here.

Not necessarily a blocker for this PR, but before we release these together I'd want the number-normalization rules pinned to the Go reference impl plus shared testbed vectors for float/int/map inputs. WDYT?

NeaguGeorgiana23:
> Hey @NeaguGeorgiana23; I pushed a change to correct the fractional test tags (v3 for in-process, observing your new hash assertions) and v2 for RPC (since the published flagd still doesn't have those new behaviors).
> 
> Overall this implementation looks good to me except perhaps [this](https://github.com/open-feature/java-sdk-contrib/pull/1831#issuecomment-5441589248), and I think if we implement the rest of the providers similarly we can merge them all and release in a tight timeframe. The challenge will be making sure everything stays consistent across languages and modes - we need consistency here or users get bucketed differently in different apps for the same flag. I think we have 2 choices:
> 
> 1. Add a switch/option for the new fractional behavior everywhere, defaulting to `v2` for now, and switching the default to `v3` in a later release; or
> 2. Keep this PR and all the others just pending, and merge/release all of its siblings and the flagd-core CBOR change at the same time.
> 
> I'm fine with either; it's really just a question of what's better for you in terms of maintenance. What do you think?
> 
> cc @cupofcat

I think that the best option here depends on which languages you are thinking of when you say "release all of its siblings". So far, we have the implementations for Go and Java and we have the ones for C++ and Python about 50% implemented.

If these 4 languages would be enough, maybe the second option is best. But if you are thinking to release this for all languages at once, IMO maybe the first option is best. 

WDYT @cupofcat?

NeaguGeorgiana23:
> @NeaguGeorgiana23 one concern on the CBOR number handling: I don't think numeric bucketing is provably consistent cross-language yet. Floats always encode as f64 (the ADR wants shortest-form), and whole floats like `1.0` collapse to a CBOR int (type 0 vs type 7) which may not match Go ([open-feature/flagd#2011](https://github.com/open-feature/flagd/pull/2011)). And none of the `@fractional-v3` testbed scenarios feed a number/map as the bucket input, so green CI doesn't actually prove Java and Go agree here.
> 
> Not necessarily a blocker for this PR, but before we release these together I'd want the number-normalization rules pinned to the Go reference impl plus shared testbed vectors for float/int/map inputs. WDYT?

I am not sure I understand exactly what you are thinking of. But I tried to double check the new Gherkin scenarios from `flagd-testbed` and the implementations for `fractiona.

You mentioned that floats always encode as f64, but I don't think that is the case. 
Go uses `cbor.CoreDetEncOptions()` which uses RFC 8949 Core Deterministic Encoding. AFAIK, this enforces shortest-float representation. 
Java uses `com.upokecenter.cbor` which AFAIK also defaults to shortest-float encoding, compressing to f16 or f32 whenever lossless. 
And, IMO, the fact that both languages act the same is proven by the fact that they pass the scenario `Fractional operator with floats limits`. This scenario is testing precision boundaries:
- **65504.0** (max f16 boundary) vs **65505.0** (exceeds f16 --> uses f32)
- **3.4028234663852886e+38** (max  f32 boundary) vs **3.5e+38** (exceeds f32 --> uses f64).
Both Go and Java assign the exact same buckets across all these float boundaries.


Also, the concern about whole floats like 1.0 collapsing to a CBOR int (type 0 vs type 7) I think is already handled since both Go and Java have functions for `normalization`:
- In Go:
```
if v == math.Trunc(v) {
    if v >= 0 && v <= float64(math.MaxUint64) {
        return uint64(v) // -> Encodes as Major Type 0 (Unsigned Int)
    }
    if v < 0 && v >= float64(math.MinInt64) {
        return int64(v)  // -> Encodes as Major Type 1 (Negative Int)
    }
}
```

- in Java:
```
double val = node.asDouble();
if (val == Math.floor(val) && val >= Long.MIN_VALUE && val <= Long.MAX_VALUE) {
    return CBORObject.FromObject((long) val); // -> Encodes as Major Type 0 / 1
}
```

This behavior is tested in the `Fractional operator with float mapping` scenario:
```
  | hashing_input                | type      | value       |
  | 1.0                                   | Float     | bucket23 |
  | 1                                      | Integer | bucket23  |
  | -2.0                                 | Float    | bucket12   |
  | -2                                    | Integer | bucket12   |
  | 9007199254740992.0 | Float     | bucket10   |
  | 9007199254740992    | Integer | bucket10   |
```

LMK if this is what you were thinking of or if there are other concerns that need to be discussed, or if you'd like to jump on a quick call to help me understand what I'm overlooking regarding the consistency across languages.

m-olko:
Hey, @toddbaert following on [Georgiana's comment](https://github.com/open-feature/java-sdk-contrib/pull/1831#issuecomment-5477033308), the implementations for Python and C++ are in review:
https://github.com/open-feature/python-sdk-contrib/pull/415
https://github.com/open-feature/cpp-sdk-contrib/pull/131
So we can align on merging them as well

tools/flagd-core/src/main/java/dev/openfeature/contrib/tools/flagd/core/targeting/Fractional.java
```diff
@@ -42,34 +47,47 @@ public Object evaluate(List arguments, Object data, String jsonPath) throws Json
                 log.debug("Missing fallback targeting key");
                 return null;
             }
-            bucketBy = properties.getFlagKey() + properties.getTargetingKey();
+            bucketBy = java.util.Arrays.asList(properties.getFlagKey(), properties.getTargetingKey());
```
toddbaert, line 50: nit: add an import instead of the FQCN

tools/flagd-core/src/main/java/dev/openfeature/contrib/tools/flagd/core/targeting/Fractional.java
```diff
@@ -42,34 +47,47 @@ public Object evaluate(List arguments, Object data, String jsonPath) throws Json
...
             distributions = List.of(arguments);
-        } else if (arguments.get(0) instanceof String) {
-            // first arg is a String, use for bucketing
-            bucketBy = (String) arguments.get(0);
+        } else if (arguments.get(0) instanceof String
+                || arguments.get(0) instanceof Boolean
+                || arguments.get(0) instanceof Number
+                || arguments.get(0) instanceof java.util.Map) {
+            // first arg is a primitive or Map, use for bucketing
+            bucketBy = arguments.get(0);
```
toddbaert, line 57: I didn't test this manually, so I'm not 100% sure, but I think that if you have a rule like this:

```json
{"fractional": [{"var": "someAttr"}, ["one", 50], ["two", 50]]}
```

And `someAttr` is null or unset, we revert to the default `$flag.flagKey + targetingKey` behavior, instead of defaulting. I don't think the new python impl does the same, and I think this is a change from fractional v2 not fully covered in the test suite.

All that said, it's a small change so we can address it separately with an improvement to the testbed, or just fix it here and do the testbed improvement later if you agree.

toddbaert:
> > @NeaguGeorgiana23 one concern on the CBOR number handling: I don't think numeric bucketing is provably consistent cross-language yet. Floats always encode as f64 (the ADR wants shortest-form), and whole floats like `1.0` collapse to a CBOR int (type 0 vs type 7) which may not match Go ([open-feature/flagd#2011](https://github.com/open-feature/flagd/pull/2011)). And none of the `@fractional-v3` testbed scenarios feed a number/map as the bucket input, so green CI doesn't actually prove Java and Go agree here.
> > Not necessarily a blocker for this PR, but before we release these together I'd want the number-normalization rules pinned to the Go reference impl plus shared testbed vectors for float/int/map inputs. WDYT?
> 
> I am not sure I understand exactly what you are thinking of. But I tried to double check the new Gherkin scenarios from `flagd-testbed` and the implementations for `fractiona.
> 
> You mentioned that floats always encode as f64, but I don't think that is the case. Go uses `cbor.CoreDetEncOptions()` which uses RFC 8949 Core Deterministic Encoding. AFAIK, this enforces shortest-float representation. Java uses `com.upokecenter.cbor` which AFAIK also defaults to shortest-float encoding, compressing to f16 or f32 whenever lossless. And, IMO, the fact that both languages act the same is proven by the fact that they pass the scenario `Fractional operator with floats limits`. This scenario is testing precision boundaries:
> 
>     * **65504.0** (max f16 boundary) vs **65505.0** (exceeds f16 --> uses f32)
> 
>     * **3.4028234663852886e+38** (max  f32 boundary) vs **3.5e+38** (exceeds f32 --> uses f64).
>       Both Go and Java assign the exact same buckets across all these float boundaries.
> 
> 
> Also, the concern about whole floats like 1.0 collapsing to a CBOR int (type 0 vs type 7) I think is already handled since both Go and Java have functions for `normalization`:
> 
>     * In Go:
> 
> 
> ```
> if v == math.Trunc(v) {
>     if v >= 0 && v <= float64(math.MaxUint64) {
>         return uint64(v) // -> Encodes as Major Type 0 (Unsigned Int)
>     }
>     if v < 0 && v >= float64(math.MinInt64) {
>         return int64(v)  // -> Encodes as Major Type 1 (Negative Int)
>     }
> }
> ```
> 
>     * in Java:
> 
> 
> ```
> double val = node.asDouble();
> if (val == Math.floor(val) && val >= Long.MIN_VALUE && val <= Long.MAX_VALUE) {
>     return CBORObject.FromObject((long) val); // -> Encodes as Major Type 0 / 1
> }
> ```
> 
> This behavior is tested in the `Fractional operator with float mapping` scenario:
> 
> ```
>   | hashing_input                | type      | value       |
>   | 1.0                                   | Float     | bucket23 |
>   | 1                                      | Integer | bucket23  |
>   | -2.0                                 | Float    | bucket12   |
>   | -2                                    | Integer | bucket12   |
>   | 9007199254740992.0 | Float     | bucket10   |
>   | 9007199254740992    | Integer | bucket10   |
> ```
> 
> LMK if this is what you were thinking of or if there are other concerns that need to be discussed, or if you'd like to jump on a quick call to help me understand what I'm overlooking regarding the consistency across languages.

@NeaguGeorgiana23 

You're right, sorry about that! I did some manual testing before, and my hand-written test had an error itself! I see the testbed also does explicitly test this [here](https://github.com/open-feature/flagd-testbed/blob/b308c1bbb12fc4511af6220f1b072a20a1a8831f/gherkin/targeting.feature#L544-L561), so the implementation seems good in the numeric handling :pray: . I have no objections to this implementation now. My _only_ question is the [releasing one](https://github.com/open-feature/java-sdk-contrib/pull/1831#issuecomment-5441523976). @m-olko [opened Python](https://github.com/open-feature/python-sdk-contrib/pull/415) and C++, I will have some time next week, so I'm happy to help with either JS or .NET... 

If we get those done, then I'll coordinate a release across all the languages, and I think we can skip the "v3 opt in option" dance.... but I'm still ok going either way.

I may have found one other issue when comparing this to the Python impl. Please give it a quick check: https://github.com/open-feature/java-sdk-contrib/pull/1831/changes#r3982812450

toddbaert (approved):
One possible bug noted [here](https://github.com/open-feature/java-sdk-contrib/pull/1831#issuecomment-5624411772), and we have to decide on a release strategy, but this looks good.

Approving but I'd note that if we merge, we should plan to finish other implementations relatively quickly so it might be better to leave this open until we have others ready to go.

NeaguGeorgiana23:
I will look over the issues that you found and address them. 

Also, IMO the easiest option is to release the update in all languages at once since we already have 4 languages that implement this logic.

toddbaert:
> I will look over the issues that you found and address them.
> 
> Also, IMO the easiest option is to release the update in all languages at once since we already have 4 languages that implement this logic.

Agree.

I'm starting JS now.

:warning: I found a problem though! The [CBOR dep we chose](https://github.com/open-feature/java-sdk-contrib/pull/1831/changes#diff-5896de910970bec3a4a21ba8ee74e048e2037fc6a398f6e0e8de7ef5c60b6665R43-R48) is not on the [CNCF licence allow-list](https://github.com/cncf/foundation/blob/main/policies-guidance/allowed-third-party-license-policy.md#approved-licenses-for-allowlist).

Other options:

- the Jackson version can't emit float16 (breaks byte-parity).
- [cbor-java](https://mvnrepository.com/artifact/co.nstant.in/cbor) can (with a custom helper, I think) but its Maven artifact has been stale since 2020... 
- build our own small encoder based on the spec, for our narrow use-case (UPDATE: I did a PoC of this, see below). I can also try to get a CNCF exception for the dep we are using now, but I think that might be troublesome. I think [convertNode](https://github.com/open-feature/java-sdk-contrib/pull/1831/changes#diff-b9ae97777b4cd9d91ddae97cef4a19e798498c50d062b278aa97d8e4cfa469cdR174) does a lot of the work (sorting) already; it's just adding some byte conversion and float handling there... I don't think conforming to the spec would be a huge amount of effort. 

Let me know what you think - I guess another option is to change the ADR to behavior so that the Jackson version (most popular) can satisfy it, but I think that's a bad idea as well (it's not best for our use case and it means our test best assertions will all change).

cc @m-olko @cupofcat @oxddr 

--- 

:exclamation: EDIT: I made a tiny PoC to your branch here: https://github.com/NeaguGeorgiana23/java-sdk-contrib/pull/1 - it seems to work, at least it passes all the e2e tests. 

I'd rather not maintain this code, but I don't see a clean alternative here. WDYT?

tools/flagd-core/src/main/java/dev/openfeature/contrib/tools/flagd/core/targeting/Fractional.java
NeaguGeorgiana23: I looked closer over the Java implementation and compared it to the Go and Python implementation and you are right, it is different. And, while checking the tests, I noticed why this case was not caught by them. The test case that is checking this `missing bucketingKey` scenario does not define a `targetingKey`. Because of this, both the bucketting property and the targeting key are null, thus missing this bug.

PR #1835, opened 2026-08-19
fix: correct grace-period/max-retry-backoff defaults to prevent log spam

### LINKED ISSUE #1833: flagd: align retry backoff / grace period defaults with spec, fix retry `maxAttempts`
We must implement open-feature/flagd#2027, which proposes changing the spec defaults so the stream-reconnect backoff can't leave us disconnected longer than the stale grace period.

- `Config.java:20` `DEFAULT_MAX_RETRY_BACKOFF_MS`: 12000 -> 5000
- `Config.java:22` `DEFAULT_STREAM_RETRY_GRACE_PERIOD`: 5 -> 10

Separately, `ChannelBuilder.java:57` sets `maxAttempts` to 3, but the spec's retry policy specifies 4 (the initial attempt plus retries at 1s, 2s, 4s). The adjacent `// 1 + 2 + 4` comment suggests 4 was the intent, so this is an off-by-one; should be 3 -> 4. Harmless to fix now since nothing clamps at a 5000 cap.

toddbaert:
Corrects grace-period/max-retry-backoff options to prevent some logspam associated with this degenerate settings combination.

Fixes: https://github.com/open-feature/java-sdk-contrib/issues/1833

### DIFF
--- a/providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/State.java
+++ b/providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/State.java
@@ -21,7 +21,10 @@ public class State {
     /** The container borrowed from {@link ContainerPool} for this scenario. */
     public ContainerEntry containerEntry;
 
-    public ConcurrentLinkedQueue<Event> events = new ConcurrentLinkedQueue<>();
+    // events not yet consumed by a positive assertion; drained as they are matched
+    public ConcurrentLinkedQueue<Event> assertedEvents = new ConcurrentLinkedQueue<>();
+    // complete log of every event emitted, used for "never fired" assertions
+    public ConcurrentLinkedQueue<Event> allEvents = new ConcurrentLinkedQueue<>();
     public Optional<Event> lastEvent;
     public FlagSteps.Flag flag;
     public MutableContext context = new MutableContext();
--- a/providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/steps/EventSteps.java
+++ b/providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/steps/EventSteps.java
@@ -1,6 +1,7 @@
 package dev.openfeature.contrib.providers.flagd.e2e.steps;
 
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.awaitility.Awaitility.await;
 
 import dev.openfeature.contrib.providers.flagd.e2e.State;
@@ -24,7 +25,9 @@ public class EventSteps extends AbstractSteps {
     public void a_stale_event_handler(String eventType) {
         state.client.on(mapEventType(eventType), eventDetails -> {
             log.info("{} event tracked", eventType);
-            state.events.add(new Event(eventType, eventDetails));
+            Event event = new Event(eventType, eventDetails);
+            state.assertedEvents.add(event);
+            state.allEvents.add(event);
         });
     }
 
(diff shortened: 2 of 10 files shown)


providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/steps/EventSteps.java
```diff
@@ -24,7 +25,9 @@ public EventSteps(State state) {
     public void a_stale_event_handler(String eventType) {
         state.client.on(mapEventType(eventType), eventDetails -> {
             log.info("{} event tracked", eventType);
-            state.events.add(new Event(eventType, eventDetails));
+            Event event = new Event(eventType, eventDetails);
+            state.assertedEvents.add(event);
+            state.allEvents.add(event);
```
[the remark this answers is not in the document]
toddbaert: I think this is real but harmless - it would require:

- some event is positive asserted
- the same event is negatively asserted
- the handler is pre-empted between `assertedEvents.add` and `allEvents.add`

But no such test exists and would be nonsense anyway...

In any case I've reversed the order.

JamieSinn (approved):

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/steps/ProviderSteps.java
```diff
@@ -72,9 +72,9 @@ public void setupProvider(String providerType) throws InterruptedException {
         state.builder
                 .deadline(1000)
                 .keepAlive(0)
-                .retryGracePeriod(2)
+                .retryGracePeriod(5)
                 .retryBackoffMs(500)
-                .retryBackoffMaxMs(2000);
+                .retryBackoffMaxMs(500);
```
leakonvalinka, line 77: since i saw that on slack, should just scale the real values by a factor?

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/State.java
```diff
@@ -21,7 +21,10 @@ public class State {
     /** The container borrowed from {@link ContainerPool} for this scenario. */
     public ContainerEntry containerEntry;
 
-    public ConcurrentLinkedQueue<Event> events = new ConcurrentLinkedQueue<>();
+    // events not yet consumed by a positive assertion; drained as they are matched
+    public ConcurrentLinkedQueue<Event> assertedEvents = new ConcurrentLinkedQueue<>();
```
leakonvalinka, line 25: not sure i'm understanding this correctly but would `notAssertedEvents` be a better name then?

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/steps/ProviderSteps.java
toddbaert: What I actually want to do in a follow-up, is get all these specific adjustments FROM gherkin.

Right now they are just hard coded in the test implementation which is annoying. I don't think we'll do a scaling thing because that would be complicated or a new config - but getting it from the gherkin will be much better... I just didn't want to do that here.

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/State.java
toddbaert: This is for "negative" assertions - ie: make sure some event X was NOT fired - so we keep an un-drained list of all of them.

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/e2e/steps/EventSteps.java
```diff
@@ -53,20 +56,29 @@ public void eventHandlerShouldBeExecuted(String eventType) {
...
     }
 
+    @Then("the {} event handler should not have been executed")
+    public void eventHandlerShouldNotHaveBeenExecuted(String eventType) {
+        // checks the full log, not assertedEvents: preceding positive
+        // assertions may have drained an intervening event of this type
+        assertThat(state.allEvents.stream().anyMatch(event -> event.type.equals(eventType)))
+                .as("no %s event should have fired", eventType)
+                .isFalse();
+    }
```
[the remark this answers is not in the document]
toddbaert: The timing in the suite does this itself.

providers/flagd/src/test/resources/junit-platform.properties
```diff
@@ -8,10 +8,12 @@ cucumber.execution.parallel.enabled=true
...
-# Scenarios tagged @env-var mutate System env vars globally.
-# Serialise them behind an exclusive resource lock so concurrent scenarios
-# don't clobber each other's environment variable state.
-cucumber.execution.exclusive-resources.env-var.read-write=ENV_VARS
+# Scenarios tagged @env-var mutate System env vars globally, which every config
+# scenario reads when building FlagdOptions defaults. A plain ENV_VARS lock only
+# serialises @env-var scenarios against each other, so a concurrent default-reading
+# scenario could still observe a mutated var. Acquire the JUnit global lock so
+# @env-var scenarios run in full isolation (no other scenario runs concurrently).
+cucumber.execution.exclusive-resources.env-var.read-write=org.junit.platform.engine.support.hierarchical.ExclusiveResource
```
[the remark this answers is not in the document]
toddbaert: Good catch, fixed.

PR #1846, opened 2026-09-11
feat(provider-tck): let adopters add scenarios by convention

aepfli:
Lets an adopter run their own scenarios and steps inside the TCK's suite, in the same backend lifecycle phase, and exposes the suite's annotation values as constants.

Stacked on the report PR. Answers @toddbaert's review request on open-feature/spec#423: a vendor with provider-specific behaviour (flagd's `fractional`, say) should **extend** the TCK rather than maintain a parallel harness.

## What an adopter writes

Nothing. Drop a feature file in `src/test/resources/tck-extensions/` and a step class in `openfeature.tck.extensions`; both are picked up by the base suite. Java can scan, so this is convention rather than configuration — unlike Go and JavaScript, which need explicit options.

## Why the extension directory is separate

Measured on this module, not assumed:

| | |
| --- | --- |
| `@SelectClasspathResource` naming a resource that does not exist | hard discovery error |
| same directory name in two classpath roots | additive, both scanned |
| same directory **and filename** in two roots | **one silently wins** (`test-classes` beats the jar) |
| directory holding only a non-`.feature` file | resolves fine |

The third is why `tck-extensions/` is a separate name: had extensions shared `features/`, a vendor's `features/errors.feature` would silently *replace* the canonical file and the suite would go green having run theirs. The first is why the jar ships `tck-extensions/README.md` — the selector must resolve even when no adopter uses it.

An absent glue package is tolerated silently, so the extension glue package costs nothing unused.

## Constants

The features path, extension path, glue packages and Cucumber configuration values are now `public static final String` on `ProviderTck`. An adopter who does write their own `@ConfigurationParameter` can use `ProviderTck.GLUE + ",com.vendor.steps"` rather than restating our package as a magic string; annotation values permit compile-time constant concatenation.

## Second commit is separable

`feat(provider-tck): fail a run that executes less than the canonical set` can be dropped independently. It exists because a partial run otherwise produces a well-formed, valid report: in Go, `-run` on a single scenario passed green and emitted a report covering 1 of 29 canonical scenarios; the same hazard was live in Python. Extension scenarios cannot close a gap in the canonical set.

## Verified

- `mvn -pl tools/provider-tck test`: **31 tests, 0 failures**, including `ExtensionPointTest` and `CanonicalScenarioGuardTest`.
- Spotless could not be verified locally: this is a Windows checkout with `core.autocrlf=true`, so the working tree is CRLF and spotless reads the working tree. The committed blobs are LF (checked directly), so CI's Linux checkout should be clean — but that is inference, not a local green run.

The implementing agent was stopped before it finished its own verification pass; the commits are its work, the test run above is mine. Treat the end-to-end extension demonstration as claimed-but-unconfirmed until CI runs.


aepfli:
The extension point moved into #1830 and the canonical-set guard into #1841, so the chain stays a simple progression.

PR #1848, opened 2026-09-11
fix: shutdown blocks for unary deadline time

toddbaert:
* an earlier refactor changed the way the shutdown works, in order to remove a sleep
* resulted in incorrect order of some method calls in shutdown - we need to await AFTER we kill the channel
* without this fix, we always awaited for the configured unary RPC deadline before actually closing the channel

### DIFF
--- a/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamQueueSource.java
+++ b/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamQueueSource.java
@@ -174,9 +174,10 @@ public class SyncStreamQueueSource implements QueueSource {
 
         synchronized (this) {
             retryScheduler.shutdownNow();
+            // shut the channel down before awaiting
+            grpcComponents.channelConnector.shutdown();
             ShutdownUtils.awaitTerminationQuietly(
                     () -> retryScheduler.awaitTermination(deadline, TimeUnit.MILLISECONDS));
-            grpcComponents.channelConnector.shutdown();
         }
     }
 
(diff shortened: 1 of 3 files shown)


providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamQueueSource.java
```diff
@@ -174,9 +174,10 @@ public void shutdown() {
 
         synchronized (this) {
             retryScheduler.shutdownNow();
+            // shut the channel down before awaiting
+            grpcComponents.channelConnector.shutdown();
             ShutdownUtils.awaitTerminationQuietly(
                     () -> retryScheduler.awaitTermination(deadline, TimeUnit.MILLISECONDS));
-            grpcComponents.channelConnector.shutdown();
```
toddbaert, line 179: This ordering is wrong - we wait first, and then shutdown (this was due to a refactor which didn't correctly re-order these). 

Obviously, we need to shut down the channel and then wait.

providers/flagd/src/test/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/sync/SyncStreamQueueSourceTest.java
```diff
@@ -314,6 +316,30 @@ void onErrorEnqueuesDataPayload() throws Exception {
         verify(stub, times(2)).syncFlags(any(), any());
     }
 
+    @Test
```
toddbaert, line 319: This test fails without the fix. I also verified the bug and fix manually.

JamieSinn (approved):

PR #1851, opened 2026-09-13
fix: update Config defaults to boolean types and enhance environment variable handling

### LINKED ISSUE #1673: [flagd] Add log output on invalid env vars

matheusandre1:
## This PR

- `Config.fallBackToEnvOrDefault(String, int)` and `(String, long)`: catch `NumberFormatException` and log an error containing the env var name, the invalid value, the expected type and the default that is applied.
- New `Config.fallBackToEnvOrDefault(String, boolean)`: only accepts `true`/`false` (case-insensitive) and logs an error for any other value. Previously `Boolean.parseBoolean` silently treated any value other than `true` (e.g. `yes`, `1`) as `false`.
- `FlagdOptions`: `tls` and `reinitializeOnError` now use the boolean helper. `DEFAULT_TLS` and `DEFAULT_REINITIALIZE_ON_ERROR` became `boolean` constants (package-private, no public API change).

Example output:

ERROR dev.openfeature.contrib.providers.flagd.Config - Invalid value 'abc' for environment variable FLAGD_DEADLINE_MS: expected an integer. Falling back to default value '500'.
ERROR dev.openfeature.contrib.providers.flagd.Config - Invalid value 'yes' for environment variable FLAGD_TLS: expected 'true' or 'false'. Falling back to default value 'false'.

When flagd options are provided via environment variables and the value cannot be parsed, the provider silently fell back to the default value. This PR adds an `ERROR` log entry for such misconfigurations, as suggested in the issue, since they require user intervention and can lead to unexpected behavior.

### Related Issues
Closes: #1673 

### Notes

- Log level is `ERROR`, following the preference expressed in the issue.
- Existing `WARN` logs for `FLAGD_RESOLVER`, `FLAGD_COMPILE_TARGETING` and the port fallbacks are left untouched. The port case is intentionally lenient because of Kubernetes service-link env injection (#1835).
- Fallback behavior is unchanged: invalid values still resolve to the default.

### How to test

Added tests in `FlagdOptionsTest.InvalidEnvironmentVariables` covering invalid int, decimal int, invalid long, invalid boolean and case-insensitive boolean parsing.


providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/Config.java
```diff
@@ -84,21 +84,55 @@ static String fallBackToEnvOrDefault(String key, String defaultValue) {
...
+        final String value = System.getenv(key);
+        if (value == null) {
+            return defaultValue;
+        }
         try {
-            return System.getenv(key) != null ? Long.parseLong(System.getenv(key)) : defaultValue;
-        } catch (Exception e) {
+            return Long.parseLong(value);
+        } catch (NumberFormatException e) {
+            logInvalidEnvValue(key, value, "an integer", defaultValue);
```
toddbaert:
```suggestion
            logInvalidEnvValue(key, value, "a whole number", defaultValue);
```

providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/Config.java
```diff
@@ -84,21 +84,55 @@ static String fallBackToEnvOrDefault(String key, String defaultValue) {
...
+        final String value = System.getenv(key);
+        if (value == null) {
+            return defaultValue;
+        }
         try {
-            return System.getenv(key) != null ? Integer.parseInt(System.getenv(key)) : defaultValue;
-        } catch (Exception e) {
+            return Integer.parseInt(value);
+        } catch (NumberFormatException e) {
+            logInvalidEnvValue(key, value, "an integer", defaultValue);
```
toddbaert:
```suggestion
            logInvalidEnvValue(key, value, "a whole number", defaultValue);
```

toddbaert (approved):

PR #1778, opened 2026-04-21
fix: various custom operator conformance fixes

toddbaert:
fix: various custom operator conformance fixes

* support v/V prefix in semver
* support partial versions in semver
* support numeric context values in semver
* return null on errors
* fix fractional single-entry flattening

:warning: wait until flagd/RPC is updated to merge.

Fixes: https://github.com/open-feature/java-sdk-contrib/issues/1774

---

Related:
* https://github.com/open-feature/flagd/pull/1950
* https://github.com/open-feature/js-sdk-contrib/pull/1519
* https://github.com/open-feature/dotnet-sdk-contrib/pull/635
* https://github.com/open-feature/python-sdk-contrib/pull/386
* https://github.com/open-feature/go-sdk-contrib/pull/874

PR #1855, opened 2026-09-14
chore: document and complete the .gitmodules branch pins

aepfli:
Follow-up to #1853, which restored the `branch` pins but not the reasoning behind them.
Parent: open-feature/flagd-testbed#398

Adds a header comment to `.gitmodules` explaining what the `branch` lines are for, pins the two `schemas` submodules, and switches all four flagd submodule URLs to their canonical form.

The pin has been removed as drive-by cleanup five times across the SDK repos — twice here, in #1778 — and each time it silently switched off Renovate's submodule updates with no visible failure. Nothing in the file said otherwise, so this is the cheapest place to say it.

| submodule | commit | pin |
|---|---|---|
| `providers/flagd/schemas` | `1daf5ff` | `json/json-schema-v0.2.15` |
| `tools/flagd-core/schemas` | `1daf5ff` | `json/json-schema-v0.2.15` |

`json/json-schema-*` is a release-please component tag and does not parse as semver, so these pins stay inert until open-feature/community-tooling#126 merges. Harmless until then.

**URLs.** `test-harness` → `flagd-testbed` and `schemas` → `flagd-schemas`. Both repos were renamed and the old URLs only work through GitHub's redirect, which stops working if a repo with the old name is ever created again. Nothing moves: existing clones keep their cached URL and carry on, `git submodule sync` adopts the new one.

**Merged `main` in.** The url change sat on the line directly above the testkit pin that #1856 had just bumped to `v3.10.1`, so the two edits overlapped and conflicted. The merge keeps both. As a side effect this clears the drift that open-feature/flagd-testbed#398 flagged: `providers/flagd/test-harness` and `tools/flagd-api-testkit/test-harness` are back in lockstep, both on `v3.10.1` / `b308c1b`.

#1856 is worth a look on its own — it is Renovate bumping the testkit testbed through the restored pin, which is exactly the behaviour this whole effort is trying to get back.

Left unpinned: `spec` (untagged; pinning would roll it back to `v0.9.0`) and `wasm-releases` (no tags, only `publish-wasm-evaluation-v*` branches).

Refs: open-feature/community-tooling#125


aepfli:
Pushed a commit extending this PR from the explanatory comment to the two `schemas` submodules, so the repo is covered in one pass.

Both already sit exactly on `json/json-schema-v0.2.15`, so neither moves:

| submodule | recorded commit | tag |
|---|---|---|
| `providers/flagd/schemas` | `1daf5ff` | `json/json-schema-v0.2.15` |
| `tools/flagd-core/schemas` | `1daf5ff` | `json/json-schema-v0.2.15` |

That is a release-please component tag and does not parse as semver, so it needs the prefix rule in open-feature/community-tooling#126 to resolve as a version. Until that merges the pin is inert, but harmless.

**One thing to flag before merging.** This branch is 3 commits behind `main`, and the drift is visible in the testbed pins:

| | this branch | `main` |
|---|---|---|
| `tools/flagd-api-testkit/test-harness` pin | `v3.8.0` | `v3.10.1` |
| its gitlink | `7575a1d` (= v3.8.0) | `b308c1b` (= v3.10.1) |

Both sides are internally consistent — the branch simply predates main's bump — and I left it alone rather than hand-editing it, since the branch's pin and gitlink agree. Git's three-way merge should keep main's `v3.10.1`, because this branch never touched that line. Worth a glance at the merge result all the same, since silently reverting the testkit testbed to `v3.8.0` is exactly the class of regression open-feature/flagd-testbed#398 is about. Rebasing on `main` first would remove the doubt entirely.

Still deliberately unpinned: `spec` (pending the cadence question in open-feature/community-tooling#125) and `wasm-releases`, which has no tags, only `publish-wasm-evaluation-v*` branches.

